Issue #18, opened 2018-02-20
shell injection


zricethezav:
https://github.com/zricethezav/gitleaks/blob/master/leaks.go#L29 could be pwned by malicious repo urls. need to do some regexing or use https://github.com/src-d/go-git for the clones.

anant2614:
Should we use go-git for now ? I can take this up.

zricethezav:
@anant2614 we should use go-git for cloning at the very least. The other git operations should be fine to shell out since `rev-list` is a defined string. `git diff` should also be to shell out too. If you want to take this on that would be much appreciated! 👍

mcuadros:
One question why `git diff` should be the shell output? It's something wrong with the diff output from go-git?

zricethezav:
Ideally, gitleaks would use go-git for all git commands. I raised this issue specifically for the clone. Using go-git for other commands are nice-to-haves... but also easy to implement so I agree with you, go-git should be used for the diff and rev-list (if supported).

anant2614:
@zricethezav well I worked just on the clone part.I went through the go-git docs of supported commands,could not find git-diff and rev-list features and these are open feature requests in their issues.

zricethezav:
closing this. @mcuadros feel free to open an issue for getting `rev-list` and `diff` working for go-git.

mcuadros:
I just asking because diff and rev-list are being used intensively, and
documented ad supported.

https://github.com/src-d/go-git/blob/master/COMPATIBILITY.md

So I wanted to know if you had any missing features.

>

zricethezav:
@mcuadros I see that now. I'll be looking into replacing all shelled cmds with go-git in the future when I have some more time to develop this.

zricethezav:
@mcuadros v1.0.0 is now powered by go-git. Awesome package! Thanks source-d :+1:

PR #20, opened 2018-02-21
use go-git for cloning

anant2614:


anant2614:
This refers to #18 . @zricethezav can you please review?

zricethezav:
This is great, thanks! I'll look at it more closely this evening. 👍

leaks.go
```diff
@@ -25,14 +26,25 @@ func start(opts *Options) {
...
+	}
+
+	repoName := getLocalRepoName(opts.RepoURL)
 	fmt.Printf("Cloning \x1b[37;1m%s\x1b[0m...\n", opts.RepoURL)
-	err := exec.Command("git", "clone", opts.RepoURL).Run()
+	r,err := git.PlainClone(dir + "/" + repoName, false, &git.CloneOptions{
+		URL:               opts.RepoURL,
+		RecurseSubmodules: git.DefaultSubmoduleRecursionDepth,
+	})
+	log.Printf("Cloned the repo::",r)
```
zricethezav: lets remove this log as the "Evaluating" message indicates a successfully cloned repo

leaks.go
```diff
@@ -25,14 +26,25 @@ func start(opts *Options) {
...
 	if err != nil {
 		log.Printf("failed to clone repo %v", err)
```
zricethezav: before the "failed to clone repo" message, call `cleanup` to remove the directory. I was doing some testing and noticed even if go-git fails to clone the directory will still be made

anant2614:
@zricethezav the PR has been updated with the requested changes.

leaks.go
```diff
@@ -25,14 +26,25 @@ func start(opts *Options) {
 	c := make(chan os.Signal, 2)
 	signal.Notify(c, os.Interrupt, syscall.SIGTERM)
 
+	//get current working directory
+	dir,err := os.Getwd()
```
zricethezav: one more change, use appRoot instead of dir

leaks.go
```diff
@@ -27,15 +27,15 @@ func start(opts *Options) {
 	signal.Notify(c, os.Interrupt, syscall.SIGTERM)
 
 	//get current working directory
-	dir,err := os.Getwd()
+	appRoot, err := os.Getwd()
```
zricethezav: appRoot is already defined here: https://github.com/zricethezav/gitleaks/blob/master/main.go#L24

anant2614: Ya,changing again.

PR #24, opened 2018-02-21
Use GOs flag package for command line parsing

timdev21:
Changed the command line parser to use the [flag](https://golang.org/pkg/flag/) package already in go included, makes the code more readable and expandable in my opinion. The -h --help flags are added automatically. Maybe even add the contents of options.go to main.go? 

Happy to hear what you think!

### DIFF
--- a/leaks.go
+++ b/leaks.go
@@ -21,18 +21,18 @@ type LeakElem struct {
 }
 
 // start clones and determines if there are any leaks
-func start(opts *Options) {
+func start(repoURL string) {
 	c := make(chan os.Signal, 2)
 	signal.Notify(c, os.Interrupt, syscall.SIGTERM)
 
-	fmt.Printf("Cloning \x1b[37;1m%s\x1b[0m...\n", opts.RepoURL)
-	err := exec.Command("git", "clone", opts.RepoURL).Run()
+	fmt.Printf("Cloning \x1b[37;1m%s\x1b[0m...\n", repoURL)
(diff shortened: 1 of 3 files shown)


zricethezav:
This is great, thanks! I'll look at it more closely this evening. 👍

zricethezav:
@timdev21, I would like to maintain the ability to use gitleaks on a repo url without any required keyword arguments. It looks like `./gitleaks <url>` breaks as flag is looking for keyword arguments, not positional arguments. Ideally I would like it to look like `./gitleaks [options] <url>` where `<url>` is positional and defaults to a repo target. `<url>` can be switched from a repo target by supplying a `-o` or `-u` keyword argument. 

Some samples:
`./gitleaks -u https://github.com/zricethezav`
`./gitleaks -o https://github.com/myorg`
`./gitleaks https://github.com/zricethezav/gronit`
`./gitleaks -r https://github.com/zricethezav/gronit`
`./gitleaks -s -c 100 -x 60 https://github.com/zricethezav/gronit`

I think I had originally decided to roll my own for options parsing as I was unable to get flag to function like this for another project. However, this PR cleans up quite a bit of code and makes it look nicer. If you want to attempt to implement what I'm asking, go for it, if not, no worries. This looks good. The only thing I would require you do is update the README. 

I agree that options.go is small enough it makes sense to axe it and put its contents in main.go

timdev21:
I've added the positional argument. With this method we could implement that it is possible to pass more than one url after the options, but thats just a quick thought. 
What do you think?

PR #26, opened 2018-02-22
Added support for docker

devnixs:
It would be even better if you could setup automated builds so we would automatically have the latest version pushed to docker hub.
I did that with my fork raphaelareya/gitleaks, but it won't get automatically updated.

It's pretty straightforward https://docs.docker.com/docker-hub/builds/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,20 @@
+FROM golang:1.10.0-alpine3.7
+
+WORKDIR /app
+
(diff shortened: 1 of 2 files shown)


arjunyel:
What about using a multistage build?

zricethezav:
https://hub.docker.com/r/zricethezav/gitleaks/ is up. We will want multistage build

PR #28, opened 2018-02-22
Added pagination for users and orgs that have more than 30 repositories

timdev21:
By default the GitHub API only returns 30 entries, so we need to paginate the results to correctly gather all available repositories. This is a quick fix to accomplish this. Another option would be to use the github-go package. Also added the option to specify an oauth token to increase the ratelimit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,7 @@ Options:
 	-s 			Strict mode uses stopwords in checks.go
 	-e 			Base64 entropy cutoff, default is 70
 	-x 			Hex entropy cutoff, default is 40
+	-t 			GitHub OAuth Token
 	-h --help 		Display this message
 ```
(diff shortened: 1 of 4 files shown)


zricethezav:
@timdev21 good catch and thanks for working on all these issues. Master now implements go-git so we could see if go-git supports fetching the whole org repo list?

timdev21:
I'm on my phone so I can't verify, but iirc go-git is just for git and not for GitHub, so I don't think it implements the API. But it might be worth it to check again. For a clean version we could use go-github, it mirrors the API very well imo.On 22 Feb 2018 18:39, Zachary Rice <notifications@github.com> wrote:@timdev21 good catch and thanks for working on all these issues. Master now implements go-git so we could see if go-git supports fetching the whole org repo list?

zricethezav:
I like the idea of keeping it clean so lets implement this in go-github. Very clean implementation of it used here: https://github.com/anshumanbh/git-all-secrets/blob/master/main.go

zricethezav:
implemented with go-github here: https://github.com/zricethezav/gitleaks/releases/tag/v0.2.0

PR #32, opened 2018-02-22
Remove `|` characters from character classes

ryan-copperleaf:
Feel free to not include the test.


zricethezav:
using `(?i)` https://github.com/zricethezav/gitleaks/blob/master/main.go#L53

PR #42, opened 2018-02-24
Case insensitive regex matching

Shawn-Smith:
This changes the current regexs to ones that aren't case sensitive.


zricethezav:
cleaner 👍

PR #46, opened 2018-02-25
Simplify the stopwords check

Shawn-Smith:
Change to strings.EqualFold for a case insensitive search of stopwords.


Shawn-Smith:
Whoops. Misunderstood the way stopwords were handled.

Will fix and re-request the pull.

PR #47, opened 2018-02-25
Make stopwords lowercase

Shawn-Smith:
By using strings.ToLower we can reduce the number of times we need to loop in the stopwords check.

### DIFF
--- a/main.go
+++ b/main.go
@@ -37,8 +37,7 @@ func init() {
 	base64Chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/="
 	hexChars = "1234567890abcdefABCDEF"
 
-	stopWords = []string{"setting", "Setting", "SETTING", "info",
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -37,8 +37,7 @@ func init() {
 	base64Chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/="
 	hexChars = "1234567890abcdefABCDEF"
 
-	stopWords = []string{"setting", "Setting", "SETTING", "info",
```
zricethezav: Looks like you removed the Upper case cases so we would be calling `ToLower` on already lowered case strings. A better way to implement this would be to check this would be doing some case insensitive regex checks.

Shawn-Smith: We're calling ToLower on what we check the stopwords against, not the stopwords themselves.

zricethezav: I've been staring at a screen too long. >.<
👍

Shawn-Smith: We all have days like that lol

PR #54, opened 2018-03-03
Adding DSA key check to regex list

keirans:
This is another common filetype we should add to the base set of searches.


zricethezav:
👍

PR #82, opened 2018-08-03
Update repo.go

zamarrowski:


zamarrowski:
This is broken when you execute in local mode and you don't have internet connection

zricethezav:
v1.0.0 removes `repo.go`

PR #85, opened 2018-08-09
GitHub Enterprise support + improved error handling. Fixes #84

### LINKED ISSUE #84: Add support for GitHub Enterprise
When scanning single repositories it's possible to provide a custom URL to a GitHub Enterprise instance, however when running the Org/User scan the tool assumes the public GitHub instance.

It would be create to have an option like "Base GitHub URL" to override this to support GitHub Enterprise deployments.

eripa:


eripa:
Yeah it's tricky to test for sure, but given the way the library and GitHub works where it's only the URL that changes (it's still GitHub on the other side) there isn't much that differs. I've verified it.

I can update the README+CHANGELOG.

eripa:
A couple of options from `--help` was added by this change, like `--version` and `--branch`. I suppose you just forgot to update it the last time.

I'll let you do the actual tagging and release ;)

zricethezav:
Awesome perfect. Thanks for picking these up 👍

PR #88, opened 2018-08-23
Added Support for branches

Fabse333:
This bugfix should allow to specify a Branch (opts.Branch is elsewise not used in project). Is my first Go program, so I'm sorry in case the syntax is not optimal.

### DIFF
--- a/main.go
+++ b/main.go
@@ -445,19 +445,21 @@ func auditRepo(r *git.Repository) ([]Leak, error) {
 			return nil
 		})
 	} else {
-		foundBranch := false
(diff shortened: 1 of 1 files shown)


zricethezav:
Care to add a test in this test function: https://github.com/zricethezav/gitleaks/blob/master/gitleaks_test.go#L294? Use https://github.com/gitleakstest/gronit as a test target repo. 

Otherwise this looks good, thanks for grabbing this

Fabse333:
My time does not allow to write a test unfortunately.

PR #89, opened 2018-08-31
Fixes

zamarrowski:


zamarrowski:
I added a regex to find uri credentials, move constants to constants.go file and update version.

zricethezav:
Stylistically, I prefer to have this relatively small app contained inside one file, main.go. I appreciate you taking a look at this though.

PR #94, opened 2018-09-24
 Fix string formatting directive

mahirrudin:
There are some warning about string formatting directive


zricethezav:
Looks good but lets leave the Dockerfile in there

mahirrudin:
> Looks good but lets leave the Dockerfile in there

Yup, dockerfile was accidentally commit.. pardon

PR #107, opened 2018-10-08
Add option to check string entropy

### LINKED ISSUE #97: Entropy checks
Although the first version did entropy checks apart from regexes (https://github.com/zricethezav/gitleaks/blob/v0.1.0/checks.go#L34), that's no longer the case in the latest version.

I know entropy checks could create lots of false positives, but they can also catch things that regex don't so I see a lot of value in there. Also, using the whitelisting options can reduce a lot of the false positives.

Could this be added again?

camaya:
Strings with high entropy have a good probability of being a secret.
This would close #97

### DIFF
--- a/main.go
+++ b/main.go
@@ -7,6 +7,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"math"
 	"net"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -81,13 +82,14 @@ type Options struct {
...
-	SSHKey        string `long:"ssh-key" description:"path to ssh key"`
-	ExcludeForks  bool   `long:"exclude-forks" description:"exclude forks for organization/user audits"`
+	MaxGoRoutines int     `long:"max-go" description:"Maximum number of concurrent go-routines gitleaks spawns"`
+	Disk          bool    `long:"disk" description:"Clones repo(s) to disk"`
+	AuditAllRefs  bool    `long:"all-refs" description:"run audit on all refs"`
+	SingleSearch  string  `long:"single-search" description:"single regular expression to search for"`
+	ConfigPath    string  `long:"config" description:"path to gitleaks config"`
+	SSHKey        string  `long:"ssh-key" description:"path to ssh key"`
+	ExcludeForks  bool    `long:"exclude-forks" description:"exclude forks for organization/user audits"`
+	Entropy       float64 `long:"entropy" short:"e" description:"Report a finding when a string has at least the entropy level you defined"`
```
zricethezav: I would change the description to:
`Include entropy checks during audit. Entropy scale: 0.0(no entropy) - 8.0(max entropy)`

PR #108, opened 2018-10-09
Fix commit option

camaya:
The --commit option was being ignored because the `Close()` function doesn't stop
the iterator.
According to the documentation:
https://godoc.org/gopkg.in/src-d/go-git.v4/plumbing/object#FileIter.ForEach
an `ErrStop` error should be send to stop it.

I'm not sure what the what the `Close()` function actually does and  maybe that can be removed.

### DIFF
--- a/gitleaks_test.go
+++ b/gitleaks_test.go
@@ -560,6 +560,14 @@ func TestAuditRepo(t *testing.T) {
 			numLeaks:    0,
 			configPath:  path.Join(configsDir, "repo"),
 		},
+		{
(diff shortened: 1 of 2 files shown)


zricethezav:
awesome, will be looking at these. not sure why PRs are getting a github rate limit, might need to update my api token

zricethezav:
Ah, https://blog.travis-ci.com/2016-07-07-security-advisory-encrypted-variables makes sense. I should probably be mocking those api calls. I'll pull this down, test, then merged in if passing

PR #113, opened 2018-10-10
Validate --report option

camaya:
I found the need for this the hard way haha.
I was using the tool on a large repo and I set the `--report`option to a directory instead of a file so when the execution finished, I had no report :(

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 CHANGELOG
 =========
 
+1.11.2
+----
+- Validate report option
 1.11.1
 ----
 - fix commit whitelist logic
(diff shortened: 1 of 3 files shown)


zricethezav:
great catch!

zricethezav:
This is a experience thing.. but what if we remove `--csv` and enforce either `.json` or `.csv` if we are checking for `.json` as we are here. I'm happy to implement that in another release.

camaya:
I like the idea! I think it's enough if the help for the `--report`option is clear about it.

zricethezav:
Check out https://github.com/zricethezav/gitleaks/pull/114

PR #116, opened 2018-10-14
Documentation enhancements

keirans:
Hey @zricethezav ,
I've taken the new version for a spin, it's looking good, I've got a few small enhancements coming your way.

I have updated the documentation and examples a bit for you.

Cheers!

K


zricethezav:
@keirans gonna be giving this a look. Travis fails because it depends on encrypted env vars and unfortunately does not work w/ forks :/

Another thought that I had was that with the new github software releases: https://blog.github.com/2018-10-16-future-of-software/#github-token-scanning-for-public-repositories-public-beta it would be worthwhile to expound on how gitleaks differs from this tool... auditing large repos/orgs, white-listing, etc.

keirans:
I've updated the PR to talk a bit about the token scanning capability, in the context of gitleaks :)

Cheers,

K

keirans:
Hello !
I've updated the documentation to include the github PR support, and docker example, and removed the version string that you mentioned.

I think we are good to go now, Let me know if I have missed anything else. :)

PR #121, opened 2018-10-22
Added Telegram Bot Token

jfowl:
Telegram's bot API also has tokens (https://core.telegram.org/bots/api#authorizing-your-bot) that you don't want to upload to github.


jfowl:
Where did this other Telegram Regex come from? In my opinion it is not sufficient to catch bot tokens.

Taken from https://github.com/telegraf/telegraf:
```
const Telegraf = require('telegraf')
const bot = new Telegraf(process.env.BOT_TOKEN)
```

You often just add the bot token without the leading api.telegram.org.

jfowl:
:(

PR #133, opened 2018-10-27
fix slack token format

akerl:
This fixes the Slack token format. I'm not sure if there was a time when `xoxo[bapr]` was the correct regex, but per https://api.slack.com/docs/token-types , the 2nd `o` doesn't occur in their tokens. This PR additionally adds coverage for `xoxs`, which is the prefix for Slack official client session tokens.


akerl:
Updated that file as well.

PR #145, opened 2018-11-06
Remove break lines in commit messages

camaya:
Breaklines in a csv report caused issues when importing the data to an
spreadsheet application like LibreOffice Calc or Microsoft Excel.


zricethezav:
good deal, ill cut a release sometime today

PR #149, opened 2018-11-13
Fix files whitelisting when checking Github PRs

camaya:
Files whitelisting code was not executed when running the tool against a
Github PR so the config was ignored.

To avoid duplication, the whitelist files check has been moved to the `inspect`
function which is always executed.


camaya:
Hey Zachary, 

Let me know if there's something pending from this PR

Cheers!

PR #151, opened 2018-11-15
Add support for tokens

AdamKobi:
Gitleaks doesn't handle the case when working with github api for cloning the repo.
At the moment you can only clone repos with SSH key or if the repo is public.
When scanning an organisation with many private repos and running this on Jenkins for example without SSH key the cloning will fail with "authentication required".

Added new functionality to authenticate with same GITHUB_TOKEN provided for listing and other methods.

No tests were added as this is a bit problematic to test without real token.
Please notice that gitHttp receives a fake username and the token, the username can't be an empty string but it has no meaning, only the token matters.

I've tested the new change on my private repos and it's working as it should.

In addition, the code no longer tries to use ssh keys if providing a token (previous version defaulted to using id_rsa in ~/.ssh/ which is an issue if the user trying to clone the repo wants to clone it with the token or using normal https clone as the flow in the code automatically assumes you want to clone the repo with ssh if it finds it which it usually finds in Linux machines).

Sidenote: i'm rather new to GO so hope it's ok.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 CHANGELOG
 =========
 
+1.21.0
+----
+- added support for cloning repositories using github api
+
 1.20.0
 ----
 - adding gitlab user and group support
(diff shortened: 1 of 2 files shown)


zricethezav:
awesome, appreciate it

PR #158, opened 2018-11-30
Change redacted behaviour on the line information

camaya:
The line information sometimes contains useful context information so
instead of redacting the whole line, this change redacts only the
secret


zricethezav:
appreciate these PRs! will be putting together a new release soon

PR #159, opened 2018-12-04
Reduce noise when running entropy checks

### LINKED ISSUE #155: Reduce noise from report when running entropy checks
It's hard to find all the secrets using regexes and entropy checks help to find other secrets I may have missed otherwise. The problem is that it generates a lot of noise. After checking multiple reports (sometimes more than 1GB of a CSV file having entropy configured in 5), I'd say we could reduce the noise by checking the line of the find for keywords like: "secret", "password", "token", "signature", "curl", "wget", "http", "https", etc. 

The noise reduction would be very noticeable and the report much more accurate.

I'm not sure whether this should be the default behaviour of the entropy option or it'd be better to add a new option like --noise-reduction that would only work when entropy checks are enabled.

What do you think?

camaya:
When there's a word on a line with high entropy, we check if the line
matches any of the regex configured on the `lineregexes` element of the
`entropy` section. If so, it's marked as a leaked.

This has reduced the number of leaks found with entropy enabled up to
80%

This would close #155

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,30 +63,31 @@ Usage:
   gitleaks [OPTIONS]
 
 Application Options:
-  -r, --repo=          Repo url to audit
-      --github-user=   Github user to audit
-      --github-org=    Github organization to audit
-      --github-url=    GitHub API Base URL, use for GitHub Enterprise. Example: https://github.example.com/api/v3/ (default: https://api.github.com/)
-      --github-pr=     Github PR url to audit. This does not clone the repo. GITHUB_TOKEN must be set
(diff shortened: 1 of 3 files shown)


camaya:
@zricethezav Let me know your thoughts on this :)
Also, the wiki page should be updated to include the config for the new regexes

Cheers!

camaya:
@zricethezav sure! So, `--entropy` would include everything and `--noise-reduction` would apply these regexes, is that ok?

zricethezav:
@camaya exactly

Issue #162, opened 2018-12-13
whiteListFiles regexes are checked unnecessarily for every chunk of each file


milo-minderbinder:
The whiteListFiles regexes are checked in the `inspect(diff gitDiff) []Leak` function, instead of being called in the `auditGitReference(repo *RepoDescriptor, ref *plumbing.Reference) []Leak` function where `inspect` is called. This is redundant, since inspect is called for every 'chunk' and there may be many chunks in a single `FilePatch`.

Instead, the whiteListFiles regexes should be checked against each `FilePatch` _before_ generating the list of `Chunk`s for the `FilePatch` and calling `inspect` on those chunks. @

zricethezav:
thanks for handling this!

Issue #161, opened 2018-12-13
Whitelist regexes checks inconsistent and sometimes redundant


milo-minderbinder:
# Expected Behavior
For each line with a regex finding, an entropy finding, or both, each whitelist regex should be checked once and only once.

# Observed Behavior
Currently, for a given line if:
1. No whitelist regexes match the line
2. Two or more finding regexes match the line

then the full set of whitelist regexes are checked against the line for each matched finding regex in *[2]* unnecessarily.

Alternatively, for a given line, if:
1. One or more whitelist regexes match the line
2. No finding regexes match the line
3. One or more entropy findings are found in the line

then the whitelist regex is never checked and the entropy finding(s) are incorrectly reported, even though one or more whitelist regex would match the line if checked.

zricethezav:
again, thanks!

PR #163, opened 2018-12-13
fix file (#162) & line (#161) whitelist checks

milo-minderbinder:
* only check entropyRegexes and whiteListRegexes once per line, and only
  if regex or entropy leak type was found, since regex checks are expensive
* check whiteListFiles once for each FilePatch, instead of for
  every Chunk

### DIFF
--- a/main.go
+++ b/main.go
@@ -23,6 +23,7 @@ import (
 
 	"gopkg.in/src-d/go-git.v4/plumbing"
 
+	diffType "gopkg.in/src-d/go-git.v4/plumbing/format/diff"
 	"gopkg.in/src-d/go-git.v4/plumbing/object"
 	"gopkg.in/src-d/go-git.v4/plumbing/transport/ssh"
 	"gopkg.in/src-d/go-git.v4/storage/memory"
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -618,12 +624,6 @@ func inspect(diff gitDiff) []Leak {
 		skipLine bool
 	)
 
-	for _, re := range whiteListFiles {
```
zricethezav: We need to add this to https://github.com/zricethezav/gitleaks/blob/master/github.go#L53-L57. Test for github PR is failing with these changes

main.go
```diff
@@ -575,12 +576,17 @@ func auditGitReference(repo *RepoDescriptor, ref *plumbing.Reference) []Leak {
 					} else if to != nil {
 						filePath = to.Path()
```
milo-minderbinder: Definitely related, but perhaps beyond the scope of this PR: would it make more sense to default the `filePath` to the `to.Path()`, since that is likely the value expected/wanted by the user since it reflects the current (in the context of the commit) path of the file or the _new_ path in the case where the commit renamed/moved the file? Perhaps I'm missing something or misinterpreting the API, but thought I'd ask.

Issue #168, opened 2018-12-28
fatal error: runtime: out of memory due at --report


arbazkiraak:
Hi @zricethezav ,

```
→ gitleaks --repo=https://github.com/pricelinelabs/design-system -e 5 --depth=100 --report=test.json
INFO[2018-12-28T09:12:18Z] cloning https://github.com/pricelinelabs/design-system 
Enumerating objects: 304, done.
Counting objects: 100% (304/304), done.
Compressing objects: 100% (146/146), done.
Total 10075 (delta 136), reused 228 (delta 113), pack-reused 9771
INFO[2018-12-28T09:13:17Z] writing report to test.json                  
fatal error: runtime: out of memory

runtime stack:
runtime.throw(0xb09987, 0x16)
        /usr/local/go/src/runtime/panic.go:616 +0x81
runtime.sysMap(0xc457830000, 0x247f0000, 0xc420000100, 0xef9738)
        /usr/local/go/src/runtime/mem_linux.go:216 +0x20a
runtime.(*mheap).sysAlloc(0xee0060, 0x247f0000, 0x67aba1)
        /usr/local/go/src/runtime/malloc.go:470 +0xd4
runtime.(*mheap).grow(0xee0060, 0x123f2, 0x0)
        /usr/local/go/src/runtime/mheap.go:907 +0x60
runtime.(*mheap).allocSpanLocked(0xee0060, 0x123f2, 0xef9748, 0x7f86d91e5dd8)
        /usr/local/go/src/runtime/mheap.go:820 +0x301
runtime.(*mheap).alloc_m(0xee0060, 0x123f2, 0x7f86da990101, 0x4521b0)
        /usr/local/go/src/runtime/mheap.go:686 +0x118
runtime.(*mheap).alloc.func1()
        /usr/local/go/src/runtime/mheap.go:753 +0x4d
runtime.(*mheap).alloc(0xee0060, 0x123f2, 0x7f86d9010101, 0x415a2c)
        /usr/local/go/src/runtime/mheap.go:752 +0x8a
runtime.largeAlloc(0x247e3cf4, 0x450101, 0x7f86da99a6c8)
        /usr/local/go/src/runtime/malloc.go:826 +0x94
runtime.mallocgc.func1()
        /usr/local/go/src/runtime/malloc.go:721 +0x46
runtime.systemstack(0x7ffc00000000)
        /usr/local/go/src/runtime/asm_amd64.s:409 +0x79
runtime.mstart()
[...]

PR #171, opened 2019-01-04
fix memory issue when writing json report (#168)

milo-minderbinder:
Previously, gitleaks.writeReport used json.MarshallIndent on the entire
set of leaks when writing json reports, which would cause failures due
to excessive memory consumption when a large number of leaks were
identified.

This PR addresses the problem by doing incremental marshalling and
writing of leaks to the report file, and reduces the memory footprint by
several GB in tests with >30k leaks.

### DIFF
--- a/main.go
+++ b/main.go
@@ -365,8 +365,41 @@ func writeReport(leaks []Leak) error {
 		}
 		w.Flush()
 	} else {
-		reportJSON, _ := json.MarshalIndent(leaks, "", "\t")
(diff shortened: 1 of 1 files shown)


milo-minderbinder:
including some profile results I generated based on the findings set from the original issue thread (which had 32,934 leaks in total)

original:
image

incremental writes:
image

zricethezav:
Looks good, thanks for this. I might iterate on this and do send these writes to a queue that writes the leaks to a file as leaks come in. Thanks for the contribution.

PR #176, opened 2019-01-14
Fix for the github default regex

phundisk:
Exclude finding if the line has reference to github.com which could pull a repo ref hash and be a false positive
Issue https://github.com/zricethezav/gitleaks/issues/175


phundisk:
Need to fix something with the regex

PR #191, opened 2019-04-02
Regex for AWS Secret Keys + minor fixes

sjuxax:
Fixed a couple of typos, updated the sample configuration constant, disambiguated between AWS Identifiers and Keys, and added a regex that will match many AWS Secret Keys by default.


zricethezav:
Gonna close this since gitleaks has a new config structure. Thank you for the PR

PR #216, opened 2019-05-16
Scan the both public and private repo for a given user

cafeliker:
In GitHub Enterprise, a personal access token is always required to clone the either public or private repos. Updated the repo.go to read the environment variable GITHUB_TOKEN for repo clone activity.


cafeliker:
@zricethezav can you share me some comments? I'd like to contribute to this nice project, thanks.

cafeliker:
Add another fix in the github.go to be able to scan the both public and private repo for a given user.

Pls see the source code at https://github.com/google/go-github/blob/master/github/repos.go#L165

if user != "" {
	u = fmt.Sprintf("users/%v/repos", user)
} else {
	u = "user/repos"
}
It means only user=="", it will call List your repositories API, otherwise it will call List user repositories
which only list public repos.

src/github.go
```diff
@@ -133,7 +133,7 @@ func auditGithubRepos() ([]Leak, error) {
 			break
 		}
 		if opts.GithubUser != "" {
-			pagedGithubRepos, resp, err = githubClient.Repositories.List(ctx, opts.GithubUser, githubOptions)
+			pagedGithubRepos, resp, err = githubClient.Repositories.List(ctx, "", githubOptions)
```
zricethezav: I think this logic needs some additional work. What about the case when a user is attempting to audit a some user's public repos?

cafeliker: @zricethezav thanks for sharing your comments, and I totally agree with you.

I just added the logic that the tool would scan both public and private repo for the authenticated user but only public repo for another user.

cafeliker: @zricethezav could you pls help review my changes? many thanks.

cafeliker:
@zricethezav do you have any more comments before merge this PR?

cafeliker:
@zricethezav I am seeing you have already made the fix of reading the environment variable GITHUB_TOKEN for repo clone activity in repo.go for the last commit, so I synced with your master, and changed the PR title to realign the purpose with the code change.

zricethezav:
https://github.com/zricethezav/gitleaks/blob/master/hosts/github.go#L72-L81 same logic here. Sorry this wasn't merged. The code went through a couple refactors. I appreciate you opening a PR though!

PR #224, opened 2019-05-28
Solved bug on error being overwritten and not passed to postAudit, closes #223

### LINKED ISSUE #223: Errors not reported properly
In core.go, Run(), the variable err has no value when the postAudit label is reached.  That's because the variable is assigned within the scope of the if statement for each call like repoInfo, err := newRepoInfo() on line 65 so the method scoped variable err is void.

eargollo:
The err variable is being overwritten and it reaches postAudit always as nil.

You can test this by running:
```
$ gitleaks --repo-path ../does-not-exist
?[36mINFO?[0m[2019-05-28T15:37:59-07:00] opening ../does-not-exist
?[31mERRO?[0m[2019-05-28T15:37:59-07:00] unable to open ../does-not-exist
?[36mINFO?[0m[2019-05-28T15:37:59-07:00] 0 leaks detected. 0 commits inspected in 1 second 505 milliseconds
```
With this patch it start capturing the error:
```
$ go run . --repo-path ../does-not-exist
?[36mINFO?[0m[2019-05-28T15:39:02-07:00] opening ../does-not-exist
?[31mERRO?[0m[2019-05-28T15:39:02-07:00] unable to open ../does-not-exist
?[31mERRO?[0m[2019-05-28T15:39:02-07:00] repository does not exist
exit status 2
```

This PR closes #223 .

### DIFF
--- a/main.go
+++ b/main.go
@@ -2,14 +2,19 @@ package main
 
 import (
 	"os"
+	"strings"
 
 	log "github.com/sirupsen/logrus"
-	"github.com/zricethezav/gitleaks/src"
+	gitleaks "github.com/zricethezav/gitleaks/src"
 )
 
 func main() {
 	report, err := gitleaks.Run(gitleaks.ParseOpts())
 	if err != nil {
+		if strings.Contains(err.Error(), "whitelisted") {
+			log.Info(err.Error())
+			os.Exit(0)
+		}
 		log.Error(err)
 		os.Exit(gitleaks.ErrExit)
 	}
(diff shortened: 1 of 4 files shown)


eargollo:
@zricethezav, all test errors are related to non existing repositories `"TestRun test empty"`. 

This code gets an `"authentication required"` early on instead of an `"reference not found"`.

This error was being overwritten before and the output of the program was as if it had scanned the repository. 

The error comes from the call (line 101 of `src/repo.go`):
```
			repo, err = git.Clone(memory.NewStorage(), nil, &git.CloneOptions{
				URL:      opts.Repo,
				Progress: os.Stdout,
			})
```

PR #238, opened 2019-07-04
Multiple new features (see Description)

Zekii:
Hi Zachary, and thank you for this awesome project :)

I decided to learn Golang and work on Gitleaks with the approval of my company, during a dedicated time for cybersecurity projects. All features that i developed are separated by commit with a description, feel free to comment on them too.

## Context

Only a long format `--context`, takes a integer and disabled by default.
It is an option to get a given number of characters before and after the rule match. I wanted to have a more readable and comprehensible leak for a match. It happens that a leak is in a line too long or too short.

I do not have one for the "too long", but it can happen on a very long json (not formatted) or a compressed project code.
And for the "too short"

```
func locate_something(x: int, \
                      y: int, \
                      api_key: string) # Only this line would be matched
```
In this case, we need context, because it doesn't mean anything alone.

## Date limit

Only a long format `--date-limit`, takes a string in `YYYY/MM/DD` date format.
Pretty straightforward, set a date limit to the search.

## Additions only

Only a long format `--adds-only`, a boolean option and disabled by default.
Just scan the additions to the repository, since every deletions are ex-additions. I saw an issue suggesting that feature (that you approved). Don't know if you're already on it, but wanted to push mine too since i did it.
 
## Filtetype scan on binaries

Extend the scan on binary files too. Even if Gitleaks cannot read the file, if the filepath match a rule, it still can have leaks in it. If the report is read by a human at the end, he would be able to read it manually anyway.
Oh and the filetype scan was not done in the parent commit.

## Before & After

Only a long format `--before` & `--after`, both takes a string and disabled by default.
An option to add text around a leak in the report, in a previous example i talked about huge lines that could potentially pollute the report. In another way, even with a fixed number of characters, the rule match can be hard to find. So i added these options to locate a match easily by adding text around it.

To give an example, we can add html tags to highlight them, if they iare going to be displayed on a web page. We can also add some fire emoji around for an notification by text message.

An example for the html tags:

```
func locate_something(x: int, \
                      y: int, \
                      <div class="gitleaks">api_key</div>: string)
```

## Persistent repository

Only a long format `--persist`, takes a string to a destination folder for the cloning, disabled by default.
An option to keep a scanned repository in the disk.

Can helps to avoid a cloning step, that can be very long depending of the project (and network).

---

So.. Yeah, that's it haha

I'm totally new to Golang, so i'm not sure i did it the right way, but would love to have review on it ! Hope you'll at least keep some features :sweat_smile:

I'll add a last commit to update the changelog depending on the kept features :)

Tony

EDIT: First participation on open-source project, hope i did it right haha


cfzlp:
This is the first time a make a PR to an open source project. And I am not a Go programmer so if I didn't follow "proper protocol" or if anything is amiss with my PR, please bare with me and accept my apologies.
Thank you for this great tool!

bufferoverflow:
I recommend to create one pull request per feature, this makes handling for them maintainer and other contributors easier.

bufferoverflow:
@Zekii I suggest to use an iso8601 date format such as `2019-07-14T23:34:30` the date-limit function

zricethezav:
@Zekii I'll echo what @bufferoverflow suggested. Splitting these features into separate PRs would be nice.

Zekii:
Hey @bufferoverflow @zricethezav !

Sorry i was out for a long time, and that's fine for me, i'll split them then :)

PR #246, opened 2019-08-02
Regex whitelist example in sample config

hughsaunders:
Users looking to the sample config to discover possible options may
not notice that its possible to whitelist by regex. This commit adds
an example to aid discovery.

Related: conjurinc/ops#246


hughsaunders:
Passes when I run locally, and only difference from master (which is passing) are commented out config lines, so failure seems unrelated. Also at least one other PR has the same failure: https://travis-ci.org/zricethezav/gitleaks/builds/566434540

zricethezav:
@hughsaunders thanks for this.

Issue #183, opened 2019-02-28
Feature Request: Specify Github Token as Command-Line Arg


galiagante:
As a companion to Issue 182, there appears to be valid use cases for specifying a Github token via command-line argument.

zricethezav:
`--access-token=` is an option in the current version. I believe this was added in v3.0

PR #267, opened 2019-10-04
Adds Github personal access token command-line argument (#183)

timetoplatypus:
Adds the feature requested in issue #183.

This change is purely additive, meaning there shouldn't be any functional change to how the program is executed; it simply adds a new way for users to provide the Github personal access token.

With respect to the code, the only real change in logic is that instead of calling `os.Getenv("GITHUB_TOKEN")` every time gitleaks needs the token, it simply fetches the value from the `opts` struct that is populated at program start. The `opts.GithubToken` field will use the `GITHUB_TOKEN` environment variable if, and only if, the user did not provide the token via command-line argument.


timetoplatypus:
@zricethezav Can I get some feedback on this PR? It won't merge cleanly now, but it would have before the codebase was reorganized and the `src` directory was removed.

If there's something wrong with the changes, let me know and I can fix it so this can be merged. If this isn't a wanted feature, can we document that here and close the PR?

amrityam:
Does this access token will for AzureDevOps git repositories.
Because I am facinf issue both for public and private AzureDevOps git repositories.

Created a Bug for that: https://github.com/zricethezav/gitleaks/issues/440

zricethezav:
`--access-token` is now an option

PR #268, opened 2019-10-07
Add option to inspect only additions from diffs

Zekii:
Hi Zach !

It's been a long time, with the 30 new commits since my last PR i did not wanted to rebase them. But it was easier than i thought so.. Here i am :)

For my first PR i did a one line modif to add an option `--adds-only` to only inspect additions in diffs.

My first "real" PR so feel free to tell me if i something is wrong.


src/repo.go
```diff
@@ -311,7 +311,7 @@ func (repo *Repo) audit() error {
 					}
 					chunks := f.Chunks()
 					for _, chunk := range chunks {
-						if chunk.Type() == diffType.Add || chunk.Type() == diffType.Delete {
+						if (!opts.AdditionsOnly && chunk.Type() == diffType.Add) || chunk.Type() == diffType.Delete {
```
avishnyakov: @Zekii @zricethezav can we also propagate `chunk.Type()` value into json report?

That way, every leak will have additional property indicating where it came from - from addition or deletion.
Can be useful for analytics on top of standard gitleaks reports.

Zekii: Sorry about the late response, i'll check if it has been done by someone else already. But if not, i'll do it for sure, that's an nice data to have in reports ! :)

Zekii: Finally, after a month, i added the diffType to the report. Yaaay !

I noticed that the option `github-pr` doesn't only search in modified chunks but in the commit content, which contains the context (3 lines after and before the chunk). For example, if we take one of my commit:
```
 // Leak represents a leaked secret or regex match.
 type Leak struct {
        Line     string    `json:"line"`
+       DiffType string    `json:"diffType"`
        Commit   string    `json:"commit"`
        Offender string    `json:"offender"`
        Rule     string    `json:"rule"`
```

And a basic rule "Offender" would match in this case :(

First things first, i decided to not fix it since it is not the point of my PR.

Incisive:
Yes, so excited for this 🙌!

Zekii:
Hi,

A lot happened since my last update, so i decided to rebase my branch first :)

About @avishnyakov question/suggestion:
```
@Zekii @zricethezav can we also propagate chunk.Type() value into json report?

That way, every leak will have additional property indicating where it came from - from addition or deletion.
Can be useful for analytics on top of standard gitleaks reports.
```

I am currently working on it, just have to update all tests (since i am modifying the `Leak` structure)

Zekii:
Done, do I have to do something to validate it ?

Zekii:
Rebased my pull request, i very much like the new changes :)

Nklya:
Are there any news on this PR?
It's a really nice option.

Zekii:
No idea, sorry :/ Resolved the conflicts few times, but can't seem to merge it, and i already re-requested reviews. Maybe i should create a new PR ?

zricethezav:
https://github.com/zricethezav/gitleaks/releases/tag/v5.0.0 this feature has been released in v5.0.0. The default behavior now is to scan only additions. Deletions can be optionally scanned with `--include-deletions`

PR #276, opened 2019-11-06
add rule for DSA keys

nijikon:
Just like for `RSA`, introduce a rule to look for `DSA` keys.


zricethezav:
This is part of the default config

Issue #265, opened 2019-09-19
Missing option to ignore certificate issues


gkroon:
Please consider the below example:

```
$ gitleaks -r https://foo.bar/foo/bar --ssh-key=~/.ssh/id_ed25519
INFO[2019-09-19T17:47:10+02:00] cloning https://foo.bar/foo/bar 
ERRO[2019-09-19T17:47:11+02:00] Get https://foo.bar/foo/bar/info/refs?service=git-upload-pack: x509: certificate is valid for bar.foo, not foo.bar
```

It's nice to see that the certificate is being checked for (common) errors. But would it not be an idea to add an option that ignores certificate issues like this? I'm not asking to change the default behaviour, but just an option to override this default _and correct_ behaviour.

rgmz:
If you are using macOS you can work around this by adding those certificates to your keychain and trusting them; Go/Gitleaks should respect the system keychain settings.

But I agree that it would be nice to have an option that passes through the underlying Git/GitHub client contexts. 👍

technopwnsauce:
Go-git doesn't directly provide this option, but they have documented a way to define a custom http.Client to accept insecure certificates.

https://github.com/src-d/go-git/issues/817

technopwnsauce:
@gkroon @rgmz Are there any public repos with invalid certs you know off-hand I could use for testing?

gkroon:
@technopwnsauce You could use any public git repo with your local burp or zaproxy in between. Without importing their local CA cert, this will show up an invalid (untrusted) certificate warning by default.

rgmz:
> @gkroon @rgmz Are there any public repos with invalid certs you know off-hand I could use for testing?

I don't know any off-hand — we have a corporate proxy, so everything is invalid.

:+1: to this:
>  You could use any public git repo with your local burp or zaproxy in between.

zricethezav:
gitleaks is not responsible for cloning anymore

Issue #258, opened 2019-09-05
Positive look ahead support or multiple regexps per rule?


avishnyakov:
Looks that indernearh regexp library does not support feature like positive look-ahead?
Trying to use anything like (?=regex) yields exception during compilation.
> https://stackoverflow.com/questions/30305542/using-positive-lookahead-regex-with-re2

Any ways to support more advanced regexp support?

Other way around is to support array of regexps per rule with AND / OR logic.
For example, handling 32 long tokens which kind of look random.

```toml
[[regexes]]
description = "Catching this AND this AND that"
regex = [
 "has_to_be_string_of_32_chars",
 "has_to_have_at_least_one_number",
 "has_to_have_at_least_one_upper_case_letter",
 "has_to_have_at_least_one_lower_case_letter",
]
```

Similar approach can be used for OR match across several regexps. 
This is helpful to catch general tokens all across the codebase without particular format - in hardcoded strings within the code or configurations.

rgmz:
> Looks that indernearh regexp library does not support feature like positive look-ahead?
> Trying to use anything like (?=regex) yields exception during compilation.
> 
> > https://stackoverflow.com/questions/30305542/using-positive-lookahead-regex-with-re2

The same is true for look-behind; my understanding is that those features cannot be safely compiled, and thus are deliberately not supported.

> Other way around is to support array of regexps per rule with AND / OR logic.
> For example, handling 32 long tokens which kind of look random.
> 
> ```toml
> [[regexes]]
> description = "Catching this AND this AND that"
> regex = [
>  "has_to_be_string_of_32_chars",
>  "has_to_have_at_least_one_number",
>  "has_to_have_at_least_one_upper_case_letter",
>  "has_to_have_at_least_one_lower_case_letter",
> ]
> ```

This is an interesting solution. How would you differentiate AND vs OR in this fashion? 

Not to clutter the syntax, but perhaps something explicit like `regex_any_of = []` for OR and `regex_all_of = []` for AND?

avishnyakov:
@rgmz good point. Agree, multiple regexp per rule should be coming with ability to define matching pattern. This is a challenge for GitLeaks and number of other similar projects out there.

There might be a better way to handle rules configuration.

Why not to consider YARA rules? The pattern matching swiss knife BY VirusTotal used by many security companies.
* https://github.com/virustotal/yara
*  http://virustotal.github.io/yara/

Leaning towards industry standards might be a good idea. Surely, pros and cons but...
[...]

Issue #257, opened 2019-09-05
Ability to define a folder with *.toml configirations


avishnyakov:
Following Ability to define positive/negatives datasets for rules? #256, having a folder with *.toml files might be very helpful.

Organizing dozens rules in the same *.toml is not fun.
Adding positive/negative dataset might expand rule significantly. At the end, file name might be a good indicatino of the rule, and then the *.toml itself would have all we need - the rule, tags, metadata and regression datasets.

Right now it is a challenge to organize and author dozen rules within one file, not to mention handling regression and assertions against positive/negative datasets and whitelistings.

zricethezav:
I don't have plans to support this. I like the simplicity of the current configuration

Issue #255, opened 2019-09-05
Whitelisting per-rule?


avishnyakov:
v2 config looks good. 
https://github.com/zricethezav/gitleaks/wiki/config-toml

Would it be useful to have white-listing per rule?
Right now, white-listing is global - for both files and strings. It is too broad, would be much nicer to have per-rule whitelisting so that final .toml file does not become fat and unmanageable.

rgmz:
I was just thinking about this, in addition to per-filetype whitelists.

Contrived examples:
**Per-Rule Whitelist**
```toml
[[rules]]
description = "Generic client secret"
regex = '''(?i)client(_|-)?secret(.{0,20})?['|"][0-9a-zA-Z]{32,45}['|"]'''
tags = ["key", "API", "generic"]
[[rules.whitelist]]
# $VENDOR provides everyone the same hard-coded client secret.
# (Yes, this actually happens)
regexes = ["AA97B177-9383-4934-8543-0F91A7A02836"]
```

**Per-Filetype(s) Whitelist**
```toml
[[whitelist.rules]]
description = "Ignore common Jenkinsfile methods)"
regexes = ["withCredentials", "usernamePassword", "usernameColonPassword"]
filetypes = ['''(.*)?Jenkinsfile$''']
```

zricethezav:
This is a good idea, I'll try to get this into my next release

zricethezav:
Added whitelists per rule in 3.0, check out https://github.com/zricethezav/gitleaks/tree/master/examples and https://github.com/zricethezav/gitleaks/tree/master/test_data/test_configs for examples @avishnyakov @rgmz

zricethezav:
Closing this issue

PR #278, opened 2019-11-07
feat/report-writer: supporting AWS S3 and syslog, and multiple report outputs

mtulio:
Supporting new report outputs to:
- AWS S3 bucket
- syslog remote server
- multiple outputs

## AWS S3 bucket

The output object will be saved in **JSON** format directly to the object path in argument --report.  See more on README

## Syslog

The output object will be saved in **JSON** sending the payload, by leak result, to Sylog server. See more on README

## Multiple output

Now, to keep compatibility and improve the report options, multiple outputs could be informed spited by comma. See more on README

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@
 
 examples
 src/gitleaksConfig
+build/
--- a/Makefile
+++ b/Makefile
@@ -1,18 +1,27 @@
-.PHONY: test build-all deploy
+.PHONY: test build-all build-deps build-os deploy
 
 test:
 	go get golang.org/x/lint/golint
 	go fmt
 	golint
 	go test --race --cover github.com/zricethezav/gitleaks/src -v
+
 deploy:
 	@echo "$(DOCKER_PASSWORD)" | docker login -u "$(DOCKER_USERNAME)" --password-stdin
 	docker build -f Dockerfile -t $(REPO):$(TAG) .
 	echo "Pushing $(REPO):$(COMMIT) $(REPO):$(TAG)"
 	docker push $(REPO)
-build-all:
+
+build-deps:
 	rm -rf build
 	mkdir build
+
+GOOS ?= linux
+GOARCH ?= amd64
+build-os: build-deps
+	env GOOS="$(GOOS)" GOARCH="$(GOARCH)" go build -o "build/gitleaks-$(GOOS)-$(GOARCH)"
+
+build-all: build-deps
 	env GOOS="windows" GOARCH="amd64" go build -o "build/gitleaks-windows-amd64.exe"
 	env GOOS="windows" GOARCH="386" go build -o "build/gitleaks-windows-386.exe"
 	env GOOS="linux" GOARCH="amd64" go build -o "build/gitleaks-linux-amd64"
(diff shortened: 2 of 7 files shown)


avishnyakov:
Nice work but there is a question if this is something GitLeaks should do. Looks like a feature creep.

Expanded features increases complexity and testing. For example, these are exact features which probably should not even be in the GitLeaks in the first place. Why to bother HOW the git repository is delivered for scanning, supporting dozens of scenarios with increasingly complex testing? Let people define and implement that on their own pace, why not?
  
* Hit error when trying to scan github enterprise using user #245
* Can't scan an org #220
* Missing option to ignore certificate issues #265
* There is an error trying to connect to ssh in the --github-user option. #277

As for this PR, why not to transform current output outside of GitLeaks and send it to s3, Azure, Google, syslog or other systems? For example, GitLeaks can be run in docker container and then anther docker container can handle output: post-process, enhance with more data and send to other systems. Not that many lines in any programming language, isn't it? That way, GitLeaks does one things, and does it REALLY great avoiding feature creeps. Just keep it simple. 

Meanwhile, GitLeaks has a room for improvement in very core functionality. Not that this one if not important, but these are the gaps in actual scanning capabilities.

* Positive look ahead support or multiple regexps per rule? #258
* More tracing for findings and rules? #254
* Only scan additions, ignore removals #207
* Whitelisting per-rule? #255
* Ability to define positive/negatives datasets for rules? #256 
* Ability to define a folder with *.toml configirations #257
* and others

cc @zricethezav

mtulio:
Hi @avishnyakov,

Thanks for your feedback.

I agree with you that is out of the scope of gitleaks to support different “backends”.

It was a result of an improvement that I identify, and I had to develop after writing too many lines in the shell script to:
- format json output lines
- too many string transformation to pipe the result to logger command
- taking too long time waiting for execution, sometime dozens hours waiting for an result, takes me necessity to streaming results to external syslog processor (ELK, Graylog, etc), and then process to create some metrics and alarms.

I understand that has a lot of improvements and issues, mainly when we need to run in thousands of repos a day (taking more than a day), but I just wanted to give back the local implementation for this project that is helping us to improve the security of our repos.

Thanks.,
Marco

avishnyakov:
@mtulio agree on everything said.

I also work at a scale of ~1000 repositories processed daily. They come from different systems, and then GitLeaks output is enhanced, post-processed and pushed to analytics systems. GitLeaks outputs its raw data, and then custom scripts take over enhancing data, post-processing and pushing to data storage and analytics systems.

Definitely, there is an opportunity to contribute. The question is how to get it right. I am very much interested in your work around Elasticsearch and Syslog server. From my end, I can help with Azure storage and Azure Application Insights. 

Maybe we can figure out a small contrib project around GitLeaks output? I also have several more ideas which aren't in issues yet, keen to get your feedback.

mtulio:
Hi @avishnyakov,

It sounds good. I am currently using Graylog (that is also using ES as storage) as log processor from syslog raw data. I started the project last week and, as we’ve talked, we can explore too much more from this raw data. Now I am carrying about to accurate and dedup of issues found on every output.

It will be a pleasure to discuss/contrib to this project.

PR #281, opened 2019-11-13
V3

zricethezav:


zricethezav:
TODO: 
- timeout option
- depth option
- commit-stop option

PR #297, opened 2019-11-25
Updating generic credentials regex

petegallagher:
I found the default regex for the generic credentials was not able to find as many secrets as I anticipated. I've tried to make some improvements to this, let me know what your thoughts are.

**Features:**
1. Extend keywords to include `token`.
2. Be more prescriptive around what characters can surround the keywords (i.e. `\.` matches any character, but `[\w-_]` should is a more restrictive character set that should match valid variables).
3. Require an assignment operator i.e. `[=:]`.
4. Match the secret contents within either single quotes or double quotes i.e. `('(?:[^'\\]|\\.){6,100}'|"(?:[^"\\]|\\.){6,100}")`. This matches secrets between 6-100 characters that are within opening/closing quotes of the same character. 

As an example the following patterns would be matched:

```
my_apikey = 'test123'
api_key_service      : "test123"
'defaulttoken'  =  "example\"blah''123"
"hidd3n_secret_field":    'unique secret-_120983@£'
```

### DIFF
--- a/config/default.go
+++ b/config/default.go
@@ -68,7 +68,7 @@ title = "gitleaks config"
 
 [[rules]]
 	description = "Generic Credential"
-	regex = '''(?i)(api_key|apikey|secret)(.{0,20})?['|"][0-9a-zA-Z]{16,45}['|"]'''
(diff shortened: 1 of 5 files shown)


petegallagher:
I've been doing some testing with this and it might be raising too many false positives. I'll do some more testing and come back to you on this one.

PR #298, opened 2019-11-25
Whitelist default AWS Example keys

petegallagher:
I noticed that the AWS example key is mentioned in a few places in your docs. But I was curious why this was not whitelisted by default in the tool? I've included a PR for this but I assume you're already not including this for a good reason, so you can close this one if you want, but it would be good to understand why this is.


zricethezav:
Closing this out as I don't want to include a default rule specific to the gitleaks repo

PR #301, opened 2019-11-28
Fix repo-config and file whitelist for AuditUncommitted function

petegallagher:
The audit uncommitted function does not honour the `--repo-config` flag or the `whitelist` file section of the config file. This MR fixes these issues.

I think this fix would be better by refactoring the code base so both audit functions use a single logical function, rather than copying it into 2 different places. I will try and look at this when I get a bit more time. But for now this should be an improvement.

### DIFF
--- a/audit/repo.go
+++ b/audit/repo.go
@@ -104,6 +104,11 @@ func (repo *Repo) AuditUncommitted() error {
 	if err != nil {
 		return err
 	}
+	c.Hash = plumbing.Hash{}
+	c.Message = "STAGED CHANGES"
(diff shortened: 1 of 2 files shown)


zricethezav:
Awesome! Thanks for the contribution, could you update the unittest test data file to reflect the changes in this PR?

petegallagher:
Updated, sorry I didn't realise it had failed the tests.

Is it possible to push a new release (3.0.3) after this is merged? I'm very keen to get this rolled out to the rest of the team.

Many thanks,

Peter

PR #303, opened 2019-12-01
Adding baseurl flag to allow self-hosted GitLab/GitHub servers

petegallagher:
I have tested this with a self-hosted GitLab instance. I have not tested this with a GitHub Enterprise installation, but it *should* work...

### DIFF
--- a/hosts/github.go
+++ b/hosts/github.go
@@ -27,7 +27,8 @@ type Github struct {
 
 // NewGithubClient accepts a manager struct and returns a Github host pointer which will be used to
 // perform a github audit on an organization, user, or PR.
-func NewGithubClient(m manager.Manager) *Github {
+func NewGithubClient(m manager.Manager) (*Github, error) {
+	var err error
 	ctx := context.Background()
 	token := oauth2.StaticTokenSource(
 		&oauth2.Token{AccessToken: options.GetAccessToken(m.Opts)},
@@ -39,7 +40,6 @@ func NewGithubClient(m manager.Manager) *Github {
 	if m.Opts.BaseURL == "" {
 		githubClient = github.NewClient(httpClient)
 	} else {
-		var err error
 		githubClient, err = github.NewEnterpriseClient(m.Opts.BaseURL, m.Opts.BaseURL, httpClient)
 		if err != nil {
 			log.Error(err)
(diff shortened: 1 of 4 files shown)


hosts/github.go
```diff
@@ -33,9 +33,22 @@ func NewGithubClient(m manager.Manager) *Github {
...
+		githubClient = github.NewClient(httpClient)
+	} else {
+		var err error
+		githubClient, err = github.NewEnterpriseClient(m.Opts.BaseURL, m.Opts.BaseURL, httpClient)
+		if err != nil {
+			log.Error(err)
+		}
+	}
+
 	return &Github{
```
zricethezav: This all looks good but let's add a nil check here https://github.com/zricethezav/gitleaks/blob/master/hosts/host.go#L31. Something like 
```
	if host == nil {
		return fmt.Errorf("unable to create %s client", getHost(m.Opts.Host))
	}
```

zricethezav: Or we could change the function signature to return an error

petegallagher: Yeah, makes a lot more sense to return the errors.

hosts/github.go
```diff
@@ -27,16 +27,29 @@ type Github struct {
...
 
+	var githubClient *github.Client
+	httpClient := oauth2.NewClient(ctx, token)
+
+	if m.Opts.BaseURL == "" {
+		githubClient = github.NewClient(httpClient)
+	} else {
+		githubClient, err = github.NewEnterpriseClient(m.Opts.BaseURL, m.Opts.BaseURL, httpClient)
+		if err != nil {
+			log.Error(err)
```
zricethezav: This check could be removed since the error will be bubbled up in main.go

hosts/gitlab.go
```diff
@@ -22,12 +22,23 @@ type Gitlab struct {
...
+	gitlabClient := &Gitlab{
 		manager: m,
 		ctx:     context.Background(),
 		client:  gitlab.NewClient(nil, options.GetAccessToken(m.Opts)),
 	}
+
+	if m.Opts.BaseURL != "" {
+		err := gitlabClient.client.SetBaseURL(m.Opts.BaseURL)
+		if err != nil {
+			log.Error(err)
```
zricethezav: This check could be removed since the error will be bubbled up in main.go

hosts/github.go
petegallagher: Updated

hosts/gitlab.go
petegallagher: Updated

Issue #306, opened 2019-12-02
Docker automated builds


petegallagher:
I noticed the last time the [Docker Hub image](https://hub.docker.com/r/zricethezav/gitleaks) was updated was 2 months ago.

![Screenshot 2019-12-02 at 13 27 54](https://user-images.githubusercontent.com/688331/69963194-94957a80-1507-11ea-83bb-7a92c9e66a72.png)

Is it possible to setup automated builds? The screenshot below shows the required config for both a "latest" (i.e. built from master) and a released version.

![Screenshot 2019-12-02 at 13 19 36](https://user-images.githubusercontent.com/688331/69962709-78450e00-1506-11ea-8d7e-273b18f2b89d.png)

| Source Type | Source | Docker Tag | 
| -------------|--------|-------------|
| Branch | `master` | `latest` |
| Tag | `/^v([0-9.]+)$/` | `release-{\1}` |

petegallagher:
@zricethezav is there anything else I can do to help this one? Happy to configure this for you if want to make [me](https://hub.docker.com/u/petergallagher) a collaborator on Docker Hub for this project.

prabhu:
Looks like the author is struggling to deal with the popularity of this library :(

zricethezav:
I'll get on it this week

zricethezav:
https://hub.docker.com/r/zricethezav/gitleaks/tags @prabhu @petegallagher there ya go. I'll be updating the makefile/dockerfile next release

petegallagher:
Awesome, thank you.

PR #313, opened 2019-12-18
adding docker push support again to Make and fixed a go vet nit

zricethezav:


petegallagher:
What is the reason for including Docker push as part of the Make file? I thought the Docker builds were configured automatically now based on #306 ?

zricethezav:
Releases aren't automated in any way, I manually build and push them with the Makefile. Main reason is there are less moving parts and when I make a release I have to manually type up the changes and description so adding a docker push to that process is not a big deal.

PR #316, opened 2020-01-10
Fix empty repo fields when running organization scan

jporzuceksfx:
When running organization scan, all values of `repo` field were empty (fille with dot ".") because `repo.Manager.Opts.Repo` was empty. Now when cloning we chck if scan is run for organization and if so, don't set `repo.Name` which is already set correctly.

### DIFF
--- a/audit/repo.go
+++ b/audit/repo.go
@@ -74,8 +74,10 @@ func (repo *Repo) Clone(cloneOption *git.CloneOptions) error {
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


lghakamo-paf:
Nice, I was looking at fixing this myself and thus logged https://github.com/zricethezav/gitleaks/issues/335

lghakamo-paf:
This is fixed with https://github.com/zricethezav/gitleaks/pull/344.

zricethezav:
I am no longer supporting gitlab/github integrations

PR #333, opened 2020-02-12
Enhance entropy

NoelAlgora:
### Description:
It does not make a lot of sense to calculate entropy on all the line when you are matching a regex, even more i dont think it makes a lot of sense to create rules for calculating the entropy on lines without any regex.

I have changed the implementation on how entropy works so it only gets calculated for the matching regex (offender), it is also possible to only calculate the entropy on a specific capture group.

Example:
```
[[rules]]
	description = "entropy and regex"
	regex = '''secret(.{0,20})['|"]([0-9a-zA-Z-._{}$\/\+=]{20,120})['|"]'''
	tags = ["entropy"]
		[[rules.Entropies]]
			Min = "4.5"
			Max = "4.7"
```
If we run gitleaks with this config against:

```
aws_secret_access_key='ABCDEF+c2L7yXeGvUyrPgYsDnWRRC1AYEXAMPLE'
```
will calculate the entropy of all the matching regex which is:
```
secret_access_key='ABCDEF+c2L7yXeGvUyrPgYsDnWRRC1AYEXAMPLE'
```

We can set on which group the entropy will get calculated to only calculate it on the token:
```
[[rules]]
	description = "entropy and regex"
	regex = '''secret(.{0,20})['|"]([0-9a-zA-Z-._{}$\/\+=]{20,120})['|"]'''
	tags = ["entropy"]
		[[rules.Entropies]]
			Min = "4.5"
			Max = "4.7"
			Group = "2"
```
Now iff we run gitleaks with this config against the same secret entropy will get only calculated on the second capture group which is:
```
ABCDEF+c2L7yXeGvUyrPgYsDnWRRC1AYEXAMPLE
```

### Checklist:


config/config.go
```diff
@@ -67,7 +68,11 @@ type TomlLoader struct {
 		Description string
 		Regex       string
 		Tags        []string
-		Entropies   []string
+		Entropies   []struct {
+			Min		string
+			Max		string
+			Group	string
```
zricethezav: `string` indentation needs lining up

config/config.go
```diff
@@ -130,17 +135,41 @@ func (tomlLoader TomlLoader) Parse() (Config, error) {
...
+			min, err := strconv.ParseFloat(e.Min, 64)
+			if err != nil {
+				return cfg, err
+			}
+			max, err := strconv.ParseFloat(e.Max, 64)
+			if err != nil {
+				return cfg, err
+			}
+			if(e.Group == "") {
+				e.Group = "0"
```
zricethezav: why assign to "0"?

zricethezav:
This is awesome work! Just needs one small formatting change then it's good to go.

Cedric-Magnan:
⚠️ [Warning]
This modification of the leaky-repo.toml file makes the gitleaks docker image return the following error :
ERRO[2020-02-26T15:57:26Z] toml: cannot load TOML value of type map[string]interface {} into a Go string

zricethezav:
ah, good catch. The example tomls will need to be updated. Would you mind opening an issue?

PR #340, opened 2020-02-21
Add file regex support

NoelAlgora:
### Description:
This pull request is dependant of the pull request "Enhance entropy"  https://github.com/zricethezav/gitleaks/pull/333

I have modified the rules with the objective to make them more granular. I have added two new fields to the rule: fileNameRegex and filePathRegex. The rule will only search the content regex (the standard one) if the file gets matched with the fileNameRegex and filePathRegex in case they exists. If there are fileNameRegex and/or filePathRegex but no content regex, file will be marked as a leak if it matches filename and path regexes.

Look at the following examples:

Leak based on filename:
```
[[rules]]
	description = "Potential cryptographic private keys"
	fileNameRegex = '''\.(?i)(pem|pkcs12|p12|pfx|asc|private_key|priv|pub|jks|key(pair))'''
	tags = ["filename", "crypto-keys"]
```
Leak based on path:
```
[[rules]]
	description = "SSH configuration file"
	filePathRegex = '''\.?(?i)ssh'''
	tags = ["filepath", "ssh"]
```
Leak based on filename and regex:
```
[[rules]]
	description = "Yaml config files"
        fileNameRegex = '''\.ya?ml'''
	filePathRegex = '''config(uration)?'''
	tags = ["filename"]
```
I have also changed the type of the parameter file of the global whitelists to an array, and added another new field called paths that is also an array so you can set multiple regex for filenames and paths that should be whitelisted for every rule.

Example:
```
[whitelist]
	description = "Whitelisted files"
	files = [
		'''(.*?)\.(jpg|jpeg|png|gif|bmp||bin)$''',
		'''(.*?)\.(tiff|tif|psd|doc|pdf)$''',
		'''(.*?)\.(bin)$''',
	]
	paths = [
		'''translations\/messages'''
	]
```
I have also added a new field called path to the Rule.Whitelist structure so now you can also set certain paths to be whitelisted from a specific rule. I have modified a bit how whitelists work so a leak will be skipped if filename matches file regex or if path matches path regex or if the OFFENDER matches the regex.

Example of whitelist for yml and yaml files:
```
	[[rules.whitelist]]
		description = ""
		regex = '''\s*:\s*["']?(password|(in)?valid|12345)["']?$'''
		file= '''|.ya?ml$'''
```

I have created a new testing repo and new tests for the new functionalities.

### Checklist:


NoelAlgora:
Looks there was a problem with a new repo i made for testing, gonna fix it

zricethezav:
Great work! Thank you for a thorough description. Before diving into the code, question about this:
```
# Leak based on filename and regex:

[[rules]]
	description = "SSH configuration file"
	filePathRegex = '''\.?(?i)ssh'''
	tags = ["filename"]

```
To me this looks like the leak is based soley on the file path and not filename and regex
cc @NoelAlgora

config/config.go
```diff
@@ -165,37 +180,32 @@ func (tomlLoader TomlLoader) Parse() (Config, error) {
...
 			return cfg, fmt.Errorf("problem loading config: %v", err)
 		}
-		cfg.Message = re
+		cfg.Whitelist.Files = append(cfg.Whitelist.Files, re)
 	}
 
-	// global whitelists
-	if tomlLoader.Whitelist.File != "" {
-		re, err := regexp.Compile(tomlLoader.Whitelist.File)
+	// global file whitelists
```
zricethezav: // global file path whitelists

examples/regex_and_file.toml
```diff
@@ -6,5 +6,7 @@
     regex = '''(A3T[A-Z0-9]|AKIA|AGPA|AIDA|AROA|AIPA|ANPA|ANVA|ASIA)[A-Z0-9]{16}'''
     tags = ["key", "AWS"]
 
-[Global]
+[[rules]]
```
zricethezav: much cleaner 👍

test_data/test_configs/bad_aws_key_message_regex.toml
```diff
@@ -7,7 +7,3 @@
     description = "AWS Manager ID"
     regex = '''(A3T[A-Z0-9]|AKIA|AGPA|AIDA|AROA|AIPA|ANPA|ANVA|ASIA)[A-Z0-9]{16}'''
     tags = ["key", "AWS"]
-
-[Global]
```
zricethezav: I would think this would cause the tests to break? If not then we should probably remove this test config as it is a valid config now

zricethezav:
Just a couple of comments. Mostly just nits and some clarification. Overall this looks ready to ship 👍. Thanks a ton for opening this!

test_data/test_configs/bad_aws_key_message_regex.toml
NoelAlgora: I dont think its necessary this test anymore either but this should have caused the tests to break but it didnt so maybe its something going on with config_tests? I will remove this test

NoelAlgora:
> 
> 
> Great work! Thank you for a thorough description. Before diving into the code, question about this:
> 
> ```
> # Leak based on filename and regex:
> 
> [[rules]]
> 	description = "SSH configuration file"
> 	filePathRegex = '''\.?(?i)ssh'''
> 	tags = ["filename"]
> ```
> 
> To me this looks like the leak is based soley on the file path and not filename and regex
> cc @NoelAlgora



> 
> 
> Great work! Thank you for a thorough description. Before diving into the code, question about this:
> 
> ```
> # Leak based on filename and regex:
> 
> [[rules]]
> 	description = "SSH configuration file"
> 	filePathRegex = '''\.?(?i)ssh'''
> 	tags = ["filename"]
> ```
> 
> To me this looks like the leak is based soley on the file path and not filename and regex
> cc @NoelAlgora

You are right it looks like i got confused while writing the explanation of the pull request, i have fixed the description by now.

config/config.go
NoelAlgora: Fixed!

test_data/test_configs/bad_aws_key_message_regex.toml
NoelAlgora: All tests were passing but indeed this should have made config tests to break... I have removed the test as there is no point to keep it anymore but maybe we should take a look closer of what happened there

zricethezav: Actually I take that back, this should not cause the tests to break since `Global` is being deprecated

Issue #335, opened 2020-02-13
Report only states repo as "." when scanning an organisation


lghakamo-paf:
**Describe the bug**
The report when scanning repos in an organisation only states dot as repo.

**To Reproduce**
Scan all repos in an organisation.

**Expected behavior**
Repo should be reported as <repo-name>.git
This worked in versions < 3.x. Scanning a single repo reports the expected repo correctly.

**Basic Info (please complete the following information):**
 - OS: MacOS 10.15.3
 - Gitleaks Version: 3.3.0 (Docker)

**Additional context**
N/A

cc @zricethezav

zricethezav:
Thanks for bringing this up. I noticed that this past weekend too actually. Great first issue for anyone who wants to open a PR.

NdagiStanley:
This issue is done, right?

lghakamo-paf:
> This issue is done, right?

Yeah, seems to be fixed by https://github.com/zricethezav/gitleaks/pull/344. Verified with release 4.1.1 but probably fixed already with 4.0.

PR #344, opened 2020-02-26
Fixed bug where the repo name was removed on a user or org audit.

apccurtiss:
### Description:
Fixes Issue #335. The Repo.Name field is removed during a clone if the '--repo' argument is missing, which happens when scanning a user or organization. This was fixed by setting the name _after_ the clone in hosts/github.go. This is the same pattern that's used in hosts/gitlab.go.


zricethezav:
very nice, great catch. Damn side effects

PR #356, opened 2020-03-18
Improve default pattern configuration

stone-z:
### Description:
From https://github.com/zricethezav/gitleaks/issues/200, makes patterns not require quotation marks around strings and adds "password", "pass", "pw", and "key" to the generic credential pattern, and cleans up some other patterns.

This does result in some duplicates for e.g. AWS keys (since they match the generic credential as well as the AWS rule) but handles most of these lists: [list 1](https://github.com/Shashank-In/TravisLeaks/blob/master/wordlist.txt) [list 2](https://gist.githubusercontent.com/EdOverflow/8bd2faad513626c413b8fc6e9d955669/raw/06a0ef0fd83920d513c65767aae258ecf8382bdf/gistfile1.txt)

### Checklist:


zricethezav:
solid, I agree with this PR. Thanks!

Issue #367, opened 2020-04-14
Incorrect results with the --redact option


steeve85:
**Describe the bug**
Similar issue than #332
When I use the --redact option, gitleaks found less secrets than when I omit this option

**To Reproduce**

```sh
$ go get github.com/zricethezav/gitleaks/v4@latest 
go: github.com/zricethezav/gitleaks/v4 latest => v4.1.1
$ gitleaks --repo-path .
WARN[2020-04-13T17:10:32-07:00] 69 leaks detected. 25 commits audited in 515 milliseconds 466 microseconds 
$ gitleaks --redact --repo-path . 
WARN[2020-04-13T17:10:38-07:00] 65 leaks detected. 25 commits audited in 436 milliseconds 890 microseconds
```

I used https://github.com/Plazmaz/leaky-repo for testing

**Expected behavior**

Number of findings with or without `--redact` should be the same

**Basic Info (please complete the following information):**
 - OS: Archlinux
 - Gitleaks Version: v4.1.1

cc @zricethezav

zricethezav:
fix released in https://github.com/zricethezav/gitleaks/releases/tag/v4.2.0. Reopen this issue if needed.

PR #368, opened 2020-04-14
fixing issue #367

steeve85:
### Description:

This PR fix an issue regarding the `--redact` option ( #332 #367 ). 
Because the SHA1 fingerprint was calculated after the line is redacted, some similar (but different) lines had the same fingerprint after the "secret" was redacted.

Example:

'django.contrib.auth.password_validation.MinimumLengthValidator'
'django.contrib.auth.password_validation.CommonPasswordValidator'

were both redacted to: 'django.contrib.auth.REDACTED' and as this string is used to calculate a fingerprint to have "unique" findings, the results were incorrect.

### Checklist:


zricethezav:
Thanks for the PR! Great catch

PR #455, opened 2020-10-28
Fixing a bug in commit-until and commit-since scanning options

steeve85:
### Description:

PR https://github.com/zricethezav/gitleaks/pull/385 added some interesting scanning features based on the time of the commit (`--commit-since` & `commit-until`).

I identified an issue where a scan with one of these options doesn't find secrets while the default scan did. The reason for that gap was because the `All` option from [LogOptions](https://github.com/go-git/go-git/blob/master/options.go#L359-L362) was missing.

This PR adds the `All` log option for `--commit-since` and `commit-until` scanning features to cover all the refs in refs/.

### Checklist:


zricethezav:
released in v7: https://github.com/zricethezav/gitleaks/blob/master/scan/utils.go#L271-L281

PR #456, opened 2020-10-30
Add support for reportGroup

### LINKED ISSUE #451: Add reportGroup option to config
Users might want to fine tune the "offender" field in the report to a specific capture group in their regex.

A config that supports `reportGroup`:
```
[[rules]]
	description = "AWS Secret Key"
	regex = '''(?i)aws(.{0,20})?(?-i)(['\"][0-9a-zA-Z\/+]{40}['\"])'''
	tags = ["key", "AWS"]
    reportGroup =  2
```
If the leak contained the line: `awsSecretKey = "thiswillmatchexampleawskey11111111111111"` only `"thiswillmatchexampleawskey11111111111111"` would be reported in the `Offender` field of the report

n/a

**Additional context**
n/a

cc @zricethezav

squaresurf:
Why:

* There are times where it would be nice to have a group from the regex
  reported rather than the complete match.

This change addresses the need by:

* Add the reportGroup configuration option for a rule.
* Update scan.CheckRules to return the reportGroup if set.

### Description:

### Checklist:

closes #451

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -2,6 +2,8 @@ package config
 
 import (
 	"fmt"
+	"io/ioutil"
+	"os"
 	"regexp"
 	"testing"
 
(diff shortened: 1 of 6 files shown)


config/config_test.go
```diff
@@ -129,3 +129,23 @@ func TestParse(t *testing.T) {
...
+	if err != nil {
+		t.Error(err)
+	}
+
+	expectedNumOfRules := 2
+	if len(config.Rules) != expectedNumOfRules {
+		t.Fatalf("expected %v rules", expectedNumOfRules)
+	}
+
+	if config.Rules[0].ReportGroup != 0 {
```
theoretick, line 144: **not blocking:** super minor but IMO it would be a slightly more robust to use a description reference  rather than index; i.e.

```go
idx := sort.Search(len(config.Rules), func(i int) bool {
        return config.Rules[i].Description ==  "Some Groups without a reportGroup"
    })

if config.Rules[idx].ReportGroup != 0 {
```

From a more general-musing standpoint, that's one issue I have with dedicated fixtures outside the specific test cases. This wouldn't matter as long as no other code uses `fields_test.toml`, but if another relies on the fixture and tweaks something, this could inadverently break with the relative index reference.

config/config_test.go
```diff
@@ -129,3 +129,23 @@ func TestParse(t *testing.T) {
 		}
 	}
 }
+
+func TestParseFields(t *testing.T) {
```
theoretick, line 133: **question:** Should this be more specific to ReportGroups?

scan/rule.go
```diff
@@ -121,6 +121,10 @@ func (repo *Repo) CheckRules(bundle *Bundle) {
 						continue
 					}
 
+					if 0 < rule.ReportGroup && rule.ReportGroup < len(groups) {
```
theoretick, line 124: **question:** Would it be worth adding a comment clarifying `0` is a submatch for the entire expression rather than a zero-count of capture groups? I know this is fairly universal so maybe I'm just not in a super regexy mindset but it was a thought since it took me a little bit to parse this one. Feel free to disregard

theoretick (commented):
Looks good, thanks @squaresurf!

config/config_test.go
squaresurf: I was thinking that this would eventually cover testing that all fields are parsed correctly, but I also wanted to scope my changes to a minimum. I added a comment to hopefully make that clearer. What do you think?

scan/rule.go
squaresurf: I added a comment. What do you think?

squaresurf (commented):
@theoretick I've replied to two of your comments and refactored the config test a bit. I updated the test to have the config inline next to the test and updated the assertions to look up rules by description rather than index. Let me know what you think.

@zricethezav this does change some of the paradigms with how this project tests things. I'm looking forward to your thoughts as well :smile:

config/config_test.go
```diff
@@ -129,3 +131,83 @@ func TestParse(t *testing.T) {
...
+[[rules]]
+	description = "Some Groups without a reportGroup"
+	regex = '(.)(.)'
+
+[[rules]]
+	description = "Some Groups"
+	regex = '(.)(.)'
+  reportGroup = 1
+`
+	configPath, err := writeTestConfig(tomlConfig)
```
zricethezav, line 148: **suggestion**: @squaresurf What about loading the config from a file in `test_data/test_configs`?

scan/rule.go
zricethezav: This is clear to me 👍

Issue #429, opened 2020-08-26
Allow custom config on top of the default config


stone-z:
**Is your feature request related to a problem? Please describe.**
Normally, I want to use a set of default detection rules (either the gitleaks default or my own), but some projects have files which I know I want to ignore.

I'd like to be able to use the default configuration as a baseline and supply only the allowlist items I want _in addition to_ the default allowlist. This way I can use the same config for many repos but tweak individual behaviors as needed.

Such a merge behavior could also be used to supply additional rules on top of the default (but I don't currently need this).

**Describe the solution you'd like**
A parameter allowing me to give a partial config file which will be applied on top of the default (or a specified config or repo-config file as a baseline).

Something like: 
Use the default config plus my overrides
`gitleaks --verbose --pretty --redact --extra-config=/my/config.toml` 

or

Use a common config plus my overrides
`gitleaks --verbose --pretty --redact --config=/my/base/config.toml --extra-config=/my/override/config.toml`

or

Use the repo's baseline plus my overrides
`gitleaks --verbose --pretty --redact --repo-config --extra-config=/my/override/config.toml`

**Describe alternatives you've considered**
I currently use a config file which consists of the entire contents of the default config, plus one additional line with one allowlist path. This is functionally fine, it just means I have to manually keep this config file up to date with changes in the default.

I would also be open to supplying exceptions as separate arguments but that seems like more implementation work.

cc @zricethezav

Twanislas:
Hi !

Thanks for the great work on Gitleaks !

I'm also interested in this, and will use @stone-z 's workaround in the meantime.

Cheers !

PR #458, opened 2020-11-02
Add ability to merge repo configurations, and user supplied configurations with specified or default configuration

eddie-northcutt-wfp0:
### Description:
Currently, users can either define a configuration, or define a repo config, but not both as one would be overwritten by the other. Now users can merge configurations together. For example, a repo can be ran against a master configuration, but say there is a one-off false positive in that repo. Instead of changing the master config, users can now include a repo config and merge it with the master config, allow listing the file in question. Additionally, users can define an additional config outside or inside of the context of a repo. This allow for merging up to three configurations at once. This works also works in the context of a bare repo if gitleaks is being ran as a pre-receive/update hook given the gitleaks config file can be accessed from the hook script.

#### New Options:
`--merge-repo-config`: merges `gitleaks.toml` or `.gitleaks.toml` stored in repo with the config passed to the manager, or the default config if one is not passed in.
`--additional-config`: path to an additional config to be merged with the manager's config, or the default config.

### Checklist:

Majority of changed files are new test_data. See 9e82914cbfa8c64a759b3ad87b3223a0e461c1a3 and 572545ea53deaa4f6315bfe0a8abad0b4df71a5b for the implementation changes.

resolves: #429


zricethezav:
@eddie-northcutt-wfp0 Thank you for the PR! I just got back from vacation and plan to do a thorough review this week or weekend hopefully.

eddie-northcutt-wfp0:
Closing as https://github.com/zricethezav/gitleaks/pull/487 now includes this functionality, and it is compatible with the `7.x.x` release

Issue #399, opened 2020-06-16
Enable traversing multi-level project groups in GitLab


crunchtime-ali:
**Is your feature request related to a problem? Please describe.**

GitLab supports multi-level organizations (project groups) and therefore I would like to be able to traverse the complete organization (or even GitLab instance) with gitleaks

**Describe the solution you'd like**
There should be a `traverse` option which traverses the organization to the most specific level and checks all projects for secrets when set to `true` (default would be `false`)

**Describe alternatives you've considered**
Right now I am collecting a list of all organizations (by calling the GitLab v4 API via cURL) before calling `gitleaks` for each individual one of them.

cc @zricethezav

matsskoglund:
I want this too. 
It can easily be done by adding in gitlab.go from line 63 to something like:

var recursive = true
	glOpts := &gitlab.ListGroupProjectsOptions{
	ListOptions:      listOpts,
	IncludeSubgroups: &recursive,
}

Need to get the traverse flag into the function and fix the argument flags though to make it an option. 

I am pretty new at Go but if I can get some assistance and suggestions how to proceed I could probably manage to create a pull request for this feature if wanted.

cc @zricethezav

zricethezav:
I am not supporting scans that use gitlab/github apis. See https://github.com/zricethezav/gitleaks/releases/tag/v7.0.0

PR #459, opened 2020-11-04
Added feature to search gitlab groups recursively

matsskoglund:
### Description:
Added a flag to be able to search gitlab subgroups recursively and not only the top group specified in the --org flag.

### Checklist:

* [x ] Does your PR pass tests?
The original ```manager``` test failed in the master branch with ```"manager/manager.go:161: conversion from int to string yields a string of one rune, not a string of digits (did you mean fmt.Sprint(x)?)"```

After my change the same test fails with the same error. Otherwise all tests works.

No

Yes
Automatic formatting changed the structure in the ```options.go```but it is only one line added, i.e. the addition of the recursive flag.

This should solve the Issue #399

### DIFF
--- a/hosts/gitlab.go
+++ b/hosts/gitlab.go
@@ -61,7 +61,8 @@ func (g *Gitlab) Scan() {
 
 		} else if g.manager.Opts.Organization != "" {
 			glOpts := &gitlab.ListGroupProjectsOptions{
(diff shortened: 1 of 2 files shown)


zricethezav:
I am no longer supporting github/gitlab integrations

PR #487, opened 2020-12-10
v7.x Update - Add ability to append repo configurations, and user supplied configurations to the specified or default configuration

eddie-northcutt-wfp0:
### Description:
Note: this is an update of https://github.com/zricethezav/gitleaks/pull/458 with changes that are compatible with `v7.x` version of Gitleaks.

Currently, users can either define a configuration, or define a repo config, but not both as one would be overwritten by the other. Now users can merge configurations together. For example, a repo can be ran against a master configuration, but say there is a one-off false positive in that repo. Instead of changing the master config, users can now include a repo config and merge it with the master config, allow listing the file in question. Additionally, users can define an additional config outside or inside of the context of a repo. This allows for merging up to three configurations at once. 

#### New Options:
`--append-repo-config`: appends config stored in repo with the config located at `--config-path `, or the default config if one is not specified.
`--additional-config`: path to an additional config to be appended to the config located at `--config-path `, and/or the repo config , and/or the default config.

### Checklist:

resolves: #429


zricethezav:
@eddie-northcutt-wfp0 thanks for opening a new MR for v7.x.x. Hopefully the code was a little easier to understand. 

I agree with the ability to merge in a repo-config, that seems useful. But I'm failing to see the need for `--additional-config`. Can you give me a specific use case where that would be needed? I'm hesitant to add 2 more config options as that would bring the total to 4. 

I can see how `--config-path=/path/to/main/config` `--repo-config-path=/repo/specific/config` `--merge-repo-config` makes sense but not really seeing where `--additional-config` would be useful.

I'm also wondering if maybe `--merge-repo-config` should be renamed `--append-repo-config` as it doesn't actually resolve merge conflicts like name rule description collisions

scan/scan.go
```diff
@@ -45,6 +45,13 @@ func NewScanner(opts options.Options, cfg config.Config) (Scanner, error) {
 		return nil, err
 	}
 	if st == typeDirScanner {
+		if opts.AdditionalConfig != "" && err == nil {
```
zricethezav, line 48: Err check is not necessary
```suggestion
		if opts.AdditionalConfig != ""  {
```

scan/scan.go
```diff
@@ -56,12 +63,25 @@ func NewScanner(opts options.Options, cfg config.Config) (Scanner, error) {
 		}
 	}
 
-	// load up alternative config if possible, if not use manager's config
+	// load up alternative config if possible, if not use default/specified config. Will merge if MergeRepoConfig is true
```
zricethezav, line 66: 👍

scan/scan.go
```diff
@@ -56,12 +63,25 @@ func NewScanner(opts options.Options, cfg config.Config) (Scanner, error) {
...
 	if opts.RepoConfigPath != "" && !opts.NoGit {
-		cfg, err = config.LoadRepoConfig(repo, opts.RepoConfigPath)
+		repoCfg, err := config.LoadRepoConfig(repo, opts.RepoConfigPath)
+		if opts.MergeRepoConfig && err == nil {
```
zricethezav, line 69: I would prefer 
```golang
	if opts.RepoConfigPath != "" && !opts.NoGit {
		repoCfg, err := config.LoadRepoConfig(repo, opts.RepoConfigPath)
		if err != nil {
			return nil, err
		}
		if opts.MergeRepoConfig {
			cfg = cfg.Merge(repoCfg)
		} else {
			cfg = repoCfg
		}
	}
```

eddie-northcutt-wfp0:
@zricethezav Thanks for taking some time to review this! The `--additional-config` option was added to enable the option of applying an extra configuration on top of the default config, or the config supplied to `--config-path`. This also allows for combining config files outside of the context of a repo since it is read from the file system. This idea came from #429. If you think the cli complexity is too much of a trade off for this feature, I am not opposed to removing it as I understand where you are coming from. Let me know what you think.

As for the `--append-repo-config` suggestion, you make a good point. I will get that verbiage corrected, in addition to your other suggestions.

scan/scan.go
eddie-northcutt-wfp0: Removed

scan/scan.go
eddie-northcutt-wfp0: Totally agree, much more readable this way. Not sure why I wrote it the way I did  🤷‍♂️.

zricethezav (approved):

zricethezav:
@eddie-northcutt-wfp0 sorry for taking so long on this. Approved and merged, this will be shipped in the next release, v7.3.0

PR #491, opened 2020-12-11
Quiet flag

zricethezav:
### Description:
Adds a `-q/--quiet` flag which logs leaks one line at a time instead of the default "pretty" json. This also sets the log level to errors only. Good for processes that was to injest the output of gitleaks without opening a file.

### Checklist:


aaronkollasch:
I updated gitleaks and found that the `-q/--quiet` flag is no longer present. Is there a reason for that, or an alternative?

PR #508, opened 2021-01-19
Always generate a report if --report is passed

### LINKED ISSUE #505: Always generate SARIF output
When no findings are found, Gitleaks doesn't generate a report. I think this makes sense generally, but when SARIF output is enabled, it makes sense to generate an empty report. Advantages:
- This simplifies any processing, since we can always assume that the output is always valid JSON.
- The SARIF output contains information about the tool, which is useful for whatever consumes the report.

Always generate a SARIF document regardless of whether any issues were found.

Keeping the current behaviour.

cc @zricethezav

kiwiz:
### Description:
Resolves #505 

### Checklist:

* [x] Does your PR pass tests? (I'm getting a `testing.go:1038: race detected during execution of test` that happens without any changes?)
* [ ] Have you written new tests for your changes? (`options_test.go` is empty?)

### DIFF
--- a/scan/report.go
+++ b/scan/report.go
@@ -27,7 +27,6 @@ func WriteReport(report Report, opts options.Options, cfg config.Config) error {
 		logrus.Warn("leaks found: ", len(report.Leaks))
 	} else {
 		logrus.Info("No leaks found")
-		return nil
 	}
 
 	if opts.Report == "" {


zricethezav:
@kiwiz I'm actually thinking the default behavior should be to always generate a report file if `--report` is set. Could you remove the `always-report` option and remove the return statement?

kiwiz:
That sounds even better! Fixed.

PR #510, opened 2021-01-22
Better defaults to avoid duplicated info

seb-elttam:
### Description:
This is just to tidy up the code a little bit. The path to the file is already included in the `File` attribute and so doesn't need to be duplicated in the `Offender` attribute.

It might look like I'm being pedantic (and you would be right) but I'm implementing new features in my own fork and won't be able to submit PRs unless we share the same cleaner foundations.

Happy to discuss though :)

### Checklist:


w0rmr1d3r:
The user has been deleted from Github, I'm closing this PR.

PR #511, opened 2021-01-22
Report the entropy of every regex group

seb-elttam:
### Description:
This is my take on adding entropy to the report, because I wasn't happy with #506.

When a rule's regex matches a line, we compute the entropy of every capturing group (including group 0, the full `offender` match) and add them to the `Entropy` attribute.

For example consider this rule:
```
[[rules]]
  description = "Env Var"
  regex = '''(?i)(db[a-z]+|secret|key|api|password|pass|pw|auth)=([0-9a-zA-Z-_\/+!{}/=]{8,120})'''
```
And the line:
```
POSTGRES_PASSWORD=tookahlaiphee2KieTeeg5ooxutang4o   # Change this
```
The report now includes entropy information:
```
  "offender": "PASSWORD=tookahlaiphee2KieTeeg5ooxutang4o",
  "entropy": "4.43, 2.75, 3.88",
```
Which is useful to then add an entropy range with the right values:
```
[[rules]]
  description = "Env Var"
  regex = '''(?i)(db[a-z]+|secret|key|api|password|pass|pw|auth)=([0-9a-zA-Z-_\/+!{}/=]{8,120})'''
  [[rules.Entropies]]
    Min = "3.5"
    Max = "4.5"
    Group = "2"
```

Keen on your thoughts.

### Checklist:


w0rmr1d3r:
@zricethezav since this PR is from a GitHub deleted user, can we close it?
I also believe it duplicates this other PR https://github.com/zricethezav/gitleaks/pull/506

PR #495, opened 2020-12-21
Enhance ssh auth options

alcastronic:
#### Description:
Currently gitleaks only has support for ssh keys without a pasphrase. This PR adds the ability to specify the passphrase    
for an ssh key via the command line. The password will be retrieved by `syscall.Stdin` hence the import `syscall` was added.    
Furthermore functionallity to support ssh agent authentication is added via the `NewSSHAgentAuth` function from go-git.       

Fixes #263 How to pass SSH key password as argument?

#### ssh key with passphrase
It is possible to use an ssh key with passphrase either by specifing an ssh key with the  `--ssh-key= ` option or    
use the default key file location.  The ssh passphrase will be retrieved in an interactive prompt when the flag `--ssh-pass`    
is specified on the commandline.     

Specify an ssh key and ask for the passphrase.
```bash
# ./build/gitleaks-linux-amd64 --ssh-key=/home/tronic/.ssh/id_rsa --ssh-pass -r git@github.com:zricethezav/gitleaks.git    
Enter Password: 
INFO[0004] cloning... git@github.com:zricethezav/gitleaks.git     
INFO[0024] scan time: 17 seconds 770 milliseconds 891 microseconds     
INFO[0024] commits scanned: 612                             
WARN[0024] leaks found: 745    
```

#### ssh key default location and passphrase
When the `--ssh-key` option is ommitted the default key location under `/home/$user/.ssh/id_rsa` will be used.
```bash
./build/gitleaks-linux-amd64 --ssh-pass -r git@github.com:zricethezav/gitleaks.git    
Enter Password: 
INFO[0005] cloning... git@github.com:zricethezav/gitleaks.git     
INFO[0032] scan time: 17 seconds 883 milliseconds 37 microseconds     
INFO[0032] commits scanned: 612                           
WARN[0032] leaks found: 745    
```
#### ssh agent
It is also possible to authenticate with an ssh agent.

1. Add the private key to the local ssh agent
2. Execute gitleaks with the `--ssh-agent` option

When the ssh agent option is used gitleaks opens a channel to the local ssh agent.    
```bash
 ./build/gitleaks-linux-amd64 --ssh-agent -r git@github.com:zricethezav/gitleaks.git    
INFO[0000] cloning... git@github.com:zricethezav/gitleaks.git     
INFO[0021] scan time: 17 seconds 783 milliseconds 151 microseconds     
INFO[0021] commits scanned: 612                           
WARN[0021] leaks found: 745     
```

### Checklist:

[...]

PR #536, opened 2021-03-10
SSH auth: Add custom username support

Tomy2e:
### Description:
This PR adds support for custom usernames when cloning repositories using SSH authentication.
Previously, the user `git` was hard-coded.

This PR also adds compatibility for using `ssh://` repo URLs without using ssh-agent.
A PR such as #495 will have to be merged to restore ssh-agent compatibility.
I will remove this part if such a breaking change is not desirable.

### Checklist:

### DIFF
--- a/options/options.go
+++ b/options/options.go
@@ -184,6 +184,10 @@ func SSHAuth(opts Options) (*ssh.PublicKeys, error) {
 		return nil, fmt.Errorf("user must be specified in the URL")
 	}
 
+	// the part of the RepoURL before the "@" (params[0]) can be something like:
+	// - "ssh://user" if RepoURL is an ssh:// URL
+	// - "user" if RepoURL uses scp-like syntax
+	// we must strip the protocol if it is present so that we only have "user"
 	username := strings.Replace(params[0], "ssh://", "", 1)
 
 	if opts.SSH != "" {
--- a/options/options.go
+++ b/options/options.go
@@ -144,8 +144,8 @@ func (opts Options) CloneOptions() (*git.CloneOptions, error) {
 
 	var auth transport.AuthMethod
 
-	if strings.HasPrefix(opts.RepoURL, "git") {
-		// using git protocol so needs ssh auth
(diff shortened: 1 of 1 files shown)


adrianbn:
This and https://github.com/zricethezav/gitleaks/pull/495 are very useful features. Can we help here to get this merged?

options/options.go
```diff
@@ -144,8 +144,8 @@ func (opts Options) CloneOptions() (*git.CloneOptions, error) {
 
 	var auth transport.AuthMethod
 
-	if strings.HasPrefix(opts.RepoURL, "git") {
-		// using git protocol so needs ssh auth
+	if strings.HasPrefix(opts.RepoURL, "ssh://") || (!strings.Contains(opts.RepoURL, "://") && strings.Contains(opts.RepoURL, ":")) {
```
zricethezav, line 147: @Tomy2e can we augment this to support using sshauth when `opts.RepoURL` is prefixed with `git`. The reason why `git` is hardcoded is that the default URL github/gitlab provide is `git@github/gitlab...`.

zricethezav:
> This and #495 are very useful features. Can we help here to get this merged?

@adrianbn working on the review now. Sorry this through the cracks

options/options.go
Tomy2e: Hi @zricethezav! This code already supports URLs provided by GitHub/GitLab. They provide an "scp-like" URL, which is detected by making sure the URL doesn't contain the substring `://` and it contains the `:` character. The `git` prefix you are talking about is the user.

I think it's a simple solution, maybe not the best. The go-git library [uses a regex pattern](https://github.com/go-git/go-git/blob/bf3471db54b0255ab5b159005069f37528a151b7/internal/url/url.go#L20) to detect these URLs.

Basically, it will work whether the user is `git` or something else. SSH URLs are documented [here](https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols#_the_ssh_protocol) if you need more information. (In the current implementation, it won't work if no user is specified, SSHAuth() will return an error.)

The following command works for me, let me know if  you have issues:
```console
$ go run main.go --repo-url=git@github.com:zricethezav/gitleaks.git
```

zricethezav: @Tomy2e perfect, thank you for explaining that 😄

PR #598, opened 2021-07-25
Add multi line support

rzandonai:
Create a parameter in rule s to set what rule is a multiline,
added all rules calls Inspect and Inspect File if is a multiline rule.
Created a Multiline rule for dotnet default values,
and a test to use

### Description:
Add multiline rules support.

### Checklist:


scan/nogit.go
```diff
@@ -77,6 +78,28 @@ func (ngs *NoGitScanner) Scan() (Report, error) {
...
+					offenders := rule.InspectFile(string(bytesRead))
+					if offenders != nil {
+						for _, offender := range offenders {
+							leak := NewLeak(offender.Match, offender.ToString(), offender.Line).WithEntropy(offender.EntropyLevel)
+							relPath, err := filepath.Rel(ngs.opts.Path, p)
+							if err != nil {
+								leak.File = p
+							} else {
+								leak.File = relPath
+							}
```
w0rmr1d3r, line 92: This if-else conditional can be simplified so only the if is needed

w0rmr1d3r (commented):
Besides the comment I left, the comments put here `config/rule.go` don't seem quite clear to what they preceed.

The rest of it looks awesome!

scan/nogit.go
```diff
@@ -73,6 +74,27 @@ func (ngs *NoGitScanner) Scan() (Report, error) {
 			}
 
 			for _, rule := range ngs.cfg.Rules {
+				if rule.Multiline {
+					bytesRead, _ := ioutil.ReadFile(p)
```
ed-wp, line 78: Shouldn't we check the error here? Files which are too large to fit in 1 operation will silently fail.

config/rule.go
```diff
@@ -82,6 +85,48 @@ func (r *Rule) Inspect(line string) *Offender {
 	}
 }
 
+// InspectFile checks the content of the file for  multilines leaks
+func (r *Rule) InspectFile(fileLines string) []Offender {
+	fileLines = regexp.MustCompile(`[\t\r\n]+`).ReplaceAllString(strings.TrimSpace(fileLines), "\n")
```
ed-wp, line 90: A recent multi-line regex should handle `\t\r\n` fine. This manipulates every line of prior to checking the regex which makes it very expensive on larger files.

ed-wp: Not sure what you mean by 👎 , but this is unnecessary, inefficient, and leads to errors and false negatives/positives. Automatically replacing all tabs with newlines will cause absolute chaos in tab-delimited files. We should not manipulate the content and instead recommend building good regex. Instead of filtering/manipulating `\t\r\n`, you can recommend people use `\s` (≡ `[\t\n\f\r]`) per [re2 docs](https://github.com/google/re2/wiki/Syntax). 

Example:
```
func something
\t\t\telse
```
If someone wants to write a regex:
`something\n\t\t\telse` (does not match, false negative)
`something\n\t+else` (does not match, false negative)
`something\n\s+else` (does match, true positive)
`something\n\n\n\nelse` (does match, false positive)

However, this would also match:
```
something



else
```
Which is a false positive.

zricethezav:
v8 supports multiline now

PR #612, opened 2021-08-29
fix(git-symlink-commits): fix handling symbolic links in a git repository

Ido-DY:
### Description:
gitleaks failed to scan commits contains symlinks.

How to reproduce:

```shell
mkdir /tmp/foo
cd $_
git init -q
mkdir dir
ln -s dir link
docker run --rm -v $PWD:/src --workdir /src zricethezav/gitleaks --verbose
```
Note that I confirmed that it's NOT a docker issue and the binary behave the same way.

Offered solution:
add check determines if file path is symlink, if so instead of reading it use the link path as content.

### Checklist:

### DIFF
--- a/scan/unstaged.go
+++ b/scan/unstaged.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"os"
 	"os/exec"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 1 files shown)


foolioo:
cc @zricethezav

zricethezav:
@foolioo very nice, thank you

PR #633, opened 2021-09-09
Simplify tests and continue on object not found errors

zricethezav:
### Description:
This PR handles two things:
1. Simplifying tests. Instead of having all tests inside one giant `scan_test.go` file, I have split these tests into their respective scanner types.
2. Fixed a bug that stops scans if object is not found. This can happen when commits get squashed but still linger around in ci/cd processes (like a commits_list.txt file)

### Checklist:


scan/commits.go
```diff
@@ -30,10 +32,11 @@ func NewCommitsScanner(opts options.Options, cfg config.Config, repo *git.Reposi
 // Scan kicks off a CommitsScanner Scan
 func (css *CommitsScanner) Scan() (Report, error) {
 	var scannerReport Report
-	for _, c := range css.commits {
-		c, err := obtainCommit(css.repo, c)
+	for _, commitHash := range css.commits {
+		c, err := obtainCommit(css.repo, commitHash)
 		if err != nil {
-			return scannerReport, nil
+			log.Errorf("skipping %s, err: %v", commitHash, err)
```
zricethezav, line 38: log error and continue when object not found

PR #652, opened 2021-10-01
Minimum go version should be enforced when installing gitleaks with go get command

weineran:
### Description:
Per the go.mod file, the minimum go version is 1.16. However, this needs to be enforced with build tags. (Why is the go.mod file not sufficient to enforce the minimum version? I do not know. https://github.com/golang/go/issues/48571#issuecomment-925597302)

#### Steps to reproduce
* Use a go version below 1.16.
    * I found [gvm](https://github.com/moovweb/gvm) to be useful for wrangling multiple go versions.
* Install gitleaks using this command from README.md: `GO111MODULE=on go get github.com/zricethezav/gitleaks/v7`
* **Expected**:  The command should fail with some useful message about how your go version needs to be >= 1.16
* **Actual**: The command fails with this message:
`.../pkg/mod/github.com/zricethezav/gitleaks/v7@v7.6.1/config/config.go:4:2: package embed is not in GOROOT (/Users/andrew/.gvm/gos/go1.15.15/src/embed)`

#### The Fix
* Use a build constraint (AKA a build tag) in config.go to enforce the minimum version on which that file can be built.
* Now, if you try to run `GO111MODULE=on go get github.com/zricethezav/gitleaks/v7`, you get a more helpful failure:
```
# github.com/zricethezav/gitleaks/v7/config
config/rule.go:39:16: undefined: Entropy
note: module requires Go 1.16
```

#### Additional Notes:
* Starting in go1.17, the syntax for build constraints changes. See [here](https://pkg.go.dev/cmd/go#:~:text=go%20versions%201.16%20and%20earlier%20used%20a%20different%20syntax%20for%20build%20constraints%2C%20with%20a%20%22%2F%2F%20%2Bbuild%22%20prefix.%20the%20gofmt%20command%20will%20add%20an%20equivalent%20%2F%2Fgo%3Abuild%20constraint%20when%20encountering%20the%20older%20syntax). It is allowed to use both syntaxes at the same time. In fact, "The gofmt command will add an equivalent //go:build constraint when encountering the older syntax." I will submit a separate PR for the go1.17 build constraint syntax.
* Running `make test` triggered some additional automatic changes that I don't necessarily understand:
    * Changes to go.mod and go.sum were triggered by the `go get golang.org/x/lint/golint` step in the Makefile
    * Changes to testdata/repos/basic/dotGit/index were triggered by the `go test ./... --race github.com/zricethezav/gitleaks -v` step in the Makefile.

### Checklist:

* [ ] Have you written new tests for your changes? NO


zricethezav:
@weineran thanks for the PR! This is a good change with two suggested changes:

- Remove `golang.org/x/lint` from the go.mod/go.sum file. Lint is not used in the gitleaks binary. Linting via `go vet and go lint` are only used when running a full test suite w/ `make test`. 

- Remove ` testdata/repos/basic/dotGit/index` from git. No reason this should be added unless there is something broken with how test repos are converted into git repos.

weineran:
@zricethezav Sorry I just noticed that you left me a suggestion several months ago!

Question: Should I just remove all changes from `go.sum` and `testdata/repos/basic/dotGit/index`? All of the changes to both of those files were triggered automatically by running `make test`. See my comments about that [here](https://github.com/zricethezav/gitleaks/pull/652#:~:text=Running%20make%20test%20triggered%20some%20additional%20automatic%20changes%20that%20I%20don%27t%20necessarily%20understand).

pacorreia:
@zricethezav This one needs your attention, there was a question to you
@weineran Can you sort out the conflicts meanwhile?

PR #660, opened 2021-10-09
Fixed bug in extractLine function which was returning incorrect line # …

GrahlmanMatthew:
…# when the line contains multiple vulnerability types

### Description:
I noticed an issue within the outputted JSON report in which the `lineNumber` field would display a value of 1 instead of the real value. This seems to occur only when a line of code is marked vulnerable by multiple rules in the `gitleaks.toml` file and the correct line number is only being reported for the 1rst rule (vulnerability) type.

Sample output report which outlines the issue - notice how the commit ID's & lines match but the line #'s don't
image

The problem is in the `extractLine` function. The function uses a Dictionary to track which lines have vulnerabilities & the condition statement currently it only returns the real line # for the 1rst vulnerability type identified on that line. Subsequent vulnerability types would loop until eventually returning the defaultLine value of 1. To resolve this, I added an `else if` to check whether the value of the key in the Dictionary is set to True and if so, return it's line number.

### Checklist:
* [SOME - Tests on latest release don't seem to pass with score of 100% as is] Does your PR pass tests?
* [NO] Have you written new tests for your changes?
* [YES] Have you lint your code locally prior to submission?


GrahlmanMatthew:
Hi @w0rmr1d3r,

I was having issues getting the tests to pass before on my Windows 10 desktop, even when running it against the master branch of `zricethezav/gitleaks` . It seems to be a cross-platform (OS) related issue as I'm able to run the tests successfully from my Macbook Pro.

To resolve this, I made the following changes;

- Changed `config_test.go` file to add file path support for Windows
- Re-added `testdata` directory as Windows was having issues locating some of the `.got` files. For this, I simply deleted the local contents of `testdata` then copied `zricethezav/gitleaks` master branch's `testdata` contents into the local folder. I'm unsure why this resolved the issue but it did.

After I made the changes described above, I verified that all tests passed on BOTH Windows & MacOS.

Screenshot of `go test ./... -cover` on Windows (against my fork of gitleaks)
image

Screenshot of `go test ./... -cover` on MacOS (against my fork of gitleaks)
image

## Updated Checklist:
[YES] Does your PR pass tests?
[YES] Have you written new tests for your changes?
[YES] Have you lint your code locally prior to submission?

PR #667, opened 2021-10-19
Org python script

lucas-apd:
### Description:
Python Script to scan all org repos

### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,7 +149,7 @@ gitleaks --repo-url=https://github.com/my-insecure/repo -v --report=my-report.js
 gitleaks --repo-url=https://github.com/my-insecure/repo --commit=commit-sha -v
 ```
 #### Scan All Organization repos:
- - [With python script](https://github.com/zricethezav/gitleaks/master/scripts/python/README.md)
+ - [Using a python script](https://github.com/zricethezav/gitleaks/master/scripts/python/README.md)
 
 
 #### Scan local repo:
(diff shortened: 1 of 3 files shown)


w0rmr1d3r:
Hey @lucas-apd !
I found that a single for loop might not be very efficient for organizations with more than 100 repos (for example).
So I created this repo -> https://github.com/w0rmr1d3r/cloner
in order to clone efficiently all the repos in an organization. I'm still building it, but I believe you can already test it if you want to do so. That way we would only need to use my repo first and then pass your script to only run gitleaks.

What do you think? Do you like the idea?

Cheers!

lucas-apd:
> Hey @lucas-apd ! I found that a single for loop might not be very efficient for organizations with more than 100 repos (for example). So I created this repo -> https://github.com/w0rmr1d3r/cloner in order to clone efficiently all the repos in an organization. I'm still building it, but I believe you can already test it if you want to do so. That way we would only need to use my repo first and then pass your script to only run gitleaks.
> 
> What do you think? Do you like the idea?
> 
> Cheers!
> 
> PS (and a hint): In version v8.0.0 of gitleaks, the cloning is deprecated...

Nice, @w0rmr1d3r . lets do this!

zricethezav:
Hi @lucas-apd, would you mind resolving merge conflicts, testing on the latest release of gitleaks, and renaming the file to `scripts/org-scan.py`?

zricethezav:
@foolioo good point, this would actually be better suited for a `scripts` or `tools` repo in  our org: https://github.com/gitleaks.

Gonna close this PR and follow-up in an issue

Issue #724, opened 2021-11-30
gitleaks hangs on protect command if there are no files to scan


galmalka:
**Describe the bug**
When running protect command, if there are no files to scan, the process hangs. It happens if there are no unstaged files without the --staged flag and vice versa.

**To Reproduce**
Run 'gitleaks protect' without any unstaged changes OR 'gitleaks protect --staged' without any staged changes.

**Expected behavior**
The process should return without any errors

**Screenshots**
N.A

**Basic Info (please complete the following information):**
 - OS: macOS Monterey
 - Gitleaks Version: 8.0.4

**Additional context**
Seems to work on 8.0.3 (and previous versions).

cc @zricethezav

zricethezav:
```
➜  ~/code/gitleaks (master) git status
On branch master
Your branch is up to date with 'origin/master'.

nothing to commit, working tree clean
➜  ~/code/gitleaks (master) ./gitleaks protect

    ○
    │╲
    │ ○
    ○ ░
    ░    gitleaks

10:48AM INF no leaks found
10:48AM INF scan duration: 13.515985ms
➜  ~/code/gitleaks (master)
```

Seems to be working for me, could post your commands/output like I did?

zricethezav:
This could be related to https://github.com/zricethezav/gitleaks/issues/722 although I doubt it, could you please try `v8.0.6` and see if it is fixed?

galmalka:
Thanks for the quick reply.
First of all, here is the output (with debug logging):
```
➜  ~/code git:(test) git status
On branch test
nothing to commit, working tree clean
➜  ~/code git:(test) gitleaks protect --log-level debug

    ○
    │╲
    │ ○
    ○ ░
    ░    gitleaks 

12:18AM DBG Using existing gitleaks config .gitleaks.toml
12:18AM DBG executing: /usr/local/bin/git -C . diff -U0 .
```
This is where the process seems to hang.

I checked with version 8.0.6 and it works, but then I also downloaded version 8.0.4 from the releases and when I run the bin from the download directory it works as well (for any version).
So now it seems that the issue only occurs for the version installed by homebrew on mac.

zricethezav:
@galmalka thank you for the context. Let's keep this open until its verified on v8.0.6 installed via brew.

zricethezav:
@galmalka now that https://formulae.brew.sh/formula/gitleaks is on v8.0.6, could you see if this is still an issue for you? If it's not then we can close this issue

[...]

PR #729, opened 2021-12-01
Update pre-commit step to run gitleaks checks

veleek:
### Description:
The current `entry` value for the pre-commit config causes it to just run gitleaks with no command.  This runs the same command that as defined in the `scripts/pre-commit.py` except that it doesn't check whether `git config --bool hooks.gitleaks` is true, but it's probably safe to assume that if it's configured as a pre-commit check then it should be okay to run it.

### Checklist:

* [X] Does your PR pass tests? Not a go code change.
* [X] Have you written new tests for your changes? N/A
* [X] Have you lint your code locally prior to submission? Not a go code change.


veleek:
Should this command run with --redacted?

zricethezav:
@veleek awesome, thanks for the PR.

> Should this command run with --redacted?

That makes sense to me, good to hide secrets from the terminal's scroll history

veleek:
@zricethezav - Hold off on this merging this change.  I'm running into a problem where this is hanging, and I think it's the same error as #724.  Adding more detail to that issue.

zricethezav:
@veleek ping me when this PR is ready and I'll get it merged in

veleek:
So this works if you're running pre-commit locally (e.g. as an actually git pre-commit hook) which runs it's checks against any staged files.  This matches the `--staged` flag that's passed to the tool.  Unfortunately, when running pre-commit as a Github action, the `--all-files` flag is passed to pre-commit so that the script runs against everything because it's running against a merged branch so there aren't any files staged.  

https://github.com/pre-commit/action/blob/master/action.yml

So works, locally as a GH Action, not really sure what the expected way to fix this would be.  I _think_ the correct thing to do would be to change gitleaks to read the filenames to process from the ones passed by pre-commit.  So pre-commit wouldn't need to use the `--staged` flag at all.  Bigger change for gitleaks though, so I say accept this PR as is and add an issue to make it work when using `--all-files` on pre-commit later.

zricethezav:
@veleek I'm glad to hear it works locally as that is the most important time to catch secrets before they are committed. 

> Unfortunately, when running pre-commit as a Github action, the --all-files flag is passed to pre-commit so that the script runs against everything because it's running against a merged branch so there aren't any files staged.

Gitleaks should be running `detect` during github actions as changes have already been committed. Or am I misunderstanding something?

chriselion:
> Gitleaks should be running `detect` during github actions as changes have already been committed. Or am I misunderstanding something?

I think that makes sense, but currently it's not possible to do that via the pre-commit hook.

Maybe a solution would be to add another hook that runs `detect`? I think that would look something like

```
- id: gitleaks-detect
  name: TODO
  description: TODO
  entry: gitleaks detect --no-git --redact
  args: ["--verbose"]
  language: golang
```

That will result in multiple files being passed after the entry and args, so the final command line would look like `gitleaks detect --no-git --redact --verbose foo.txt bar.go baz.py`. Am I correct in my reading of the docs that this will run on just the contents of those 3 files (not their git history)? 

Side note: it's up to you, but I feel like `--verbose` should be optional; by putting it in `args`, the user can still override it.

PR #735, opened 2021-12-03
fix: format dates in log in a portable way

fgreinacher:
The output of time.String() depends on the runtime environment and should only be used for debugging.

This commit ensures that a well-defined UTC time is written to the report.

### Description:

### Checklist:


zricethezav:
@fgreinacher awesome! Great catch. Approving and merging, will go out with next release

PR #760, opened 2021-12-29
ignore k8s apiVersion in generic-api-key pattern

amandre:
### Description:
Current implementation of the generic-api-key pattern finds plenty of false positives matching `apiVersion` (very popular Kubernetes keyword):
```
"Description": "Generic API Key",
"Match": "apiVersion: '2010-03-31'",
```
this change is to ignore the `apiVersion` string during secret lookup.

### Checklist:


zricethezav:
@amandre smart, thanks for the PR. I agree with this change and verified it on regex101

PR #802, opened 2022-03-07
Escape - character in regex character groups

idawson-gl:
### Description:
Escapes the `-` character in a number of regular expression character groups.
Removes a duplicate regex.

### Checklist:


idawson-gl:
The `-v` verbosity flag in the make test(s) was in the wrong location, i've also updated that so you can see which test fails.

Unfortunately it failed once, but did not reproduce after i enabled verbosity 🤔

zricethezav:
thanks @idawson-gl!

zricethezav:
> The `-v` verbosity flag in the make test(s) was in the wrong location, i've also updated that so you can see which test fails.
> 
> Unfortunately it failed once, but did not reproduce after i enabled verbosity 🤔

@idawson-gl what was the failure btw? There is a flaky test that fails every so often because, _i think_, the git index file gets into a weird state when converting `dotGit` to `.git` over and over in the tests. Still gotta figure that out.

idawson-gl:
```
{"level":"error","time":"2022-03-07T03:08:33Z","message":"'git <command> [<revision>...] -- [<file>...]'"}
FAIL	github.com/zricethezav/gitleaks/v8/detect	0.191s
```

PR #804, opened 2022-03-12
Refactor `detect`, add `entropy` to all findings

zricethezav:
### Description:
Refactoring the `detect` module. This change will make it easier to add features and optimizations in the future. Also it reads better.

### Added `Detector` struct which is the entry point and engine for scans:
```golang
type Detector struct {
	// Config is the configuration for the detector
	Config config.Config

	// Redact is a flag to redact findings. This is exported
	// so users using gitleaks as a library can set this flag
	// without calling `detector.Start(cmd *cobra.Command)`
	Redact bool

	// verbose is a flag to print findings
	Verbose bool

	// commitMap is used to keep track of commits that have been scanned.
	// This is only used for logging purposes and git scans.
	commitMap map[string]bool

	// findingMutex is to prevent concurrent access to the
	// findings slice when adding findings.
	findingMutex *sync.Mutex

	// findings is a slice of report.Findings. This is the result
	// of the detector's scan which can then be used to generate a
	// report.
	findings []report.Finding
}
```
### Added `Fragment` which is the subject of the detector, aka what gets scanned:

```golang
// Fragment contains the data to be scanned
type Fragment struct {
	// Raw is the raw content of the fragment
	Raw string

	// FilePath is the path to the file if applicable
	FilePath string

	// CommitSHA is the SHA of the commit if applicable
	CommitSHA string

	// newlineIndices is a list of indices of newlines in the raw content.
	// This is used to calculate the line location of a finding
	newlineIndices [][]int
}
```

### `DetectString`, `DetectBytes`, `NewDetectorDefaultConfig`, `NewDetector(cfg)`
```golang
package main

import (
   "fmt"
   
   "github.com/zricethezav/gitleaks/v8/detect"
)

func main() {
	detector, _ := detect.NewDetectorDefaultConfig()
	for _, f := range detector.DetectString("my secret") {
		fmt.Printf("Found secret: %v", f)
	}
}
```

### Entropy values added to all findings
Entropy is now present for all findings, regardless if `entropy` is set for the rule or not.

### Checklist:


zricethezav:
Note that that flaky git test has always been flaky

PR #837, opened 2022-04-08
config: better secret detection in bash scripts

pijoter:
### Description:
in bash, these assignments are equivalent (below), so the apostrophe and quotation marks should be optional
also, very often `passwd` variable name is used

`PASSWORD="secret"`
`PASSWORD='secret'`
`PASSWORD=secret`

### Checklist:


zricethezav:
The gitleaks config should handle this now

PR #839, opened 2022-04-12
Allowlist checks before extracting from a secret group

0xn3va:
### Description:

Current implementation checks if the secret is in the allowlist when the secret has been extracted from the secret group. As a result, it does not allow ignoring FP cases based on variable names in generic-based rules. For example, I can't exclude findings like `TokenSpec tspec = PKCS8EncodedTokenSpec` using the following rule:

```toml
regex = '''(?i)(token)(.{0,16})([0-9a-zA-Z]{8,64})'''
entropy = 3.7
secretGroup = 3

[rules.allowlist]
    regexes = [
        '''(?i)tokenspec'''
   ]
```

The PR swaps checking and extraction that allows such rules. So, if you only need to check the secret from the secret group you can write regex in the allowlist section.

### Checklist:

### DIFF
--- a/detect/detect.go
+++ b/detect/detect.go
@@ -194,7 +194,13 @@ func (d *Detector) detectRule(fragment Fragment, rule *config.Rule) []report.Fin
 			gitleaksAllowSignature) {
 			continue
 		}
-
+		
+		// check if the secret is in the allowlist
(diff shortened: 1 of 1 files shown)


zricethezav:
Good call, originally this was switch when I tried to introduce stopwords in the allowlists directly. However, I need to rework that idea.

Thanks for this PR.

PR #852, opened 2022-04-22
Adding quiet mode to silence banner

durkinza:
### Description:
Adding an argument to hide the gitleaks banner. 
This is useful for incorporating gitleaks into a pipeline of commands where the standard error file needs minimal output. 

### Checklist:


zricethezav:
@durkinza could you change the argument to `--no-banner`. `-q/--quiet` to me would signal suppressing findings or some other functionality. Imo it's better to be explicit here.

cmd/root.go
```diff
@@ -77,7 +78,10 @@ func initLog() {
 }
 
 func initConfig() {
-	fmt.Fprint(os.Stderr, banner)
+	hide_banner, err := rootCmd.Flags().GetBool("no-banner")
```
zricethezav, line 81: @durkinza 
`hideBanner` to be consistent please

Issue #865, opened 2022-05-05
Algolia key support


SimonGurney:
An Algolia key matches the regex `[a-zA-Z0-9]{32}`.

Algolia keys are quite common now and are often included in front end JS code to build autocomplete forms etc.  Overly permissive keys are an issue as it means attackers can manipulate data etc and gitleaks would make an awesome tool to find them.

The regex for the key is likely prone to false positives, but no more so than "generic-api-key".

I am happy to do the PR on this, but I was wondering if you would consider a regex without an obvious prefix/suffix which would eliminate false positives.
cc @zricethezav

zricethezav:
Hi @SimonGurney, all semi-generic regexes follow a similar pattern to that of https://github.com/zricethezav/gitleaks/blob/master/config/gitleaks.toml#L74, for example. I generate these rules with a utility here https://github.com/zricethezav/gitleaks/blob/master/cmd/generate/config/main.go. 

I can add this sometime in the next two weeks if you haven't gotten to it already

SimonGurney:
I'll get a PR in tomorrow.  So does the keyword need to be found in the document for the regex to find the secret?

zricethezav:
@SimonGurney yes, the keyword is essentially a ROI marker or prefilter. If a git patch fragment does not contain a the keyword, that rule can be ignored.

SimonGurney:
Ok understood.  I think in *most* cases the word "Algolia" would be on the same line, but it's definitely not required and I imagine a diff which changes the key could be as simple as 

Key=[key]

Would you want the keyword including or not?  I guess it's a balance between eliminating false positives and missing secrets right?

zricethezav:
> but it's definitely not required and I imagine a diff which changes the key could be as simple as Key=[key]

I don't follow

> Would you want the keyword including or not?  I guess it's a balance between eliminating false positives and missing secrets right?

Keywords should always be included to reduce fps. The generic key would likely catch keys that don't contain `Algolia`

SimonGurney:
Makes sense, I'll raise the PR tomorrow.

Thanks for your help 😃

SimonGurney:
Is this what you were thinking, I just based it off old PRs, or do you want me to use the rule generator and add one to here cmd/generate/config/rules/?

SimonGurney:
@zricethezav  are you happy with the PR?

PR #866, opened 2022-05-06
feat: add algolia key support

SimonGurney:
### Description:
To add Algolia key support as per issue  #865 

### Checklist:

  No, its an additional rule
  No, its an additional rule
  Not aware of a TOML linter but happy to do so

### DIFF
--- a/cmd/generate/config/rules/algolia.go
+++ b/cmd/generate/config/rules/algolia.go
@@ -1,0 +1,23 @@
+package rules
+
+import (
+	"github.com/zricethezav/gitleaks/v8/cmd/generate/secrets"
+	"github.com/zricethezav/gitleaks/v8/config"
+)
+
+func AlgoliaApiKey() *config.Rule {
+	// define rule
+	r := config.Rule{
+		Description: "Algolia API Key",
+		RuleID:      "algolia-api-key",
+		Regex:       generateUniqueTokenRegex(`[a-z0-9]{32}`),
+		Keywords:    []string{"algolia"},
+	}
+
+	// validate
+	tps := []string{
+		"algolia_key := " + secrets.NewSecret(hex("32")),
+	}
+	return validate(r, tps, nil)
+}
+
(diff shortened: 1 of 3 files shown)


cmd/generate/config/rules/algolia.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/zricethezav/gitleaks/v8/cmd/generate/secrets"
+	"github.com/zricethezav/gitleaks/v8/config"
+)
+
+func AlgoliaApiKey() *config.Rule {
+	// define rule
+	r := config.Rule{
+		Description: "Algolia API Key",
+		RuleID:      "algolia-api-key",
+		Regex:       generateUniqueTokenRegex(`[a-z0-9]{32}`),
```
zricethezav, line 13: @SimonGurney this should use [generateSemiGenericRegex](https://github.com/zricethezav/gitleaks/blob/master/cmd/generate/config/rules/rule.go#L31-L42) since the algolia token does not have any unique prefix/suffix that is part of the rule. This will slightly change the generated regular expression.

SimonGurney: hmmm, I thought the opposite was true...  It has no unique suffix/prefix so would use the unique token regex rather than generateSemiGenericRegex which wants unique identifiers?

I must have misread it.

I can't compile the go to test, I'll have to leave it you

zricethezav: > It has no unique suffix/prefix so would use the unique token regex

Using the `generateUniqueTokenRegex` means the regular expression can be less strict as there are uniquely identifiable patterns within the token. Using `generateSemiGenericRegex` means that the token follows a generic api token format like a 32 alphanumeric character sequence in this case. 

> I must have misread it.

No worries, this functionality isn't really documented anywhere and is a work in progress. The config generation `cmd` was really just a forcing function for me to finally get the default gitleaks rules in order.

> I can't compile the go to test, I'll have to leave it you

I can merge this as is then update the code so you get contribution credit if you'd like

SimonGurney: Sounds like a plan 👍

PR #881, opened 2022-05-30
Limit newlines regex

zricethezav:
### Description:
Removes newlines are part of the "identifier-suffix" regex group. This will prevent false positives like:

```
password combination.

R5: Regulatory--21
```

from being reported. 
### Checklist:


baruchiro:
But now we're missing all the key-value k8s like yamls:

```yaml
- name: db_passowrd
  value: "any-password-here"
```

What about allowing only one new line?

Issue #911, opened 2022-07-05
AllowList entire line checking


AnthonyPoschen:
**Is your feature request related to a problem? Please describe.**
when trying to whitelist / allow certain matches i would like to define regex's to be allowed that check the entire line and not just the match / secret to allow removal of false positives and would be the feature request from the outcome of #882 

**Describe the solution you'd like**
a way to have detector.config.allowlist.Regexes to scan the entire line for matches and not just the secret.

**Describe alternatives you've considered**
whitelisting file paths but that can hide true positives

**Additional context**
a use case i hit when scanning code within the company where i saw thousands of false positives when people were using vlocity https://github.com/vlocityinc/vlocity_build 

the matches looked like the following within a json file
```js
        "%vlocity_namespace%__GlobalKey__c": "58df4b82-a72c-2851-aacd-0a28e1b3e481",
```

i initially pressumed the regex section would be doing full lines and had to dig into the source code to realise it only compares the match, as i wanted to add the following rule ( as i am using this library directly i was writing it in go )
```go
	AllowVelocityRegexp = regexp.MustCompile("^\\s*\\\"\\%vlocity_namespace\\%__GlobalKey__c\\\"\\:\\s*\\\".*\"\\s*,?\\s*$")
```

cc @zricethezav

PR #912, opened 2022-07-05
Allow list global scan

AnthonyPoschen:
### Description:
the changes are a proposal to address #911 they have been tested with real workloads and do resolve the example provided in the issue #911 I am not sure if these variable names are a good fit, Happy to take suggestions and modify the PR accordingly.

### Checklist:

* [ Y ] Does your PR pass tests?
* [ N ] Have you written new tests for your changes?
* [ Y ] Have you lint your code locally prior to submission?

### DIFF
--- a/detect/detect.go
+++ b/detect/detect.go
@@ -135,7 +135,6 @@ func NewDetectorDefaultConfig() (*Detector, error) {
 func (d *Detector) AddGitleaksIgnore(gitleaksIgnorePath string) error {
 	log.Debug().Msgf("found .gitleaksignore file: %s", gitleaksIgnorePath)
 	file, err := os.Open(gitleaksIgnorePath)
-
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


detect/detect.go
```diff
@@ -201,8 +201,12 @@ func (d *Detector) detectRule(fragment Fragment, rule *config.Rule) []report.Fin
 		}
 
 		// check if the secret is in the allowlist
-		if rule.Allowlist.RegexAllowed(finding.Secret) ||
-			d.Config.Allowlist.RegexAllowed(finding.Secret) {
+		regexMatchText := finding.Secret
+		if d.Config.Allowlist.RegexFullLine {
+			regexMatchText = fragment.Raw
```
zricethezav, line 206: @zanven42 `fragment.Raw` could be multiline, you would have to use `fragment.newLineIndices` and some of the code in the `location` helper function to determine the line content

zanven42: good point, I had a think regarding this and wonder if it is more ideal for future proofing to have this feature check against the entire match including multi line cases and update the documentation and wording of variables to align.

Because i imagine one day someone will complain they can't get multi line checks to work and then either this feature is deprecated or it gets more complex with 2 booleans to turn on.

zanven42: it could be my understanding of "fragment.Raw" is wrong and it returns things outside of the scope of the match.

zanven42:
@zricethezav i modified the code to use the output from `location` helper function to get the string segment of the line to avoid multi line matches.

Maybe my idea of considering multi line matching should be left to a valid use case scenario in another ticket to address.

zanven42:
had an idea where instead of a boolean we can have a int to determine level
0 == secret
1 == full line
2 == multi line

Thoughts @zricethezav

geek-zuizui:
I also encountered this kind of problem. Will the function of this problem be updated later? @zricethezav

AnthonyPoschen:
@zricethezav any guidance on next steps? i made the recommended modifications from the review

AnthonyPoschen:
made changes to clarify this is against the full match rather than single line

PR #930, opened 2022-07-25
refactor: more precise rule for private keys

very-doge-wow:
### Description:
The current regex didn't match PGP private keys anymore, since they start with `BEGIN PGP PRIVATE KEY BLOCK` and the `BLOCK` never matched for the existing regex. I've made that part optional so that all strings matching for the current regex will still match for the new regex.

### Checklist:

### DIFF
--- a/cmd/generate/config/rules/privatekey.go
+++ b/cmd/generate/config/rules/privatekey.go
@@ -11,7 +11,7 @@ func PrivateKey() *config.Rule {
 	r := config.Rule{
 		Description: "Private Key",
 		RuleID:      "private-key",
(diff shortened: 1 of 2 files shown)


zricethezav:
@very-doge-wow could you follow the [contributing guidelines](https://github.com/zricethezav/gitleaks/blob/master/CONTRIBUTING.md) and update this PR? Thanks!

very-doge-wow:
@zricethezav Hi, I've changed the PR so that the rule can be generated. Hope it's correct this time.

very-doge-wow:
@zricethezav Did you take look?

very-doge-wow:
@zricethezav Ping

PR #938, opened 2022-08-05
Feat/ignore finding

zricethezav:
### Description:
This PR adds a new feature that gives users the ability to ignore findings from previous scans. This will be a nice way to resolve false positives gitleaks-action as well as prevent noisy full history scans.

### Checklist:


zricethezav:
> @github-actions
gitleaks / gitleaks (pull_request) Failing after 7s — gitleaks

I believe the command needs to be updated to `--no-merges --first-parent` to get the proper range. Right now this PR is including merge commits

zricethezav:
@weineran have any thoughts on this one?

detect/utils.go
```diff
@@ -21,6 +22,11 @@ func augmentGitFinding(finding report.Finding, textFragment *gitdiff.TextFragmen
 		finding.EndLine += int(textFragment.NewPosition)
 	}
 
+	// generate finding hash
+	h := sha1.New()
+	h.Write([]byte(fmt.Sprintf("%s:%s:%s:%d", finding.Commit, finding.File, finding.RuleID, finding.StartLine)))
+	finding.Fingerprint = fmt.Sprintf("%x", h.Sum(nil))
```
weineran, line 28: @zricethezav Sorry for the late comment, but I think instead of a hash, it would be better to use a plaintext concatenation as the fingerprint. That way, when someone tries to read their .gitleaksignore file later, they will be able to clearly see what they've already added.

weineran: Also, it would be nice if the user could list only some fields in the .gitleaksignore line, and let gitleaks treat the others as a wildcard. For example, if I specify only File and RuleId, it would tell gitleaks to ignore that RuleId for _all_ commits for _all_ lines in that file.

Need to think about if this can be implemented cleanly with decent performance...

PR #955, opened 2022-08-17
feat: adding support for detecting checked in OpenStack RC files.

srelf-eschercloud:
### Description:
the admin version of which very often contains passwords, this does mean that non admin versions may also be stopped from being checked in, so would need to be ignored

### Checklist:

### DIFF
--- a/cmd/generate/config/rules/openstack.go
+++ b/cmd/generate/config/rules/openstack.go
@@ -1,0 +1,26 @@
+package rules
+
+import (
+	"github.com/zricethezav/gitleaks/v8/cmd/generate/secrets"
+	"github.com/zricethezav/gitleaks/v8/config"
+	"regexp"
+)
+
+func OpenStackRCPassword() *config.Rule {
+	// define rule
+	r := config.Rule{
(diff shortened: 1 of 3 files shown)


cmd/generate/config/rules/openstack.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/zricethezav/gitleaks/v8/config"
+	"regexp"
+)
+
+func OpenStackRCPassword() *config.Rule {
+	// define rule
+	r := config.Rule{
+		RuleID:      "openstack_rc_password",
+		Description: "OpenStack RC file password",
+		Regex:       regexp.MustCompile(`export OS_PASSWORD=(.*)`),
```
zricethezav, line 14: @srelf-eschercloud can you restrict the secret group? I don't accept (.*). https://blog.gitleaks.io/finding-secrets-with-regular-expressions-d90493bb3784 might give you some ideas

srelf-eschercloud: sure ill take a look.

PR #971, opened 2022-08-31
pipe - add input from Stdin

### LINKED ISSUE #836: separate gitleaks from git so any tool that feed unified diff log can be used as source
gitleaks is a great tool but is limited to git as a scm; would be nice if gitleaks (as security detection engine) could be separated from source log data (git or raw files if --no-git option is used)

something like: 
`git log -p --ancestry-path "$(git rev-parse HEAD~1)".."$(git rev-parse HEAD~0)" | gitleaks detect -v`
`hg log -g -p -r -2..tip | gitleaks detect -v`
`diff -ubwi file_a file_b | gitleaks detect -v`

as far as I know, there is no such universal tool around

cc @zricethezav

rommi4:
### Description:

with some copy-paste added pipe:
resolves #836 

usage example:
```
curl -sS -H "Accept: application/vnd.github.patch"  \
         -H "Authorization: Bearer ${TOKEN}" \
          https://api.github.com/repos/${OWNER}/${REPO}/compare/${BASE_REF}...${HEAD_REF}  | \
  ./gitleaks detect --pipe 
```
or
```
git log -p "${BASE_REF}...${HEAD_REF}" |  ./gitleaks detect --pipe
```

### Checklist:


zricethezav:
Thanks for the PR @rommi4! 

I want the `--pipe` option to be `git` agnostic. I.e, `--pipe` should be able to accept _any_ input, not just git diffs. This is useful for streaming applications and building services around gitleaks. That said, I think there is some value in your implementation of `FromPipe` in `git/git.go`. There could be some code that quickly checks the format of stdin which would either route the detection to `gitdiff.Parse(cmd, stdout)` or straight to `detect.DetectReader` (to be implemented).

Does that make sense? It's a bit early in the morning but I can provide some more context later with what I'm looking for for this feature

rommi4:
Hello, @zricethezav

Just updated my PR

There's simple check for [git input.](https://github.com/zricethezav/gitleaks/pull/971/files#diff-843a307bde5ddd61294c58685e121707f08adfacc34e49f7399cfc6e42e57947R368)
If git, proceed to `detect.DetectPipeGit`

`detect.DetectPipe` may be updated in future with parsers other than git
---
Already working from my fork for 2 weeks this way:
```
curl -sS  https://api.github.com/repos/${OWNER}/${REPO}/compare/${BASE_REF}...${HEAD_REF}  | ./gitleaks detect --pipe
```
suitable for checking pull requests in LARGE repos without checkout in Github Actions every time.

detect/detect.go
```diff
@@ -392,6 +393,105 @@ func (d *Detector) DetectGit(source string, logOpts string, gitScanType GitScanT
...
+		DetectReader string
+	)
+
+	input := io.ReadCloser(os.Stdin)
+	buf := new(bytes.Buffer)
+	buf.ReadFrom(input)
+	inputStr := buf.String()
+
+	if strings.Contains(inputStr, "diff --git") {
+		DetectReader = "git"
```
weineran, line 413: Does it make sense to move the d.DetectPipeGit call up here and remove the switch statement below?

Also, consider adding a Debug log (or Info?) saying that the input was inferred to be a git diff and will be scanned as such.

detect/detect.go
```diff
@@ -392,6 +393,105 @@ func (d *Detector) DetectGit(source string, logOpts string, gitScanType GitScanT
...
+	} else {
+		DetectReader = "unknown"
```
weineran, line 415: If it's not "git" does it make sense to just treat the input as a file and scan it? Not sure how complex that is, so if you think that should be deferred for future work, that's fine.

cmd/detect.go
```diff
@@ -114,6 +114,10 @@ func runDetect(cmd *cobra.Command, args []string) {
 	if err != nil {
 		log.Fatal().Err(err)
 	}
+	fromPipe, err := cmd.Flags().GetBool("pipe")
+	if err != nil {
+		log.Fatal().Err(err)
+	}
```
weineran, line 120: It looks like this duplicates lines 113-116?

weineran (changes requested):
General comment: Is it possible to unit test any of the new functionality?

Some specific comments in the code...

rommi4:
may be closed because of https://github.com/zricethezav/gitleaks/pull/1012

weineran:
@rommi4 Does #1012 meet your requirements as well?

rommi4:
Testing at the moment. Will update if find anything

Issue #974, opened 2022-09-05
Baselining


gawansch:
**Is your feature request related to a problem? Please describe.**
There is no easy way to create a baseline allowing users to ignore large number of legacy false positives. 

**Describe the solution you'd like**
By adding --baseline argument pointing to a previously generated report, the user can ask gitleaks to hide known issues from the new report.

**Describe alternatives you've considered**
Write a script comparing and old report to a new report used after running gitleaks.

**Additional context**
I have a solution ready/working on the feature.

cc @zricethezav

zricethezav:
Closing with https://github.com/zricethezav/gitleaks/pull/975, thanks for the contribution @gawansch, you will be mentioned in the release

PR #975, opened 2022-09-05
Add baseline

gawansch:
### Description:
Add ability to ignore previously found issues
#974 
### Checklist:

*  [x] Have you written new tests for your changes?

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,12 +200,12 @@ gitleaks will ignore any old findings that are present in the baseline. A baseli
 gitleaks detect --report-path gitleaks-report.json # This will save the report in a file called gitleaks-report.json
 ```
 
-***NOTE***: If you decide to commit the baseline file to the repo, you must add it to the .gitleaksignore file. Otherwise, the contents of the baseline file will generate new issues.
-
 Once as baseline is created it can be applied when running the detect command again:
+
 ```
 gitleaks detect --baseline-path gitleaks-report.json --report-path findings.json
 ```
+
 After running the detect command with the --baseline-path parameter, report output (findings.json) will only contain new issues.
 
 ### Verify Findings
--- a/README.md
+++ b/README.md
@@ -156,7 +156,7 @@ Flags:
       --redact                 redact secrets from logs and stdout
   -f, --report-format string   output format (json, csv, sarif)
   -r, --report-path string     report file
-  -b, --baseline-path          path to existing a previously generated report with known issues that gitleaks should ignore
+  -b, --baseline-path          path to a previously generated report with known issues that gitleaks should ignore
   -s, --source string          path to source (git repo, directory, file)
   -v, --verbose                show verbose output from scan
 
--- a/README.md
+++ b/README.md
@@ -156,6 +156,7 @@ Flags:
       --redact                 redact secrets from logs and stdout
   -f, --report-format string   output format (json, csv, sarif)
   -r, --report-path string     report file
+  -b, --baseline-path          path to existing a previously generated report with known issues that gitleaks should ignore
   -s, --source string          path to source (git repo, directory, file)
   -v, --verbose                show verbose output from scan
 
@@ -189,6 +190,23 @@ When running `protect` on a git repository, gitleaks will parse the output of a 
 as a pre-commit.
 
 **NOTE**: the `protect` command can only be used on git repos, running `protect` on files or directories will result in an error message.
+
+### Creating a baseline
+
+When scanning large repositories or repositories with a long history, it can be convenient to use a baseline. When using a baseline, 
+gitleaks will ignore any old findings that are present in the baseline. A baseline can be any gitleaks report. To create a gitleaks report, run gitleaks with the `--report-path` parameter. 
+
+```
+gitleaks detect --report-path gitleaks-report.json # This will save the report in a file called gitleaks-report.json
+```
+
+***NOTE***: If you decide to commit the baseline file to the repo, you must add it to the .gitleaksignore file. Otherwise, the contents of the baseline file will generate new issues.
+
+Once as baseline is created it can be applied when running the detect command again:
+```
+gitleaks detect --baseline-path gitleaks-report.json --report-path findings.json
+```
+After running the detect command with the --baseline-path parameter, report output (findings.json) will only contain new issues.
 
 ### Verify Findings
 
--- a/cmd/detect.go
+++ b/cmd/detect.go
@@ -75,6 +75,15 @@ func runDetect(cmd *cobra.Command, args []string) {
 		detector.AddGitleaksIgnore(filepath.Join(source, ".gitleaksignore"))
 	}
 
+	// ignore findings from the baseline (an existing report in json format generated earlier)
+	baselinePath, _ := cmd.Flags().GetString("baseline-path")
+	if baselinePath != "" {
+		err = detector.AddBaseline(baselinePath)
+		if err != nil {
+			log.Error().Msgf("Could not load baseline. The path must point of a gitleaks report generated using the default format: %s", err)
+		}
+	}
+
 	// set exit code
 	exitCode, err := cmd.Flags().GetInt("exit-code")
 	if err != nil {
--- a/detect/baseline.go
+++ b/detect/baseline.go
@@ -23,7 +23,7 @@ func IsNew(finding report.Finding, baseline []report.Finding) bool {
 			finding.EndLine == b.EndLine &&
 			finding.Entropy == b.Entropy &&
 			finding.File == b.File &&
-			// Omit checking finding.Fingerprint - if the format of the fingerprint changes later the users will see unexecpted behaviour
+			// Omit checking finding.Fingerprint - if the format of the fingerprint changes, the users will see unexpected behaviour
 			finding.Match == b.Match &&
 			finding.Message == b.Message &&
 			finding.RuleID == b.RuleID &&
--- a/detect/baseline.go
+++ b/detect/baseline.go
@@ -1,0 +1,58 @@
+package detect
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"os"
+
+	"github.com/zricethezav/gitleaks/v8/report"
+)
+
+func IsNew(finding report.Finding, baseline []report.Finding) bool {
+	// Explicitly testing each property as it gives significantly better performance in comparison to cmp.Equal(). Drawback is that
+	// the code requires maintanance if/when the Finding struct changes
+	for _, b := range baseline {
+
+		if finding.Author == b.Author &&
+			finding.Commit == b.Commit &&
+			finding.Date == b.Date &&
+			finding.Description == b.Description &&
+			finding.Email == b.Email &&
+			finding.EndColumn == b.EndColumn &&
+			finding.EndLine == b.EndLine &&
+			finding.Entropy == b.Entropy &&
+			finding.File == b.File &&
+			// Omit checking finding.Fingerprint - if the format of the fingerprint changes later the users will see unexecpted behaviour
+			finding.Match == b.Match &&
+			finding.Message == b.Message &&
+			finding.RuleID == b.RuleID &&
+			finding.Secret == b.Secret &&
+			finding.StartColumn == b.StartColumn &&
+			finding.StartLine == b.StartLine {
+			return false
+		}
+	}
+	return true
+}
+
+func LoadBaseline(baselinePath string) ([]report.Finding, error) {
+	var previousFindings []report.Finding
+	jsonFile, err := os.Open(baselinePath)
+	if err != nil {
+		return nil, fmt.Errorf("could not open %s", baselinePath)
+	}
+
+	bytes, err := ioutil.ReadAll(jsonFile)
+	jsonFile.Close()
+	if err != nil {
+		return nil, fmt.Errorf("could not read data from the file %s", baselinePath)
+	}
+
+	err = json.Unmarshal(bytes, &previousFindings)
+	if err != nil {
+		return nil, fmt.Errorf("the format of the file %s is not supported", baselinePath)
+	}
+
+	return previousFindings, nil
+}
--- a/detect/detect.go
+++ b/detect/detect.go
@@ -66,6 +66,9 @@ type Detector struct {
 	// matching given a set of words (keywords from the rules in the config)
 	prefilter ahocorasick.AhoCorasick
 
+	// a list of known findings that should be ignored
+	baseline []report.Finding
+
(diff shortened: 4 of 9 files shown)


zricethezav:
Hi @Gawan-0e278eed, thanks for opening this PR! I know that this is still marked as a draft but wanted to give some thoughts on it anyways. First, I think this is a good addition and introducing a `--baseline` option makes sense.  `--baseline` adds yet another knob to turn when reducing false positives. I see this feature in the same category as `allowlists` and `.gitleaksignore`. As such, I would expect an entry in the README documenting how to use `--baseline`. Secondly, I noticed that this unmarshals `json` reports only. Gitleaks supports `json`, `SARIF` and `csv` reports. I think it would make sense to support all Gitleaks report file types for `--baseline`. WDYT?

cc @weineran

Gawan-0e278eed:
Hi  @zricethezav,  Really happy to hear that I am on the right track. Adding support for SARIF will be pretty straightforward (it is still json). I am sure I can find a solution for CSV. I might have to use another flag, unless I want to have a taste on the file before loading it. I'll make an update later this week.

weineran:
Agreed that this is going to be a great addition.

I also agree we should try to support all report formats, but I humbly suggest that the first version would be OK if it only supports a subset of formats if that's easier. We can always add additional formats in subsequent PRs.

zricethezav:
@Gawan-0e278eed @weineran 

> I also agree we should try to support all report formats, but I humbly suggest that the first version would be OK if it only supports a subset of formats if that's easier. We can always add additional formats in subsequent PRs.

That's fair, Gawan, ping me when you're ready for a review

Gawan-0e278eed:
@zricethezav Let's give it a go and then add support for sarif and csv later

.vscode/launch.json
```diff
@@ -0,0 +1,17 @@
+{
```
zricethezav, line 1: @Gawan-0e278eed this file shouldn't be committed

Gawan-0e278eed: File removed

zricethezav:
@Gawan-0e278eed After thinking about this feature a bit I think `baseline` should be implemented in a similar way to `.gitleaksignore`. Rather than retroactively removing findings, we should prevent baseline finding matches from being added to the list of findings. You can see how we do this with `fingerprint` checking if a `.gitleaksignore` file is present here: https://github.com/zricethezav/gitleaks/blob/master/detect/detect.go#L463-L482. 

With the current proposal we are logging secrets that should be ignored if `--baseline-path` is set (try running gitleaks with `--baseline-path` and `-v` to see this behavior).

cmd/detect.go
```diff
@@ -110,6 +110,16 @@ func runDetect(cmd *cobra.Command, args []string) {
 		}
 	}
 
+	// Remove findings from the baseline (an existing report in json format generated earlier)
+	baselinePath, _ := cmd.Flags().GetString("baseline-path")
+	if baselinePath != "" {
+		findings, err = report.Baseline(findings, baselinePath)
+		if err != nil {
+			log.Info().Msgf("Ignoring --baseline-path flag, because the baseline could not be loaded: %s", err)
```
weineran, line 118: If someone passes a bad baseline path, I think gitleaks should exit immediately with this error. We have to assume they intended to pass a good path and they'd want a noisy failure if the path was bad.

weineran (commented):
small comment

cmd/detect.go
Gawan-0e278eed: Making it fatal with exit code 1.

Gawan-0e278eed:
> @Gawan-0e278eed After thinking about this feature a bit I think `baseline` should be implemented in a similar way to `.gitleaksignore`. Rather than retroactively removing findings, we should prevent baseline finding matches from being added to the list of findings. You can see how we do this with `fingerprint` checking if a `.gitleaksignore` file is present here: https://github.com/zricethezav/gitleaks/blob/master/detect/detect.go#L463-L482.
> 
> With the current proposal we are logging secrets that should be ignored if `--baseline-path` is set (try running gitleaks with `--baseline-path` and `-v` to see this behavior).

@zricethezav  Good point - Now we are approaching something where we could simply turn the baseline feature into an autogenerated .gitleaksignore file, though keeping the findings explicit will make it easier for a response team to review additions to the baseline. WDYT?

zricethezav:
@Gawan-0e278eed 
> On the other hand: keeping the baseline in the current format will make it easier for peers to review additions to the baseline. WDYT?

I think keeping `--baseline` and `.gitleaksignore` separate for now is a good idea. We could add a disclaimer that `baseline` is still an experimental feature and subject to change like we do for `.gitleaksignore`: https://github.com/zricethezav/gitleaks#gitleaksignore

README.md
```diff
@@ -189,6 +190,23 @@ as a pre-commit.
...
+### Creating a baseline
+
+When scanning large repositories or repositories with a long history, it can be convenient to use a baseline. When using a baseline, 
+gitleaks will ignore any old findings which is present in the baseline. To create a baseline, run gitleaks with the `--report-path` parameter. 
+
+```
+gitleaks detect --report-path .gitleaksbaseline # This will save the report in a file called .gitleaksbaseline
+```
+
+***NOTE***: If you decide to save the baseline within the source three, you must add it to the .gitleaksignore file. Otherwise, the contents of this file will generate new issues.
```
zricethezav, line 202:
```suggestion
***NOTE***: If you decide to commit the baseline file to the repo, you must add it to the .gitleaksignore file. Otherwise, the contents of the baseline file will generate new issues.
```

README.md
```diff
@@ -189,6 +190,23 @@ as a pre-commit.
 
 **NOTE**: the `protect` command can only be used on git repos, running `protect` on files or directories will result in an error message.
 
+### Creating a baseline
+
+When scanning large repositories or repositories with a long history, it can be convenient to use a baseline. When using a baseline, 
+gitleaks will ignore any old findings which is present in the baseline. To create a baseline, run gitleaks with the `--report-path` parameter. 
+
+```
+gitleaks detect --report-path .gitleaksbaseline # This will save the report in a file called .gitleaksbaseline
```
zricethezav, line 199: I think we should rename the report file to something more conventional like `gitleaks-report.json`. Naming it `.gitleaksbaseline` might give readers the misconception that `--report-path` is something other than a gitleaks report.

README.md
```diff
@@ -189,6 +190,23 @@ as a pre-commit.
 
 **NOTE**: the `protect` command can only be used on git repos, running `protect` on files or directories will result in an error message.
 
+### Creating a baseline
+
+When scanning large repositories or repositories with a long history, it can be convenient to use a baseline. When using a baseline, 
+gitleaks will ignore any old findings which is present in the baseline. To create a baseline, run gitleaks with the `--report-path` parameter.
```
zricethezav, line 196:
```suggestion
gitleaks will ignore any old findings that are present in the baseline. A baseline can be any gitleaks report. To create a gitleaks report, run gitleaks with the `--report-path` parameter. 
```

README.md
```diff
@@ -189,6 +190,23 @@ as a pre-commit.
...
+
+```
+gitleaks detect --report-path .gitleaksbaseline # This will save the report in a file called .gitleaksbaseline
+```
+
+***NOTE***: If you decide to save the baseline within the source three, you must add it to the .gitleaksignore file. Otherwise, the contents of this file will generate new issues.
+
+Once as baseline is created it can be applied when running the detect command again:
+```
+gitleaks detect --baseline-path .gitleaksbaseline --report-path findings.json
```
zricethezav, line 206: Changes should be reflected here too

cmd/detect.go
```diff
@@ -75,6 +75,16 @@ func runDetect(cmd *cobra.Command, args []string) {
 		detector.AddGitleaksIgnore(filepath.Join(source, ".gitleaksignore"))
 	}
 
+	// ignore findings from the baseline (an existing report in json format generated earlier)
+	baselinePath, _ := cmd.Flags().GetString("baseline-path")
+	if baselinePath != "" {
+		detector.AddBaseline(baselinePath)
```
zricethezav, line 81: @Gawan-0e278eed I think you're missing assigning the `err` here.

cmd/detect.go
```diff
@@ -75,6 +75,16 @@ func runDetect(cmd *cobra.Command, args []string) {
...
+		if err != nil {
+			log.Fatal().Msgf("Could not load baseline. The path must point of a gitleaks report generated using the default format: %s", err)
+		}
+		err = nil
```
zricethezav, line 85: I don't think this is necessary, `log.Fatal()` will exit the program

detect/baseline.go
```diff
@@ -0,0 +1,76 @@
...
+			finding.Entropy == b.Entropy &&
+			finding.File == b.File &&
+			finding.Fingerprint == b.Fingerprint &&
+			finding.Match == b.Match &&
+			finding.Message == b.Message &&
+			finding.RuleID == b.RuleID &&
+			finding.Secret == b.Secret &&
+			finding.StartColumn == b.StartColumn &&
+			finding.StartLine == b.StartLine &&
+			tagsEqual(finding.Tags, b.Tags) {
```
zricethezav, line 50: imo we don't need to check `tags` as these are optional values and could be updated by a change in the gitleaks config

detect/baseline.go
```diff
@@ -0,0 +1,76 @@
...
+			return false
+		}
+	}
+	return true
+}
+
+func LoadBaseline(baselinePath string) ([]report.Finding, error) {
+	var previousFindings []report.Finding
+	jsonFile, err := os.Open(baselinePath)
```
zricethezav, line 59:
```suggestion
	jsonFile, err := os.Open(baselinePath)
	defer jsonFile.close()
```

remove the `jsonFile.Close()` on line 65

detect/detect.go
```diff
@@ -473,6 +487,11 @@ func (d *Detector) addFinding(finding report.Finding) {
 		return
 	}
 
+	if d.baseline != nil && !IsNew(finding, d.baseline) {
+		log.Debug().Msgf("Ignoring finding with Finterprint %s", finding.Fingerprint)
```
zricethezav, line 491:
```suggestion
		log.Debug().Msgf("baseline duplicate -- ignoring finding with Fingerprint %s", finding.Fingerprint)
```

detect/baseline.go
Gawan-0e278eed: Good point, removing that check - also removing check on fingerprint. If for some reason the format of the finger print changes, users will get unexpected (old) findings

cmd/detect.go
Gawan-0e278eed: This was actually to meet andrews point:

cmd/detect.go
Gawan-0e278eed: Added in line 81. Regarding line 78, I believe it is fine to ignore any error here. If gitleaks doesn't get a baseline path it can just continue without any baselining.

cmd/detect.go
zricethezav: `log.Fatal()` will exit the program immediately after logging a message

Gawan-0e278eed: Using log.Error()  instead

gawansch:
@zricethezav I am in doubt about whether it makes sense to keep the core baseline functionality in separate files. If you prefer I can move it into the detect/detect.go file. Otherwise I believe the feature is complete. 

I can add support for sarif and csv in another PR once this is accepted.

README.md
```diff
@@ -156,6 +156,7 @@ Flags:
       --redact                 redact secrets from logs and stdout
   -f, --report-format string   output format (json, csv, sarif)
   -r, --report-path string     report file
+  -b, --baseline-path          path to existing a previously generated report with known issues that gitleaks should ignore
```
weineran, line 159: remove "existing"?

README.md
```diff
@@ -190,6 +191,23 @@ as a pre-commit.
...
+### Creating a baseline
+
+When scanning large repositories or repositories with a long history, it can be convenient to use a baseline. When using a baseline, 
+gitleaks will ignore any old findings that are present in the baseline. A baseline can be any gitleaks report. To create a gitleaks report, run gitleaks with the `--report-path` parameter. 
+
+```
+gitleaks detect --report-path gitleaks-report.json # This will save the report in a file called gitleaks-report.json
+```
+
+***NOTE***: If you decide to commit the baseline file to the repo, you must add it to the .gitleaksignore file. Otherwise, the contents of the baseline file will generate new issues.
```
zricethezav, line 203: > ***NOTE***: If you decide to commit the baseline file to the repo, you must add it to the .gitleaksignore file. Otherwise, the contents of the baseline file will generate new issues.

@gawansch thanks for bearing with us on this review. I did some testing locally and it works great! However, I think this can be avoided by _automatically_ adding an `allowlist` entry that specifies the path to the `baseline`. We do this with `config`. See https://github.com/zricethezav/gitleaks/blob/master/detect/detect.go#L426-L429.

We could add the baseline path to that condition like:
```
diff --git a/detect/detect.go b/detect/detect.go
index ff1dcda..8494cf4 100644
--- a/detect/detect.go
+++ b/detect/detect.go
@@ -69,6 +69,9 @@ type Detector struct {
        // a list of known findings that should be ignored
        baseline []report.Finding

+       // path to baseline
+       baselinePath string
+
        // gitleaksIgnore
        gitleaksIgnore map[string]bool
 }
@@ -156,6 +159,7 @@ func (d *Detector) AddBaseline(baselinePath string) error {
                }
                d.baseline = baseline
        }
+       d.baselinePath = baselinePath
        return nil
 }

@@ -438,7 +442,7 @@ func (d *Detector) Detect(fragment Fragment) []report.Finding {

        // check if filepath is allowed
        if fragment.FilePath != "" && (d.Config.Allowlist.PathAllowed(fragment.FilePath) ||
-               fragment.FilePath == d.Config.Path) {
+               fragment.FilePath == d.Config.Path || fragment.FilePath == d.baselinePath) {
                return findings
        }

```

gawansch: Good point! It will work as long as the baseline path is present, and improve the usability. 

When generating a new baseline users will do that without the `baseline-path` argument. The new baseline will contain these extra entries, which can lead to some confusion, but at this point we can count on users being able to put two and two together.

weineran: > When generating a new baseline users will do that without the baseline-path argument. The new baseline will contain these extra entries, which can lead to some confusion

This is a good point. Once this feature gets a little mileage, hopefully we'll develop a best practice to avoid this. It might be as simple as "be sure to delete or move your old baseline before creating a new one".

zricethezav:
Going to merge this and put it in the next release. Thanks @gawansch!

PR #985, opened 2022-09-23
Minor cleanup to error handling and logging

mojotx:
### Description:
These changes are minor, and mostly concerned with cleaning up some problems with the Zerolog logger, as well as checking error values more consistently. 

With Zerolog you always need to include `.Msg()` or `.Msgf()` at the end of the chain. Calling `.Err(err)` without a `.Msg()` call will result in an error not being logged, which makes it difficult to troubleshoot. (I learned this the hard way!)

Quoting from the Zerolog README:

> It is very important to note that when using the **zerolog** chaining API, as shown above (`log.Info().Msg("hello world"`), the chain must have either the `Msg` or `Msgf` method call. If you forget to add either of these, the log will not occur and there is no compile time error to alert you of this.

I did not write any new tests, because I did not add any new functionality. 

I did replace one call to `ioutil.ReadAll()`, which has been deprecated since Go 1.16, with `io.ReadAll()`. 

### Checklist:


zricethezav:
Hey @mojotx, thanks for the PR! I was unaware of that gotcha w.r.t to `zerolog`. Approved, merging 👍🏻

zricethezav (approved):

PR #1003, opened 2022-10-06
add --max-target-megabytes : maximum size for a file/blob to be scanned

roma8389:
### Description:
Adding maximum size for a file/blob to be scanned via a new switch case in the CLI --max-target-megabytes

### Checklist:

* [ ] Does your PR pass tests?  yes
* [ ] Have you written new tests for your changes? no
* [ ] Have you lint your code locally prior to submission? yes

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,7 +160,7 @@ Flags:
   -b, --baseline-path          path to a previously generated report with known issues that gitleaks should ignore
   -s, --source string          path to source (git repo, directory, file)
   -v, --verbose                show verbose output from scan
-  --max-target-megabytes       maximum size for a file/blob to be scanned
+  --max-target-megabytes       files larger than this will be skipped
 
 Use "gitleaks [command] --help" for more information about a command.
 ```
--- a/README.md
+++ b/README.md
@@ -160,6 +160,7 @@ Flags:
   -b, --baseline-path          path to a previously generated report with known issues that gitleaks should ignore
   -s, --source string          path to source (git repo, directory, file)
   -v, --verbose                show verbose output from scan
+  --max-target-megabytes       maximum size for a file/blob to be scanned
 
 Use "gitleaks [command] --help" for more information about a command.
 ```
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -160,6 +160,7 @@ Flags:
   -b, --baseline-path          path to a previously generated report with known issues that gitleaks should ignore
   -s, --source string          path to source (git repo, directory, file)
   -v, --verbose                show verbose output from scan
+  --max-target-megabytes       maximum size for a file/blob to be scanned
```
weineran, line 163: I suggest adding something like "Files larger than this will be skipped". Just to be totally explicit about the behavior.

weineran:
Thanks for the PR @roma8389 !

Linking the PR to this issue: https://github.com/zricethezav/gitleaks/issues/1001 (Let me know if I misunderstood, but I believe this PR is intended to resolve this issue)

roma8389:
@weineran any chance u can check the PR, i adjust the push according to your comment.
if u accept the PR can u create a build after it. ty!

cmd/root.go
```diff
@@ -45,6 +45,7 @@ func init() {
 	rootCmd.PersistentFlags().StringP("baseline-path", "b", "", "path to baseline with issues that can be ignored")
 	rootCmd.PersistentFlags().StringP("log-level", "l", "info", "log level (trace, debug, info, warn, error, fatal)")
 	rootCmd.PersistentFlags().BoolP("verbose", "v", false, "show verbose output from scan")
+	rootCmd.PersistentFlags().Int("max-target-megabytes", 0, "maximum size for a file/blob to be scanned")
```
zricethezav, line 48: This "help" message should be changed to match what is in the README per @weineran's suggestion

zricethezav (changes requested):
@roma8389 did some local testing of this feature and it seems to be working great! I think this is a really useful feature. Thanks for the PR. This will go into the next release. Just need to update one "help" message then this is good to go.

Per https://github.com/zricethezav/gitleaks/issues/1001

> i was wonder if it possible to get a CLI switch where i will add timeout pre rule

I think this is also a worthwhile feature to introduce FWIW

cmd/root.go
roma8389: @zricethezav Hi, ty for your comment. i pushed the change.
Will wait for the release its pretty impotent for me to have it so if its possible from u to have new version in couple of d will be great! ty for the tool\help\comments :)

PR #1010, opened 2022-10-11
Add support for following symlinks

RickyGrassmuck:
### Description:
Add support for optionally following symlinked files, resolves https://github.com/zricethezav/gitleaks/issues/934

### Checklist:


RickyGrassmuck:
The failures look like they are all caused by my addition of a private key in the testdata for use in the symlink detection tests. I'm assuming there needs to be exceptions added to prevent these tests from triggering but I'm not sure how or if that is even something I would be able to handle.

detect/detect.go
```diff
@@ -403,15 +415,33 @@ func (d *Detector) DetectFiles(source string) ([]report.Finding, error) {
...
+					}
+				}
+				if fInfo.Mode().Type() == fs.ModeSymlink && d.FollowSymlinks {
+					realPath, err := filepath.EvalSymlinks(path)
+					if err != nil {
+						return err
+					}
+					realPathFileInfo, _ := os.Stat(realPath)
+					if realPathFileInfo.IsDir() {
+						fmt.Printf("Found symlinked directory: %s -> %s [skipping]\n", path, realPath)
```
zricethezav, line 430: @RickyGrassmuck could we change this to a debug log instead? 
```suggestion
						log.Debug().Msgf("found symlinked directory: %s -> %s [skipping]", path, realPath)
```

RickyGrassmuck: Ahh, missed one, all fixed up!

cmd/detect.go
```diff
@@ -89,6 +91,11 @@ func runDetect(cmd *cobra.Command, args []string) {
 		}
 	}
 
+	// set follow symlinks flag
+	if detector.FollowSymlinks, err = cmd.Flags().GetBool("follow-symlinks"); err != nil {
+		log.Fatal().Err(err)
```
zricethezav, line 96:
```suggestion
		log.Fatal().Err(err).Msg("")
```
@RickyGrassmuck one last thing... this needs to call `.Msg("")`  otherwise the line will not be logged :/ see https://github.com/zricethezav/gitleaks/pull/985 for more details

RickyGrassmuck: Never used the Zerolog package before, gonna try to remember that one in the future. Appreciate you taking care of this!

PR #1011, opened 2022-10-12
fix bug in readme

sergiomarotco:
--max-target-megabytes INT

### Description:
 Added int type to --max-target-megabytes in Readme file


zricethezav:
Thanks @sergiomarotco

PR #1039, opened 2022-11-22
Fix allowlist regex

weineran:
### Description:
* Fixes https://github.com/zricethezav/gitleaks/issues/1025
* The allowlist regex is only applied to the finding's secret, not the full line.
* According to the [README](https://github.com/zricethezav/gitleaks#configuration), the stopwords target the secret whereas the regexes target the "entire regex match", which I interpreted to mean the `Line` in the `Finding` struct.
* So I changed the logic in detect.go to check the regexes against the Line.
* Also added some tests

### Checklist:


zricethezav:
Closing this as `regexTarget` takes care of allowlist matching on either secret, match, or line

PR #1069, opened 2023-01-04
fix: privatekey suffix between 3 and 5 chars

SimonGurney:
### Description:

Currently the Regex match for privatekeys only matches 4 "-" characters, but private keys typically end with 5 "-" characters.

I think the current behavior might actual be useful in detecting keys where there last characters has been accidentally ommited, but its a bit of an edge case.

I was using gitleaks to extract private keys from repositories and had issues handling them as they were being parsed out without the final closing "-".  This new signature will greedy match the "-"'s at the end of the key.

### Checklist:

### DIFF
--- a/cmd/generate/config/rules/privatekey.go
+++ b/cmd/generate/config/rules/privatekey.go
@@ -11,7 +11,7 @@ func PrivateKey() *config.Rule {
 	r := config.Rule{
 		Description: "Private Key",
 		RuleID:      "private-key",
-		Regex:       regexp.MustCompile(`(?i)-----BEGIN[ A-Z0-9_-]{0,100}PRIVATE KEY( BLOCK)?-----[\s\S-]*KEY----`),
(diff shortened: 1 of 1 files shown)


weineran:
@SimonGurney Thanks for the PR and thorough description.

I believe you'll have to run `cmd/generate/config/main.go` for your changes to be reflected in the default gitleaks.toml. See here: https://github.com/zricethezav/gitleaks/blob/master/CONTRIBUTING.md

icecream-monster:
hey I already have a PR for this!!: https://github.com/zricethezav/gitleaks/pull/1043

icecream-monster:
in file: config/gitleaks.toml 
line: 2414 
can someone merge this PR please? 😅

weineran:
@icecream-monster Indeed you do! What do you think about SimonGurney's idea for detecting between 3 and 5 dashes? Could be useful for detecting keys where there last characters has been accidentally ommited...?

PR #1071, opened 2023-01-11
add rule for checking nuget.config passwords

jankonikola93:
### Description:
Currently there is no support for finding private NuGet feed passwords.
Added regex will trigger a leak if Password/ClearTextPassword attributes appears with alphanumeric values.
For more info about NuGet configuration please refer to [MSDN NuGet Config](https://learn.microsoft.com/en-us/nuget/reference/nuget-config-file)

### Checklist:

### DIFF
--- a/cmd/generate/config/rules/nuget.go
+++ b/cmd/generate/config/rules/nuget.go
@@ -1,0 +1,24 @@
+package rules
+
+import (
+	"regexp"
+
+	"github.com/zricethezav/gitleaks/v8/config"
+)
+
+func Nuget() *config.Rule {
+	// define rule
+	r := config.Rule{
+		Description: "Nuget Password",
+		RuleID:      "nuget-config-password",
+		Regex:       regexp.MustCompile(`(?i)(<add key=\"(ClearTextPassword)|(Password)\"\s*)(value=\"[A-Za-z0-9]+\"\s*/>)`),
+		SecretGroup: 2,
+		Keywords:    []string{"cleartextpassword", "password"},
+	}
+
+	// validate
+	tps := []string{
+		`<add key="ClearTextPassword" value="nvnmsklavkoneroijvoks894789532ifjklwnfi28ur" />`,
+	}
+	return validate(r, tps, nil)
+}
--- a/config/gitleaks.toml
+++ b/config/gitleaks.toml
@@ -489,9 +489,6 @@ keywords = [
     "key","api","token","secret","client","passwd","password","auth","access",
 ]
 [rules.allowlist]
-paths = [
-  '''Database.refactorlog'''
-]
 stopwords= [
     "client",
     "endpoint",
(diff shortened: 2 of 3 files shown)


cmd/generate/config/rules/nuget.go
```diff
@@ -0,0 +1,24 @@
...
+	r := config.Rule{
+		Description: "Nuget Password",
+		RuleID:      "nuget-config-password",
+		Regex:       regexp.MustCompile(`(?i)(<add key=\"(ClearTextPassword)|(Password)\"\s*)(value=\"[A-Za-z0-9]+\"\s*/>)`),
+		SecretGroup: 2,
+		Keywords:    []string{"cleartextpassword", "password"},
+	}
+
+	// validate
+	tps := []string{
```
weineran, line 20: can you add an fps (i.e. negative example) to ensure this doesn't flag environment variables like this:
```
<packageSourceCredentials>
    <Contoso>
        <add key="Username" value="user@contoso.com" />
        <add key="ClearTextPassword" value="%ContosoPassword%" />
    </Contoso>
    <Test_x0020_Source>
        <add key="Username" value="user" />
        <add key="ClearTextPassword" value="%TestSourcePassword%" />
    </Test_x0020_Source>
</packageSourceCredentials>
```
https://learn.microsoft.com/en-us/nuget/reference/nuget-config-file

I know the example in the CONTRIBUTING.md doesn't show `fps`, but you can see what I'm talking about if you look at the validate function: https://github.com/zricethezav/gitleaks/blob/cd5226711335c68be1e720b318b7bc3135a30eb2/cmd/generate/config/rules/rule.go#L57

config/gitleaks.toml
```diff
@@ -489,9 +489,6 @@ keywords = [
     "key","api","token","secret","client","passwd","password","auth","access",
 ]
 [rules.allowlist]
-paths = [
-  '''Database.refactorlog'''
-]
```
weineran, line 494: @jankonikola93 Was this deletion intentional? Not sure why this was here in the first place to be honest, but it seems unrelated to your PR so we should probably leave it as-is unless there is some reason to remove it...?

jankonikola93: It was added when I run script for converting rules to gitleaks.toml file. I was not sure also why this was generated.

cmd/generate/config/rules/nuget.go
```diff
@@ -0,0 +1,27 @@
...
+
+	"github.com/zricethezav/gitleaks/v8/config"
+)
+
+func Nuget() *config.Rule {
+	// define rule
+	r := config.Rule{
+		Description: "Nuget Password",
+		RuleID:      "nuget-config-password",
+		Regex:       regexp.MustCompile(`(?i)(<add key=\"(ClearTextPassword)|(Password)\"\s*)(value=\"[A-Za-z0-9]+\"\s*/>)`),
```
fgreinacher, line 14: @jankonikola93 Why did you add `Password` here? While not really recommended and only supported on Windows I would say it's OK, because it's encrypted, see https://learn.microsoft.com/en-us/nuget/reference/nuget-config-file#packagesourcecredentials.

jankonikola93: I added it because in my opinion is better to avoid having all kinds of secrets stored in flies on your repo (even encrypted). For me it's better to replace them with placeholders then to store that values on your repo.

PR #1072, opened 2023-01-12
Fix rule for private keys

very-doge-wow:
### Description:
The rule for private keys wouldn't work with keys that start AND end with the BLOCK statement.

### Checklist:


zricethezav:
@very-doge-wow thanks for opening this PR. Looks good to me 👍🏻

PR #1076, opened 2023-01-13
ignore package-lock.json

lawndoc:
`package-lock.json` is automatically generated by npm and can cause a lot of false positives. There should never be secrets in the file because it just lists dependencies and is usually never edited by humans.


zricethezav:
@lawndoc thanks for the PR! Could you add this to the config template here https://github.com/zricethezav/gitleaks/blob/master/cmd/generate/config/rules/config.tmpl#L13-L21?

lawndoc:
@zricethezav done :)

lawndoc:
@zricethezav anything else missing or is this good to merge?

jasikpark (approved):
this seems reasonable 👍

Issue #539, opened 2021-03-16
Gitleaks misses quite a few Azure rules.


jessehouwing:
Gitleaks is missing quite a few rules for the Microsoft ecosystem, including Visual Studio, Azure and Azure DevOps.

Microsoft used to have a competing product called `credscan`, but it was recently deprecated in favor of the GitHub Security offerings.

I've ported most of the rules from credscan to the gitleaks format and put them in a repo here:
https://github.com/jessehouwing/gitleaks-azure

I'd love to contribute them, but I'm unsure to the process of vetting and approving these rules.

cc @zricethezav

---

Some features GitLeaks may be missing:

 * `credscan` has a few built-in decoders, so you can have it match a specific kind of pattern, say a base64 encoded string of 78 characters, then decode it and check the decoded value matches a regex. This reduces positives on base-64 encoded credentials that don't have an easily recognizable pattern.
 * similarly, it has a regex, xpath and jsonpath matcher.
 * it has a set of rules that use the .NET base library to assert things about the contents) that can detect whether a base64 encoded value is an image/icon/cert/jsonwebtoken. You can use these to include or exclude the values for further probing.

Without these the ruleset will trigger more false-positives.

jessehouwing:
I contributed in a project called SpamAssassin before. There I was abne tm setup what we called a Spam and a Ham corpus of examples to test rules against, many people separated oud good code vs bad code and every time the ruleset changed, they'd run the new ruleset and reported any new false positives and false negatives.

Would be nice if people could setup a private repo with a GitHub action to give you more confidence over the quality of the rules added through a pull request.

zricethezav:
Source for rules: https://www.powershellgallery.com/packages/AzSK.AzureDevOps/0.9.8/Content/Framework%5CConfigurations%5CSVT%5CAzureDevOps%5CCredentialPatterns.xml

jessehouwing:
Yups that's the pattern source. There isn't a 1-to-1 mapping, but it's close.

SebastianSchuetze:
Any update on a possible PR @jessehouwing  and @zricethezav . We use are going to use gitleaks internally at our company and implemented your rules @jessehouwing. We also started creating a repo that triggers the rules to get findings. We may be able to work something out to push this a little bit.

Could we integrate this into the default rule set?

zricethezav:
[...]

PR #1079, opened 2023-01-18
Initial set of Azure secrets for #539

dvasdekis:
### Description:
Please find in this PR an almost-complete set of Azure secrets as [sourced from here](https://www.powershellgallery.com/packages/AzSK.AzureDevOps/0.9.8/Content/Framework%5CConfigurations%5CSVT%5CAzureDevOps%5CCredentialPatterns.xml), plus some new ones from my own experience.

### Checklist:


dvasdekis:
I've commented out the secret types I couldn't get to work (mostly the XML ones) but the rest pass `go run main.go` now. Hopefully you can approve and pull in 👍

.gitignore
```diff
@@ -18,3 +18,12 @@ cmd/generate/config/gitleaks.toml
...
 dist/
+
+# Hugo build files (created automatically by VSCode Go extension)
+.hugo_build.lock
+
+# Hugo Sitemap.xml files (created automatically by VSCode Go extension)
+public/index.xml
+public/sitemap.xml
+public/categories/index.xml
+public/tags/index.xml
```
jessehouwing, line 29:
```suggestion
```

These should probably not be part of the PR.

dvasdekis: These files are created by `go run main.go`. 

Should these `.xml` files be included in the repository? If they should, then I'll commit them. If not, they should be in the .gitignore. Given that they're automatically built, `.gitignore` seems more appropriate. Yes, it isn't an atomic PR, but they should be there.

nikolamalesevic: @dvasdekis Maybe leave this out and get the PR approved? I'm really looking forward for your changes in the next release. :)

dvasdekis:
Sure, I've accepted the changes

dvasdekis:
Hi, can I get these changed merged please?

JoostVoskuil:
Hi @zricethezav , can this be merged? This is a big addition for the Microsoft techstack. Or is there an additional step required?

cmd/generate/config/rules/azure.go
```diff
@@ -0,0 +1,444 @@
...
+	return validate(r, tps, nil)
+}
+
+func AzureStorageCredentialXStore() *config.Rule {
+	// define rule
+	r := config.Rule{
+		Description: "CSCAN0100 - Found Azure storage credential in source code file.",
+		RuleID:      "azure-storage-credential-xstore",
+		SecretGroup: 1,
+		Regex: generateUniqueTokenRegex(`[<XstoreAccountInfo].*accountSharedKey\s*=\s*['"].*['"]`),
```
rgmz, line 132: Just curious, what's the reason for changing some of the patterns?

```diff
-<XstoreAccountInfo[ -~"\s\S\n\r\t]+accountSharedKey\s*=\s*"[^"]{30}[ -~"\s\S\n\r\t]+/>
+[<XstoreAccountInfo].*accountSharedKey\s*=\s*['"].*['"]
```

In particular, this won't work as expected because `[<XstoreAccountInfo]` will only match a single character from inside the square brackets, not `<XstoreAccountInfo`.

tgrnwd:
Definitely interested in seeing these rules included. Anything I can do to help @dvasdekis ?

dvasdekis:
These rules are complete. Not sure why they're not merged

JoostVoskuil:
@zricethezav how can we complete this?

zricethezav:
Actually, jk, there are some issues here. Most of these rules do not have `keywords`, which are necessary if gitleaks is to maintain reasonable performance. Opening up a follow up commit

dvasdekis:
Okay, I can amend the original code with some keywords, when they're available from Azure. But for those in the spec that don't have keywords, are you planning to never merge then?

nikolamalesevic:
@zricethezav This is not a way to handle this. Many of us have been waiting for this PR for a long time, with inexplicable months of silence from your side, and now jokes and fun and let's throw everything in the fire.

There must be a better way to communicate a way forward.

JoostVoskuil:
@nikolamalesevic be kind please

zricethezav:
@nikolamalesevic lol I'm really really sorry for not providing good enough free labor. get real dude

Anyways, this PR adds rules to the _default configuration_. If you want to use these rules for your own scans then _you can_, you just need to specify it using [`--config`](https://twitter.com/dog_rates/status/1669469025762684928) or one of the other methods. 

I admit, I was trigger happy and merged this PR too quickly, something that should be added is a check that makes sure all new rules have at least one keyword and that the delta in scan time does not exceed some %.

zricethezav:
@dvasdekis 
> Okay, I can amend the original code with some keywords, when they're available from Azure. But for those in the spec that don't have keywords, are you planning to never merge then?

I'd be happy to merge if the rules can be updated with keywords. The reason why keywords are so important is because gitleaks does a string compare on a chunk of text before attempting to match a regex. So if no keyword is specified, then that rule would be running regex on _every_ chunk of text, which slows things down dramatically. You can see how this is implemented here: https://github.com/gitleaks/gitleaks/blob/7804d652c0460f0b61ea9bb2c1f202bfcdbbf144/detect/detect.go#L578-L596

Feel free to reopen a PR when keywords have been added. Again, sorry for the merging then reverting.

cmd/generate/config/rules/azure.go
dvasdekis: Probably because there were like 100 rules and I blew through them with all the enthusiasm that I had, now exhausted. When I get some free time I'll try it again, with the above change

zricethezav:
> The comment was meant to be an honest feedback specific to handling of this PR (however negative), and never did I say that I do not appreciate a hard work put into this. I do apologize if I came out as being rude.

It's all good, I need a performance check to make sure new rules don't increase the scan time by some %. Again, I'm happy to review new rules if someone is up to the task.

PR #1107, opened 2023-02-25
Feat/allowlist regex target

zricethezav:
#### Allowlist Regex Targets
Let's use the generic rule to demonstrate the new `regexTarget` allowlist option

```toml
[[rules]]
description = "Generic API Key"
id = "generic-api-key"
regex = '''(?i)(?:key|api|token|secret|client|passwd|password|auth|access)(?:[0-9a-z\-_\t .]{0,20})(?:[\s|']|[\s|"]){0,3}(?:=|>|:=|\|\|:|<=|=>|:)(?:'|\"|\s|=|\x60){0,5}([0-9a-z\-_.=]{10,150})(?:['|\"|\n|\r|\s|\x60|;]|$)'''
secretGroup = 1
entropy = 3.5
keywords = [
    "key","api","token","secret","client","passwd","password","auth","access",
]
```

`example.txt` will be our target and contain a single line with a fake secret:

```txt
var discord_client_secret = '8dyfuiRyq=vVc3RRr_edRk-fK__JItpZ'
```

Running gitleaks on this file using the generic rule will return one finding:

```
gitleaks detect --source=example.txt --no-git -v --config=example.toml

    ○
    │╲
    │ ○
    ○ ░
    ░    gitleaks

Finding:     discord_client_secret = '8dyfuiRyq=vVc3RRr_edRk-fK__JItpZ'
Secret:      8dyfuiRyq=vVc3RRr_edRk-fK__JItpZ
RuleID:      generic-api-key
Entropy:     4.413910
File:        example.txt
Line:        1
Fingerprint: example.txt:generic-api-key:1
```

We can add a allowlist `regexes` entry to include part of the _secret_. This will cause gitleaks to ignore the finding above.
Note that _by default_ gitleaks uses the _Secret_ to compare against allowlist regexes.

Adding the following allowlist to the generic rule will cause gitleaks to ignore the finding:
```toml
[rules.allowlist]
regexes = ["vV"]
```

But now say you don't want to use `Secret` to compare against your allowlist regexes. Well, now you can use `regexTarget` and set the value as either `line` or `match` to compare against the line or regex match:

```toml
[rules.allowlist]
regexTarget = "match"
regexes = ["discord"]
```
and

```toml
[rules.allowlist]
regexTarget = "line"
regexes = ["var"]
```

will both result in the finding being ignored because `discord` is found in the generic rule regex _match_ and `var` is in the _line_ where the finding was found.

In addition to rule allowlists, you can set `regexTarget` in the global allowlist:

```toml
[allowlist]
regexTarget = "line"
regexes = ["var"]
```

### Checklist:


detect/detect.go
```diff
@@ -284,9 +284,16 @@ func (d *Detector) detectRule(fragment Fragment, rule config.Rule) []report.Find
...
+		// check if the regexTarget is defined in the allowlist "regexes" entry
+		allowlistTarget := finding.Secret
+		switch rule.Allowlist.RegexTarget {
+		case "match":
+			allowlistTarget = finding.Match
+		case "line":
+			allowlistTarget = finding.Line
+		}
+		if rule.Allowlist.RegexAllowed(allowlistTarget) ||
+			d.Config.Allowlist.RegexAllowed(allowlistTarget) {
```
bplaxco, line 296: @zricethezav wouldn't you need to set the allowlistTarget twice like: `ruleAllowListTarget` set by rule.Allowlist.RegexTarget and `globalAllowListTarget` set by d.Config.Allowlist.RegexTarget?

And then this check would become something like:

```go
		if rule.Allowlist.RegexAllowed(ruleAllowlistTarget) ||
			d.Config.Allowlist.RegexAllowed(globalAllowlistTarget) {
```

That way the global allowlist could set it's own regex target instead of the rule defining it which might result in weird results if different rules have different regex targets.

I could also be misunderstanding the code that you have already. Heh I'm just really dependent on the feature so I can upgrade past 7.6.1, so just making sure I understand the new change ^\_^

zricethezav: @bplaxco fantastic catch, you are absolutely right. We do need two different targets here. Pushing some changes now

PR #1130, opened 2023-03-24
Update Generic API Key rule to detect secrets with special characters

SonaliRathore:
### Description
This pull request updates the Generic API Key rule to include following special characters:  @#$%^&*<>.-? in its pattern matching algorithm. This means that gitleaks will now be able to detect secrets and API keys that contain these characters, which were previously missed by the rule.

The update was made by modifying the regular expression used by the rule to include the additional characters.

This change improves the effectiveness of gitleaks in detecting potential security vulnerabilities in our codebase. It is recommended to merge this pull request as soon as possible so that it will be able to scan maximum number of potential secrets.

Furthermore, I have thoroughly tested this update and can confirm that it will not increase the number of false positives generated by the rule.

following are some examples that gitleaks is not able to scan currently, but will be able to scan once this change is added.
private String pass1 = "AQWSder12#";
private String token = "ZBo_jDN8PL!D";


SonaliRathore:
> > Furthermore, I have thoroughly tested this update and can confirm that it will not increase the number of false positives generated by the rule.
> 
> @SonaliRathore What were your methods of testing this?

To ensure that this change does not increase the number of false positives, I followed a systematic approach.
First, I identified the repository types that I wanted to test the tool on - Spring Boot, Spring, React, NodeJs, and NodeJs with K6 framework.
For each repository type, I used “gitleaks detect --no-git --report-path report.json ” command to scan for secrets with existing tool and “gitleaks detect --no-git --report-path report.json --config gitleaks.toml” with gitleaks.toml file having the modified rule.
Then, compared both the results to see if the modified rule produced any false positives.

Following are the results:
Screenshot 2023-04-01 at 6 51 12 PM

NsovoBaloyi:
You're missing a couple of special characters, such as \x60~[]{}()|

PR #1148, opened 2023-04-13
Add detection for OpenAI API keys

Becojo:
### Description:

Add detection for OpenAI API keys. Interestingly, their API keys look made up of base64 and contains the encoded string "OpenAI" (`T3BlbkFJ`) between 2 blocks of random bytes so this rule should be pretty precise.

### Checklist:

### DIFF
--- a/cmd/generate/config/main.go
+++ b/cmd/generate/config/main.go
@@ -116,6 +116,7 @@ func main() {
 	configRules = append(configRules, rules.NPM())
 	configRules = append(configRules, rules.NytimesAccessToken())
 	configRules = append(configRules, rules.OktaAccessToken())
+	configRules = append(configRules, rules.OpenAI())
 	configRules = append(configRules, rules.PlaidAccessID())
 	configRules = append(configRules, rules.PlaidSecretKey())
 	configRules = append(configRules, rules.PlaidAccessToken())
--- a/cmd/generate/config/rules/openai.go
+++ b/cmd/generate/config/rules/openai.go
@@ -13,7 +13,6 @@ func OpenAI() *config.Rule {
 		Regex:       generateUniqueTokenRegex(`sk-[a-zA-Z0-9]{20}T3BlbkFJ[a-zA-Z0-9]{20}`),
 		SecretGroup: 1,
 		Keywords: []string{
-			"sk-",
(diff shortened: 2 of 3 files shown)


zricethezav:
Can confirm, looks like `T3BlbkFJ` is present in all keys. This might need to be checked periodically as they might change the signature?

rgmz:
@Becojo Just curious, what was the reason for removing the `sk-` keyword?

Becojo:
@rgmz It was there mostly because I copy pasted another rule. I figured `sk-` is pretty short and is likely to be present in a lot of files that don't contain an API key, wasting time running the regular expression for nothing.

Becojo:
> @zricethezav This might need to be checked periodically as they might change the signature?

They have API keys in [the scope of their bug bounty program](https://bugcrowd.com/openai) I would image they make some noise if they change the format.

rgmz:
Fair enough. The `T3BlbkFJ` _does_ seem to be a constant, however, it is not guaranteed:

> OpenAI API Keys are of the form `sk-\w{48}`. Anything not of this form will be discarded.

...That being said, the fact that it's `OpenAI` base64-encoded makes me doubt that they would remove it.

---

I just wanted to clarify because I noticed it was in `gitleaks.yaml` but not `openai.go` (#1200). Thanks for confirming!

Becojo:
oh, good catch my bad

rgmz:
> Can confirm, looks like `T3BlbkFJ` is present in all keys. This might need to be checked periodically as they might change the signature?

[Azure's OpenAI Service](https://learn.microsoft.com/en-us/azure/ai-services/openai/reference) seems to use a different token format. I discovered an instance where a private Azure endpoint was authenticated with a 32 alphanumeric characters. It's possible the api_key used an Azure AD token; I'm not familiar enough to with Azure to say for certain.

```python
"""
Reference: https://github.com/openai/openai-python#microsoft-azure-active-directory-authentication
"""
import openai

# Setup parameters
openai.api_type = "azure_ad" # I need to double-check what api_type was specified.
openai.api_base = "https://example-endpoint.openai.azure.com/"
openai.api_key = "se23rtrbvtydjt30r0cwspdcpg8a4548"
```

Issue #1059, opened 2022-12-16
gitleaksignore fingerprint doesn't play well with github's super linter


jasonkarns:
**Is your feature request related to a problem? Please describe.**

gitleaks is a linter that is bundled with (and enabled by default in) GitHub's SuperLinter action. https://github.com/github/super-linter#supported-linters

When gitleaks (via superlinter) finds a secreat, the fingerpint uses an absolute path for its filename:

`/github/workspace/REPO_ROOT/PATH/TO/FILE:private-key:1`

However, that filepath is only relevant to runs of gitleaks _from github actions_.

Adding a relative-path pattern to `.gitleaksignore` does not properly allow the secret:

`PATH/TO/FILE:private-key:1`

**Describe the solution you'd like**

I'd like there to be a way to easily ignore intentionally-committed secrets using .gitleaksignore in a manner that allows gitleaks to be run both from developer's local machine and from github actions.

**Describe alternatives you've considered**

Configuration via the .toml is one option. However, one of the primary desires of using superlinter is to not need a dozen configuration files.

**Additional context**

cc @zricethezav

Sumtin:
Running into very similar issue.  I'm running Gitleaks as part of an Azure DevOps Pipeline Decorator - a [decorator ](https://learn.microsoft.com/en-us/azure/devops/extend/develop/add-pipeline-decorator?view=azure-devops ) is just a private extension.  I'm also using self-hosted build agents.

Under this setup I have two issues:

1.  As mentioned above, inability to use relative paths makes it impossible to use the `.gitleaksignore` in a multi-agent environment.  For example, the fingerprint for one secret is `C:\azdo_winagent_1\_work\7\s\Certificates\Root-A.cer:CSCAN0020:2`, but it could be `C:\azdo_winagent_[1-5]\...` depending on how busy the agent machine is.

1. Running the Gitleaks from the decorator creates a `.json` file at the root of my project at compile-time, which is named using a random GUID.  For example, all my runs detect this secret:   `C:\azdo_winagent_1\_work\7\s\2376ab4a-328b-4141-b0f9-8ae379b199f4_39_2830.json:generic-api-key:1`

Contents of the auto-generated file are:
```
[...]

Issue #1054, opened 2022-12-09
Feature request: findings deduplication


imre-uncinc:
**Is your feature request related to a problem? Please describe.**
We have recently introduced automatic secrets scanning with Gitleaks as part of our CI pipelines. Our developers are instructed to remove and rotate any secrets that the Gitleaks CI job finds, and to then add all Gitleaks fingerprints to the .gitleaksignore file.

One thing we noticed is that the findings overview that Gitleaks outputs when running `gitleaks detect -v` contains duplicate secrets. The commit hash, date and often commit author differ, but the secrets themselves are identical. Most of these commits are merge commits.

**Describe the solution you'd like**
It would be great if `gitleaks detect -v` could deduplicate the findings (possibly enabled with a flag or config option), showing one finding per unique secret. I would prefer to only see the oldest commit so I can easily see how long the secret has been compromised.
Bonus points if this works nicely with the .gitleaksignore file, as in you only have to add one fingerprint for each unique secret and Gitleaks will ignore the other ones as well.

**Describe alternatives you've considered**
We now manually copy over all fingerprints to our .gitleaksignore files, and deduplicate them during our ISO27001 incident handling sessions.

cc @zricethezav

rgmz:
I find myself wanting this feature as well. The most straightforward way of implementing this, in my opinion, would be to generate a hash of the secret using something like blake3 or sha256. That way you can easily find all occurrences of a secret across rules and files.

Of course, this would only work if the rules and capture groups are properly made.

hackerhumble:
Hi,

Is there any update on this?

PR #1162, opened 2023-04-25
Add hash of a secret support

mn3m0nic:
### Description:

In order to de-duplicate findings and provide useful reports where only actionable leaked secrets are displayed needs to have a key field. Field "Secret" match this requirement, but when output will be unacceptably sensitive. To decrease the level of sensitivity need to use another field such as **SecretSHA** which will be just SHA256() of the Secret field.

### Checklist:


mn3m0nic:
Potentially related with https://github.com/gitleaks/gitleaks/issues/1054

detect/detect.go
```diff
@@ -272,6 +273,9 @@ func (d *Detector) detectRule(fragment Fragment, rule config.Rule) []report.Find
 			loc.endLineIndex = matchIndex[1]
 		}
 
+		// Calculate the SHA256 hash sum of secret
+		secretSHA := fmt.Sprintf("%x", sha256.Sum256([]byte(secret)))
```
rgmz, line 277: How do you feel about calling `strings.TrimSpaces` before hashing the secret? I've seen a few instances where the same secret is captured with various amounts of whitespace, and doing so would ensure they all produce the same hash.

mn3m0nic: Hi @rgmz . 
Currently, the variable Secret is trimmed by new line occurrences ('\n') and the hash in SecretSHA uses the same value from the Secret variable for consistency. If you are about to add additional trim of spaces (" ") to the Secret variable - I suggest doing it as a separate Pull request with an example of the issue.

mn3m0nic:
I vote for option 2 (if secret empty return empty string `""`) by default and option 5 when enabled as a parameter "`--hash-files`" (or something close to it TBD)

**TL;DR:**

> Why 5

**SecretSHA** is intended to be a variable that could be used as a _Key_ to sort and receive a list of unique secret leaks.
For example, for a mid-size long enough git repository you may have 10K or even 50K of RAW findings in a Git history comment, but when we de-duplicate them using the "SecretSHA" (or using sensitive "Secret") field you may receive only 20 actual unique secrets which should be rotated as a first priority from all findings. For multiple projects scan, you may find the same token, password, or RSA key was used in different places. It should not increase the number of unique findings. You still have to rotate for example 20 real keys across 100 repositories and accept dozens of findings as a new baseline.

The requirements for SecretsSHA, as I see them, are next:

1. To explicitly and unambiguously identify a secret leaked content itself
2. To be non-sensitive enough to use it as a key value in reporting on different levels (including store reports)

You are suggesting adding a filename as a key identifier, which may be ok for criteria 1, but a filename is not unique enough even with a path for a multiple projects scan. Let's assume a scan of N repositories in the group.
For example in report1.json for project1 you may have a finding for the ".env" file which will be "`e9cbb0224c4a3d23a6019ba557e0cd568c1ad5e1582ff1e335fb7d99b7a1055d`" and for project2 where the file may be completely different. I see the  **content of a file** is equal to a "Secret" field here, not a filename.

As this will bring additional performance degradation when the scan will be performed, maybe it should be enabled as an option. (TBD)

> Why 2 ("") for files, not 1?

I hope an empty hash will tell the next program which will use an output of The Gitleaks as "there is no hash here" instead here is hash "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855" which you must add to the dictionary and use it as "This is hash, but there is no value, so this is not actually usable hash, so please ignore it".
No value, no hash - will make it more clear.

rgmz:
I think you make a good argument for option 2. That was initially my first choice but I was on the fence about its purpose and usefulness.

Curious what @zricethezav thinks.

zricethezav:
Hi @mn3m0nic and @rgmz,

First, thanks for opening this PR. I think it provides a useful feature as you've clearly stated. 

> Curious what @zricethezav thinks.

At first I thought hmmm maybe we should include some logic like:
``` 
hash = "from secret"
if secret == ""
then hash = "from file"
```

but I think @mn3m0nic brought up a good point on `Why 2 ("") for files, not 1?`.

Gonna check it out locally and play around with it before merging

Devvox93:
This would solve my issues mentioned in #1059 as well. Different mechanics, but it achieves the same goal (being path-agnostic).

PR #1165, opened 2023-04-27
feat(generic api key rule): handle more use cases

burahimu:
The "generic-api-key" does not handle the "passphrase" term in case we set one. In addition, we added more chars to capture (graphical (== [!-~] == [A-Za-z0-9!"#$%&'()*+,\-./:;<=>?@[\\\]^_`{|}~])) see https://pkg.go.dev/regexp/syntax for more details.

_Note: we figured that the rule does not handle Symfony env parameters so a fix is added for that case (see
https://symfony.com/doc/current/configuration.html#configuration-based-on-environment-variables)_

### Description:

### Checklist:

### DIFF
--- a/config/gitleaks.toml
+++ b/config/gitleaks.toml
@@ -504,11 +504,11 @@ keywords = [
 [[rules]]
 description = "Generic API Key"
 id = "generic-api-key"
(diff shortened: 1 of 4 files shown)


zricethezav:
I do not want to increase the scope of the generic rule at this time as that would likely increase FPs to a rate I'm not comfortable with

PR #1179, opened 2023-05-10
Added option to specify .gitleaksignore path

### LINKED ISSUE #1178: .gitleaksignore is not considered  when scanning individual files
**Describe the bug**
When scanning individual files, gitleaks does not search for a .gitleaksignore file in same folder, neither provides a way to specify the path to one.

This breaks the function of some git hook approaches where we want to scan only the changed files. And if there's a false positive, there's no way to flag it.

**To Reproduce**
Create a folder containing false possitves, with client ids, for example.
Create a .gitleaksignore file in the same folder and add the ignore line

Run `gitleaks detect --source <path to file> --no-git`

It will flag the leak!

**Expected behavior**
I would expect it to pick a .gitleaksignore file in the same folder of scanned file

 - OS: Ubuntu 20.04
 - Gitleaks Version: 8.16.3

cc @zricethezav

pacorreia:
### Description:
Added a new option for detect/protect actions to allow specify a -gitleaksignore file path or folder containing it

### Checklist:

This PR fixes #1178

### DIFF
--- a/cmd/detect.go
+++ b/cmd/detect.go
@@ -20,7 +20,7 @@ func init() {
 	detectCmd.Flags().Bool("no-git", false, "treat git repo as a regular directory and scan those files, --log-opts has no effect on the scan when --no-git is set")
 	detectCmd.Flags().Bool("pipe", false, "scan input from stdin, ex: `cat some_file | gitleaks detect --pipe`")
 	detectCmd.Flags().Bool("follow-symlinks", false, "scan files that are symlinks to other files")
-
+	detectCmd.Flags().StringP("gitleaks-ignore-path","i",".","path to .gitleaksignore file or folder containing one")
 }
 
 var detectCmd = &cobra.Command{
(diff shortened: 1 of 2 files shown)


cmd/detect.go
```diff
@@ -81,6 +81,23 @@ func runDetect(cmd *cobra.Command, args []string) {
 		log.Fatal().Err(err).Msg("")
 	}
 
+	gitleaksPath, err:= cmd.Flags().GetString("gitleaks-ignore-path")
```
zricethezav, line 84: @pacorreia could we rename this variable to `gitleaksIgnorePath`?

cmd/protect.go
```diff
@@ -74,6 +75,23 @@ func runProtect(cmd *cobra.Command, args []string) {
 		log.Fatal().Err(err).Msg("")
 	}
 
+	gitleaksPath, err:= cmd.Flags().GetString("gitleaks-ignore-path")
```
zricethezav, line 78: @pacorreia could we rename this variable to `gitleaksIgnorePath`?

cmd/detect.go
pacorreia: Done

cmd/protect.go
pacorreia: Done

alterEgo123:
When will this be released? I'm having an issue when I try to mount .gitleaksignore as a file use the docker image

PR #1191, opened 2023-06-09
Add makefile variable detections

wayne-snyk:
### Description:

This PR adds [makefile variable detections](https://www.gnu.org/software/make/manual/html_node/Setting.html#Setting) by updating the base regex used by `generateSemiGenericRegex`.

The regenerated config is a sizable diff since this is how many of the regex detections are created.

### Checklist:

### DIFF
--- a/cmd/generate/config/rules/rule.go
+++ b/cmd/generate/config/rules/rule.go
@@ -19,7 +19,7 @@ const (
 	identifierSuffix = `)(?:[0-9a-z\-_\t .]{0,20})(?:[\s|']|[\s|"]){0,3}`
 
 	// commonly used assignment operators or function call
-	operator = `(?:=|>|:=|\|\|:|<=|=>|:)`
+	operator = `(?:=|>|:{1,3}=|\|\|:|<=|=>|:|\?=)`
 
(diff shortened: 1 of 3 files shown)


cmd/generate/config/rules/rule.go
```diff
@@ -19,7 +19,7 @@ const (
 	identifierSuffix = `)(?:[0-9a-z\-_\t .]{0,20})(?:[\s|']|[\s|"]){0,3}`
 
 	// commonly used assignment operators or function call
-	operator = `(?:=|>|:=|\|\|:|<=|=>|:)`
+	operator = `(?:=|>|:{1,3}=|\|\|:|<=|=>|:|\?=)`
```
wayne-snyk, line 22: This could also be `(?:=|>|:+=|\|\|:|<=|=>|:|\?=) for a bit more flexibility with the colons. Let me know if you prefer this to {1,3}.

zricethezav:
```
operator = `(?:=|>|:{1,3}=|\|\|:|<=|=>|:|\?=)`
```
 is good with me

PR #1199, opened 2023-06-16
Detect Azure AD client secrets

rgmz:
### Description:

This PR adds a detection for [Azure Active Directory client secrets](https://learn.microsoft.com/en-us/microsoft-365/compliance/sit-defn-azure-ad-client-secret).

<details>
<summary>Outdated issue; I've removed the allowlist</summary>

> **Warning**
> For some reason the generator does not apply the `Allowlist` to the generated config.
> I think the config template logic needs to be updated.

**Input**
https://github.com/gitleaks/gitleaks/blob/d8c0321cbb639c526c562af6a7527f888f249efa/cmd/generate/config/rules/azure.go#L25-L31

**Output**
https://github.com/gitleaks/gitleaks/blob/d8c0321cbb639c526c562af6a7527f888f249efa/config/gitleaks.toml#L139-L147

</details>

### Checklist:

* [x] Have you lint your code locally prior to submission? (`go vet`, right?)

### DIFF
--- a/cmd/generate/config/main.go
+++ b/cmd/generate/config/main.go
@@ -38,6 +38,7 @@ func main() {
 		rules.Atlassian(),
 		rules.Authress(),
 		rules.AWS(),
+		rules.AzureActiveDirectoryClientSecret(),
 		rules.BitBucketClientID(),
 		rules.BitBucketClientSecret(),
 		rules.BittrexAccessKey(),
--- a/cmd/generate/config/rules/azure.go
+++ b/cmd/generate/config/rules/azure.go
@@ -1,0 +1,65 @@
+package rules
+
+import (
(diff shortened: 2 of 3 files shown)


rgmz:
> The Generic API key rule should be able to pick up Azure secrets after this pull request is merged https://github.com/gitleaks/gitleaks/pull/1130

Eh, given that Azure AD client secrets (appear to) have a known well-defined pattern, I'd rather have a separate pattern to detect them with high confidence than assume they'll be picked up by the generic rule. The generic rule is a balancing act between false and true positives, per https://github.com/gitleaks/gitleaks/pull/1165#issuecomment-1530475275.

zricethezav (approved):

PR #1202, opened 2023-06-17
test: Check the config file drift

becojo:
### Description:

Last time I contributed to Gitleaks, I edited code in my PR and I forgot to regenerate the config file. So that doesn't happen again, I'd like to modify the test workflow so that the job fails if it detects any changes to the config file after generating it. 

Example of when it fails: https://github.com/gitleaks/gitleaks/actions/runs/5298851100/jobs/9591448117

### Checklist:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,3 +27,6 @@ jobs:
 
       - name: Validate Config
         run: cd cmd/generate/config && go run main.go
+
+      - name: Config Diff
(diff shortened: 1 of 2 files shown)


rgmz:
@zricethezav this would prevent headaches from accidental config drift.

wparad:
Why throw an error when we can automate? Rather than forcing everyone to generate the file themselves, I would recommend switching to just generating it by the build.

PR #1215, opened 2023-06-29
Make case insensitivity optional for `generateSemiGenericRegex` and `generateUniqueTokenRegex`

rgmz:
### Description:

This makes it possible to toggle whether `generateSemiGenericRegex` and `generateUniqueTokenRegex` inserts the case-insensitivity flag. There should be no functional or config changes introduced by this PR.

Context: https://github.com/gitleaks/gitleaks/pull/1204#discussion_r1240021983

### Checklist:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -39,7 +39,7 @@ If you want to add a new rule to the [default Gitleaks configuration](https://gi
 
 
            // Regex used for detecting secrets. See regex section below for more details
-           Regex: generateSemiGenericRegex([]string{"beamer"}, `b_[a-z0-9=_\-]{44}`),
+           Regex: generateSemiGenericRegex([]string{"beamer"}, `b_[a-z0-9=_\-]{44}`, true)
 
            // Keywords used for string matching on fragments (think of this as a prefilter)
            Keywords: []string{"beamer"},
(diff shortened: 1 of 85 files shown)


rgmz:
For `generateSemiGenericRegex` it might be useful to make the `identifiers` and `secretRegex` independently case-insensitive.

e.g., 
```
# Match the identifier regardless of case
SNYK_TOKEN=abc123
snyk_token=abc123
# Don't match the secret if it's uppercase
SNYK_TOKEN=ABC123
snyk_token=ABC123
```
https://pkg.go.dev/regexp/syntax

zricethezav:
@rgmz I agree with these changes -- they're simple enough and it's a drop in addition that might help folks tweak their rules a little better. At first I thought supporting this might be more work since we have a `keyword` check that _is_ case insensitive since we're converting the values `ToLower`, but, after reviewing https://github.com/gitleaks/gitleaks/blob/3cbcda20d20235c8f8744bd47a76146eae3f0efc/detect/detect.go#L260 it looks like `fragment.Raw` _is not_ converted `ToLower`. 

If you can resolve the merge conflicts, I'm happy to merge this :)

rgmz:
@zricethezav done.

Blissful304 (approved):

PR #1249, opened 2023-08-11
DRAFT: Support reading .gitleaksignore using git show

L11R:
### Description:
Gitleaks is a cool utility to use as pre-receive hook. It works nice out of box (`--log-opts` option is very handy in this case!), but it developer wants to push some false-positive anyway, there is no way to do it currently. Because Gitleaks trying to read `.gitleaksignore` only from current work tree.

My patch allows to also scan `.gitleaksignore` using `git show ...` command. Now developer can push both commit with false-positive secret and `.gitleaksignore` fix and Giteleaks will omit the problem.

Also it introduces new option: `--gitleaks-ignore-rev`. By default it will point to `HEAD`, but in case of pre-receive hooks we could not have HEAD or anyother branch yet (totally new, empty repo, or new branch). This option will allow to pass it:

```
#!/bin/bash

zero_commit='0000000000000000000000000000000000000000'

while read -r oldrev newrev refname; do
  echo "$oldrev $newrev"
  branch=${refname/#refs\/heads\/}

  # Branch or tag got deleted, ignore the push
  if [[ $newrev = "$zero_commit" ]]; then
    continue
  fi

  tmpfile=$(mktemp --suffix gitleaks)
  cmd="gitleaks detect --verbose --gitleaks-ignore-rev=$newrev --log-opts"

  # Scan only pushed commits
  if [[ $oldrev = "$zero_commit" ]]; then
    if [[ $GL_PROTOCOL = "web" ]]; then
      $cmd "$newrev --not --all" --no-color --redact |& sed 's/^/GL-HOOK-ERR: /'
    else
      $cmd "$newrev --not --all"
    fi
  else
    if [[ $GL_PROTOCOL = "web" ]]; then
      $cmd "$oldrev..$newrev --not --all" --no-color --redact |& sed 's/^/GL-HOOK-ERR: /'
    else
      $cmd "$oldrev..$newrev --not --all"
    fi
  fi
fi
```

I didn't write any additions tests for now, so it's DRAFT. Want to hear some feedback.

### Checklist:


L11R:
@zricethezav ask humbly to review it :smiley:

L11R:
I guess it's better to merge after #1250.

L11R:
Will rework it accordingly.

L11R:
@zricethezav I've added tests for bare repository, commited `.gitleaksignore` file to `small` test repo in both `main` and `foo` branches, extracted `small.git` repo from `small/dotGit` and changed in config `core.bare` to `true`.

With this patch developers know can get an error from `git push`, copy secret fingerprint, add it to `.gitleaksignore`, commit and push again. Gitleaks will see new `.gitleaksignore` using `git show`. Currently it tries to open it from FS, but remote repos don't have working tree.

L11R:
@zricethezav humbly want to bring your attention here again :)

PR #1250, opened 2023-08-11
refactor: fix #722 properly

### LINKED ISSUE #722: Gitleaks exits with exit code 0 when ambiguous argument to log-opts are provided
**Describe the bug**
Gitleaks exits with exit code 0 when ambiguous argument to log-opts are provided

**To Reproduce**
Run Gitleaks with log-opts that triggers an error from git log
 
**Expected behavior**
When error occurs I would expect exit code 1

** Logging**
8:47AM ERR fatal: ambiguous argument '18ee779d989c28192f5d439a91bcbbd3bf11dac2^..ffc5f83a4f995ed146c27fa32227ddfff251f282': unknown revision or path not in the working tree.
8:47AM ERR Use '--' to separate paths from revisions, like this:
8:47AM ERR 'git <command> [<revision>...] -- [<file>...]'
8:47AM INF no leaks found
8:47AM INF scan completed in 30.9416ms

 - OS:Osx Monterey
 - Gitleaks Version: 8.0.4

L11R:
### Description:

This PR fixes #722 issue properly.

Currently `git` package is poorly written in my humble opinion. Probably this is why `git_test.go` commented out entirely, it just hard to test because of it.

Main problem: `*exec.Cmd` requires you to run `Start()` and then `Wait()` to release resources. But both `git.GitDiff()` and `git.GitLog()` return chan with git diff files, later consumer handles chan elements and parser (!) calls `Wait()` and closes stdout pipe. This is why you can't just defer `cmd.Wait()` in these functions, because otherwise you will get blocked, consumer don't consume diff files at this moment yet.

In order to fix this we need to merge this: https://github.com/gitleaks/go-gitdiff/pull/6

We shouldn't get parser know about our implementation, it should just read from somethings, that's it.

Then we can use `bytes.Buffer` to copy stdout and defer both `cmd.Wait()` and `stdout.Close()` in `git` package directly. Also for some reason currently we don't even close `stderr`, this PR also fixes it.

To properly handle deferred `stderrErr` I use named return variables.

I understand that it slightly increases memory footprint, but I am sure it's worth it.

### Checklist:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,14 +2,10 @@ module github.com/zricethezav/gitleaks/v8
 
 go 1.19
 
-replace (
-	github.com/gitleaks/go-gitdiff => ../go-gitdiff
-)
-
 require (
 	github.com/charmbracelet/lipgloss v0.5.0
 	github.com/fatih/semgroup v1.2.0
-	github.com/gitleaks/go-gitdiff v0.8.0
+	github.com/gitleaks/go-gitdiff v0.9.0
 	github.com/h2non/filetype v1.1.3
 	github.com/rs/zerolog v1.26.1
 	github.com/spf13/cobra v1.2.1
--- a/go.mod
+++ b/go.mod
@@ -2,6 +2,10 @@ module github.com/zricethezav/gitleaks/v8
 
 go 1.19
 
+replace (
+	github.com/gitleaks/go-gitdiff => ../go-gitdiff
+)
+
 require (
 	github.com/charmbracelet/lipgloss v0.5.0
 	github.com/fatih/semgroup v1.2.0
--- a/detect/detect.go
+++ b/detect/detect.go
@@ -428,9 +428,6 @@ func (d *Detector) DetectGit(source string, logOpts string, gitScanType GitScanT
 	}
 	log.Info().Msgf("%d commits scanned.", len(d.commitMap))
 	log.Debug().Msg("Note: this number might be smaller than expected due to commits with no additions")
-	if git.ErrEncountered {
-		return d.findings, fmt.Errorf("%s", "git error encountered, see logs")
-	}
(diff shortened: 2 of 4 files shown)


go.mod
```diff
@@ -2,6 +2,10 @@ module github.com/zricethezav/gitleaks/v8
 
 go 1.19
 
+replace (
+	github.com/gitleaks/go-gitdiff => ../go-gitdiff
```
zricethezav, line 6: you can use `go-gitdiff` v0.9.0 now

L11R: @zricethezav done!

zricethezav:
> Then we can use bytes.Buffer to copy stdout and defer both cmd.Wait() and stdout.Close() in git package directly. Also for some reason currently we don't even close stderr, this PR also fixes it.


@L11R what's the point of copying? Can't we just pass `stdout` directly to `gitdiff.Parse`?

L11R:
@zricethezav `cmd.Wait()` will complete only after both `stderr` and `stdout` will empty. Otherwise it will be blocked forever, because we didn't start to read `stdout` yet. With this buffer both `stdout` and `stderr` are empty, so `cmd.Wait()` completes successfully and then parser reads diffs from buffer easily.

L11R:
Probably another option is to introduce some git.Client with struct which embeds *exec.Cmd and other stuff.

```go
client := git.NewGitLog(opts...)
diffs := client.Log()
defer client.Close()
go func() {
  for diff := range diffs { // handle findings }
}()
if client.Err() != nil {}
```

L11R:
I will send PR with tests for `git` package if it will be merged.

L11R:
@zricethezav any comments? I know you are probably busy, but still think it's good to fix this issue properly.

PR #1256, opened 2023-08-24
Detect base64-encoded JWTs

rgmz:
### Description:

This is a somewhat niche rule, but it is high confidence. It's been tested against all the standard JWT headers that I could find (e.g., https://www.iana.org/assignments/jose/jose.xhtml).

In retrospect, the rule itself could have been generated programmatically based on a list of header names. 😂 

### Checklist:

### DIFF
--- a/cmd/generate/config/main.go
+++ b/cmd/generate/config/main.go
@@ -95,6 +95,7 @@ func main() {
 	configRules = append(configRules, rules.JFrogAPIKey())
 	configRules = append(configRules, rules.JFrogIdentityToken())
 	configRules = append(configRules, rules.JWT())
+	configRules = append(configRules, rules.JWTBase64())
 	configRules = append(configRules, rules.KrakenAccessToken())
 	configRules = append(configRules, rules.KucoinAccessToken())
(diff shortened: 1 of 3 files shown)


zricethezav:
@rgmz In theory I'm all for this rule, I just want to do some benchmarking to make sure we aren't introducing a rule that significantly slows down "typical" scans.

zricethezav:
> @rgmz In theory I'm all for this rule, I just want to do some benchmarking to make sure we aren't introducing a rule that significantly slows down "typical" scans.

I should probably introduce a performance check in the CI anyways 🤷🏻

rgmz:
> @rgmz In theory I'm all for this rule, I just want to do some benchmarking to make sure we aren't introducing a rule that significantly slows down "typical" scans.

I don't expect it would, but we could always add longer prefixes to the keywords. Though wasn't there an issue with overlapping keywords?

zricethezav:
Performs great! No noticeable increase in memory or scan time. 

> I don't expect it would, but we could always add longer prefixes to the keywords. Though wasn't there an issue with overlapping keywords?

Fixed in https://github.com/gitleaks/gitleaks/pull/1259 :)

PR #1267, opened 2023-09-11
feat: catch JSON stringified

baruchiro:
### Description:
I have a code sample when someone writting a hard-coded JSON as string, and it looks like this:
```java
var json = "{ \"access-key\": \"access key here\" }"
```

### Checklist:


rgmz:
I'd recommend making the escape slash repeatable. I've encountered several instances of nested escaped JSON, which does not get detected otherwise.

baruchiro:
@rgmz Give me an example and I will TDD it :-)

PR #1280, opened 2023-10-20
Refactor Snyk detection

wayne-snyk:
### Description:
This PR updates the Snyk detection rules to be more explicit. We're seeing a number of false positives where UUIDs are used by variables with the `SNYK` prefix but whose values are not actually secret.

### Checklist:

### DIFF
--- a/cmd/generate/config/rules/snyk.go
+++ b/cmd/generate/config/rules/snyk.go
@@ -5,25 +5,35 @@ import (
 )
 
 func Snyk() *config.Rule {
+
+	keywords := []string{
+		"snyk_token",
+		"snyk_key",
+		"snyk_api_token",
+		"snyk_api_key",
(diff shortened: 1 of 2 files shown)


rgmz:
> We're seeing a number of false positives where UUIDs are used by variables with the SNYK prefix but whose values are not actually secret.

Hey @wayne-snyk, would Snyk consider adding a fixed value, such as a prefix, to their auth tokens? Having a more precise pattern would help eliminate false positives and ensure quick detection + remediation due to the higher confidence matches.

e.g., Some good examples are [Stripe](https://docs.stripe.com/keys), [GitHub](https://github.blog/2021-04-05-behind-githubs-new-authentication-token-formats/), and [OpenAI](https://community.openai.com/t/what-are-the-valid-characters-for-the-apikey/288643/4)

PR #1288, opened 2023-11-10
remove secretgroup from default config

zricethezav:
### Description:
Cleans up the default gitleaks.toml by removing the unnecessary `secretGroup` field. Note that `secretGroup` still works. 

### Checklist:


zricethezav:
tests are failing because this change results in a capture group being actually used https://github.com/gitleaks/gitleaks/blob/master/testdata/config/simple.toml#L7 

Kinda wanna go forth with this change as I would like to enforce _one_ capture group per regex by default

rgmz:
I was playing with a change that would support multiple secret groups. The main motivation being that you can do: `password: (?:("[^"]")|('[^']'))` rather than something like `password: ["'](.+)["']` or `password: ("[^"]"|'[^']')` which is less precise.

Blissful304 (approved):

Blissful304 (approved):

PR #1291, opened 2023-11-16
1079

StephenHogg:
### Description:

### Checklist:


cmd/generate/config/rules/azure.go
```diff
@@ -0,0 +1,458 @@
...
+
+	// validate
+	tps := []string{
+		generateSampleSecret("azure-storage-credential-86char",
+			secrets.NewSecret(alphaNumeric("86")+"==")),
+	}
+	return validate(r, tps, nil)
+}
+
+func AzureStorageCredential43char() *config.Rule {
```
rgmz, line 57: I've tested both these out and they're far too noisy as-is. I'm not sure how they can be improved beyond adding a prefix for "accountkey", though that may overlap with an existing pattern.

StephenHogg:
Down the bottom of this link you see the keywords that Microsoft
recommends. So perhaps we could get at least partial coverage even if we
don't use accountkey itself? I think these two rules are probably two of
the highest value rules in the Azure set so at least some effort to include
them is likely worthwhile in my opinion. Let me know what you think?

https://learn.microsoft.com/en-us/purview/sit-defn-azure-storage-account-access-key

PR #1320, opened 2024-01-03
Update stripe rule to not alert on publishable keys

brownierin:
### Description:
Stripe does a weird thing with their secrets: they have publishable keys (PK) and secret keys (SK). These abbreviations are used as the prefix for the key material, e.g., `pk_blahblahblah` and `sk_blahblahblah`. The [publishable key is deliberately not secret](https://stripe.com/docs/keys#obtain-api-keys), and it's used to identify the client to their API. Confusingly named if you're in tech; usually pk means private key.

This rule checks for both sk and pk, and it should only check for sk. I've updated it to reflect only `sk` values, which are actually secret.

### Checklist:

### DIFF
--- a/cmd/generate/config/rules/stripe.go
+++ b/cmd/generate/config/rules/stripe.go
@@ -10,12 +10,10 @@ func StripeAccessToken() *config.Rule {
 	r := config.Rule{
 		Description: "Found a Stripe Access Token, posing a risk to payment processing services and sensitive financial data.",
 		RuleID:      "stripe-access-token",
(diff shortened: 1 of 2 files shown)


brownierin:
Looks like this test is supposed to fail when modifying rules.

zricethezav:
Thanks for the PR @brownierin. The reason the test is failing is because when updating the config you have to run `go generate ./...` as documented in the contributing guidelines https://github.com/gitleaks/gitleaks/blob/master/CONTRIBUTING.md

PR #1334, opened 2024-02-02
Use latest base images.

anotherbridge:
### Description:
The current `Dockerfile` uses the golang image of version `1.19` which is deprecated (c.f. https://endoflife.date/go) as well as the alpine image version `3.16` which is soon going to be deprecated (https://endoflife.date/alpine). 
To always have the latest patched versions this PR changes the images to `latest`. If this is not desired, I would recommend that we at least use the latest version explicitly and update it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,10 @@
-FROM golang:latest AS build
+FROM golang:1.21 AS build
 WORKDIR /go/src/github.com/zricethezav/gitleaks
 COPY . .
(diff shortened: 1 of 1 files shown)


zricethezav:
Hi @anotherbridge, thanks for the PR.

> If this is not desired, I would recommend that we at least use the latest version explicitly and update it.

I'd feel better pinning a version

omercnet:
can we get a version bump to include this change ?

PR #1336, opened 2024-02-06
Some passwords are treated as false negatives

rdiezsj:
### Description:
The "generic-api-key" is modified to take into account the "/" and "+" symbols, since they are very common and with the current rule are causing false negatives.

An example password that would be correctly detected and currently is not:

```
sasl_password=LIkaITK744sG54VgABW/+p4NXFr
```

### Checklist:

### DIFF
--- a/cmd/generate/config/rules/generic.go
+++ b/cmd/generate/config/rules/generic.go
@@ -19,7 +19,7 @@ func GenericCredential() *config.Rule {
 			"password",
 			"auth",
 			"access",
-		}, `[0-9a-z\-_.=]{10,150}`, true),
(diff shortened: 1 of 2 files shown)


rgmz:
> The "generic-api-key" is modified to take into account the "/" and "+" symbols, since they are very common

To give a specific example: "/" and "+" are [base64 characters](https://base64.guru/learn/base64-characters), meaning that any base64-encoded secrets won't be detected.

rdiezsj:
With these changes, base64-encoded secrets are now detected, avoiding the generation of false negatives

dfernandezcalle (approved):

PR #1337, opened 2024-02-09
For AWS creds, added \b to assert position at a word boundary

slicingmelon:
### Description:
For AWS creds, added \b to assert position at a word boundary, otherwise too many false positives were reported.


zricethezav:
@slicingmelon please follow the contributing guideline

baruchiro:
@slicingmelon Can you please give as an example of false positives, and we will add them as tests?

Blissful304 (approved):

PR #1350, opened 2024-02-20
fix: secrets found with quotas

baruchiro:
I found some rules that returns the value ended with `"`. This happened because missing `secretGroup` value.

It is an _hot fix_.

For the future, I think we need:
1. Refactor the `validate` function or create another suite of tests with declaration of the expected found secret.
1. Fix the suffix regex to not include the `"`, because even with my change, it only fixes the `Secret` but not the `Match`


baruchiro:
@zricethezav Note this change is coming from my organization fork: [Checkmarx/gitleaks](https://github.com/Checkmarx/gitleaks).

If it helps you, I can recreate my other PRs from there.

Contact me on Discord if you want to discuss more about cooperation.

rgmz:
I believe this was resolved by #1460.

PR #1353, opened 2024-02-21
Add go.work and go.work.sum to global allowlist

AislingHPE:
### Description:

Since Go 1.18, workspace files can be used to manage workspaces with multiple Go modules. Using workspaces leads to the generation of go.work and go.work.sum files, which behave similarly to go.mod and go.sum files. Since the other Go-specific files are in the global allowlist, I think it makes sense to also add  these files.

### Checklist:


zricethezav:
love it. Thanks @AislingHPE

PR #1394, opened 2024-04-18
Create mssql.go

fhverga:
Explanation:

MSSQLDatabaseCredentials() is a function that generates the rule for detecting exposure of MSSQL database credentials. The Rule struct contains fields such as Description, RuleID, Regex, and Keywords, which define the rule. The Regex field contains the regular expression used for detecting secrets. In this case, it's looking for occurrences of Password= followed by any characters except ;. The Keywords field is used for string matching on fragments as a pre-filter. validate() function is used to validate the rule against example secrets (tps). You can provide example secrets that match the rule for testing purposes. The validate() function should return the validated rule.

### Description:

### Checklist:

### DIFF
--- a/cmd/generate/config/rules/mssql.go
+++ b/cmd/generate/config/rules/mssql.go
@@ -1,0 +1,24 @@
+// MSSQLDatabaseCredentials generates a rule for detecting exposure of MSSQL database credentials.
+func MSSQLDatabaseCredentials() *config.Rule {
+    // Define Rule
+    r := config.Rule{
+        // Human readable description of the rule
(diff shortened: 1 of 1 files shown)


fhverga:
Hello, I had created a local .yml rule to identify these types of mssql "connection" password exposure. Because originally running gitleaks on pip it was not identified. And when manually reviewing via code-review I caught this type of scenario. With that I made the regex to automate and try to contribute so that if this type of exposure comes again, gitleaks will be able to catch it and alert me.

Below is an example of the manual test.

image

Issue #1418, opened 2024-06-03
secrets mistaken as Terraform passwords


elchenberg:
**Describe the bug**

When I scan a directory that does not contain any Terraform files I still get `hashicorp-tf-password` findings.

**To Reproduce**

1. Use Gitleaks version 8.18.3.
2. Have a directory with the following `auth.json` file:

    ```json
    {
      "http-basic": {
        "github.com": {
          "username": "redacted-username",
          "password": "bE5ejpYrEoKwJzeHryx4"
        }
      }
    }
    ```

    Output of `gitleaks detect --verbose --no-color --no-git`:

    ```
    Finding:     password": "bE5ejpYrEoKwJzeHryx4"
    Secret:      "bE5ejpYrEoKwJzeHryx4"
    RuleID:      hashicorp-tf-password
    Entropy:     4.095795
    File:        auth.json
    Line:        5
    Fingerprint: auth.json:hashicorp-tf-password:5
    ```

**Expected behavior**

Expected output:

```
Finding:     password": "bE5ejpYrEoKwJzeHryx4"
Secret:      bE5ejpYrEoKwJzeHryx4
RuleID:      generic-api-key
Entropy:     4.021928
File:        auth.json
Line:        5
Fingerprint: auth.json:generic-api-key:5
```

**Basic Info (please complete the following information):**
 - OS: Linux x64
 - Gitleaks Version: 8.18.3

**Additional context**

I got the expected output when I extended the "hashicorp-tf-password" rule in my `.gitleaks.toml`:

```toml
[[rules]]
id = "hashicorp-tf-password"
path = '''\.tf$'''
```

Maybe a similar path restriction can be added to the rule in [config/gitleaks.toml](https://github.com/gitleaks/gitleaks/blob/v8.18.3/config/gitleaks.toml#L2105).

zricethezav:
Thanks for opening this issue @elchenberg. @becojo graciously opened a PR that was merged that should fix this https://github.com/gitleaks/gitleaks/pull/1420. We could probably tune the regex a little more too. Closing this issue for now.

PR #1420, opened 2024-06-13
Limit hashicorp-tf-password to .tf/.hcl files

becojo:
### Description:
Related to #1418 #1302

Limit the rule hashicorp-tf-password to only look at HCL files to avoid a lot of false positives, such as this typescript code

None of the built-in rules use a `path` to limit results so I added a new validation method to validate such rules. 

### Checklist:

### DIFF
--- a/cmd/generate/config/rules/rule.go
+++ b/cmd/generate/config/rules/rule.go
@@ -97,6 +97,34 @@ func validate(r config.Rule, truePositives []string, falsePositives []string) *c
 	return &r
 }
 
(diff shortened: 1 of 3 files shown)


cmd/generate/config/rules/rule.go
```diff
@@ -97,6 +97,34 @@ func validate(r config.Rule, truePositives []string, falsePositives []string) *c
 	return &r
 }
 
+func validateWithPaths(r config.Rule, truePositives map[string]string, falsePositives map[string]string) *config.Rule {
```
zricethezav, line 100: This is a nice function. We could consolidate this into `validate` eventually.

zricethezav (approved):

PR #1442, opened 2024-07-03
openai: detect project API keys

becojo:
### Description:
OpenAI has project API keys that use a different prefix and is not currently detected by GitLeaks.

Since these are more scoped down than user API keys, I added a new rule for them.

### Checklist:


kajogo777:
This pattern doesn't seem to work, I tried it on OpenAI proj keys

Issue #1274, opened 2023-10-09
v9.0.0 wishlist


zricethezav:
Hey all, thanks for all the support and contributions. I'm thinking about doing a breaking change, namely merging `protect` _into_ the `detect` command. I haven't committed totally to this change but it's something that's been in the back of my mind for a while now. 

If you have other suggestions or features you'd like to see added, throw 'em in this issue.

Thanks! 

Zach

Devvox93:
@zricethezav I'm not sure when the v9 is planned, but I think #1059 should definitely be in there, as it currently makes using .gitleaksignore in a CI environment rather unusable.

rgmz:
I've been meaning to respond to this for weeks. To prevent me from putting this off further, I'm going to incrementally update this comment.

Right now this isn't in any particular order and is a mix of "I feel really strongly about this" or "this is an idea I had". I will try to refine and categorize it later, as some of these are half-baked.

- #1162
    In my opinion, the current `fingerprint` implementation is brittle and trying to add everything you need to `.gitleaksignore` can be a bit of a pain-point. It would be more straightforward to "fingerprint" the secret itself and use that for `.gitleaksignore`.
- **Clarify If `keywords` Matches Path**
  I wrote this down in a note a long time ago. I forget why, and I will remove this if I can't figure it out.
- **Clarify or Rework "Stopwords"**
   In my experience using Gitleaks, as well as helping other members of the community, the concept of ["stopwords"](https://github.com/gitleaks/gitleaks#configuration) is confusing and difficult to wrap your head around. Does it match the entire secret? Is it a regular expression or a static string? Is it case sensitive? etc.
- #1212
  I don't feel particularly strongly about this, but I know it's caused issues for some people.
- **Support Multiple Allowlist Targets**
  Right now you can only chose a single `regexTarget` per allowlist (global or rule). This makes it awkward when you want to ignore multiple patterns, where one is the secret (e.g., `{{ secrets.MY_SECRET}}`) and another is the match or line (e.g., `passwordFile: ...` or `token_id: ...`). The ability to definite separate regexes for `secret` and `match` or `line` would be a nice quality-of-life improvement.
- **Support Allowlist matching `path && regex`**
  Presently, `rule.path` + `rule.regex` and `rule.allowlist.paths` + `rule.allowlist.regexes` behave differently. Unlike `rule` which matches `path && regex`, the `allowlist` will match `path || regex` which can lead to unexpected results. 
[...]

PR #1454, opened 2024-07-31
(feat): Adding secret detection rule for Kubernetes secrets

### LINKED ISSUE #1227: Detect Kubernetes Secrets
I would like to detect yaml files which contain Kubernetes secrets:

```
apiVersion: v1
data:
  foo: |
    YWMzN2Iy....
kind: Secret
metadata:
  name: foo
type: Opaque
```

cc @zricethezav

localleon:
### Description:

Fixes #1227 by implementing a simple regex which matches on the "kind: Secret" available in all Kubernetes YAML files which contain secrets. I've decided against implementing complexer logic which could also check if e.g `apiVersion: v1` and `metadata.name` are present, but this is not really doable with simple regex because the order of these values can change.

For testing purposes i created a seperate repository and commited a simple kubernetes secret. 

```bash
$ ./gitleaks detect --source ../kube-test -v

    ○
    │╲
    │ ○
    ○ ░
    ░    gitleaks

Finding:     kind: Secret
metadata:
Secret:      kind: Secret
RuleID:      kubernetes-secret
Entropy:     3.418296
File:        secret.yaml
Line:        2
Commit:      b19d024563d14451a451ad8b7df1d5e144dcffd4
Author:      localleon
Email:       me@lrau.xyz
Date:        2024-07-31T17:53:44Z
Fingerprint: b19d024563d14451a451ad8b7df1d5e144dcffd4:secret.yaml:kubernetes-secret:2

7:54PM INF 1 commits scanned.
7:54PM INF scan completed in 4.02ms
7:54PM WRN leaks found: 1
```
Which demonstrates that the new secret-detection rule works! 

Currently the output is a bit weird however, since Kubernetes secrets are whole files (and this library is more focused on detecting hardcoded tokens in source-code). I've tried experimenting with a regex which outputs the whole file content after ``kind: Secret`, but because of the non-deterministc order of the yaml this doesn't yield great results. Implementing a YAML parser here to extract the base64 encoded secret data and the name of the kubernetes secret seems overkill? 

### Checklist:


cmd/generate/config/rules/kubernetes.go
```diff
@@ -0,0 +1,26 @@
...
+func KubernetesSecret() *config.Rule {
+	// define rule
+	r := config.Rule{
+		RuleID:      "kubernetes-secret",
+		Description: "Possible Kubernetes Secret detected, posing a risk of leaking credentials/tokens from your deployments",
+		// We only match based on the "kind:secret" value of all kubernetes secrets
+		Regex: generateUniqueTokenRegex(`kind:\s*Secret`, true),
+
+		Keywords: []string{
+			"kind: Secret",
```
rgmz, line 18: This should just be "kind" or "secret", otherwise it won't match things like `kind:  Secret` (two spaces).

rgmz (commented):
I worry this may be too generic and noisy to be useful.

e.g., https://github.com/stuttgart-things/docs/blob/d0e188b6227a600bdedbf2502afc012b60cecd38/flux.md?plain=1#L81, https://github.com/grafana/grafana-image-renderer/blob/355e24a790158675260bd5f31e7338154619f8e9/.drone.yml#L494

cmd/generate/config/rules/kubernetes.go
```diff
@@ -15,6 +17,8 @@ func KubernetesSecret() *config.Rule {
 		Keywords: []string{
 			"kind: Secret",
 		},
+		// Kubernetes secrets are always yaml files, we limit to common yaml-endings to make this rule more safe!
+		Path: regexp.MustCompile(`\.(yaml|yml)$`),
```
rgmz, line 21: Can be simplified to:
```regex
\.ya?ml$
```

localleon: I thinks this simplification reduces the readability for people that are not to familiar with regex. You still want me to compress it?

cmd/generate/config/rules/kubernetes.go
localleon: Hey @rgmz, 

one of your examples e.g "flux.md" would have been catched by the path validation. But the other one would be a problem! 

I'm with you one the generic side. Is it possible to implement parsing logic with the libary e.g for yaml?

cmd/generate/config/rules/kubernetes.go
rgmz: It's the most straightforward and 'correct' way. Personally, I'd prioritize maintainability over readability to people not familiar with Regex.

Though it's up to zricethezav. I'm just some guy.

rgmz: Incidentally, it may need a [case-insensitive flag](https://stackoverflow.com/a/9655186) at the front.

```regex
(?i)\.ya?ml$
```

cmd/generate/config/rules/kubernetes.go
localleon: @rgmz i tried to fix the corresponding keywords, but the tests are failing now. I ran `go generate`, could you please take a look?

rgmz: Can you share the specific error? I don't get any errors when running it.

```sh
$ go generate ./...
$ git diff
+++ b/config/gitleaks.toml
@@ -2220,8 +2220,9 @@ keywords = [
 id = "kubernetes-secret"
 description = "Possible Kubernetes Secret detected, posing a risk of leaking credentials/tokens from your deployments"
 regex = '''(?i)\b(kind:\s*Secret)(?:['|\"|\n|\r|\s|\x60|;]|$)'''
+path = '''(?i)\.ya?ml$'''
 keywords = [
-    "kind: secret",
+    "secret",
 ]
```

localleon: Resolved it, just a typo on my side ;-)

localleon:
> I worry this may be too generic and noisy to be useful. Ideally, it would only find things that have `kind: secret` and `data:` keys in proximity.
> 
> e.g., https://github.com/stuttgart-things/docs/blob/d0e188b6227a600bdedbf2502afc012b60cecd38/flux.md?plain=1#L81, https://github.com/grafana/grafana-image-renderer/blob/355e24a790158675260bd5f31e7338154619f8e9/.drone.yml#L494

Actually i tried to improve the regex a bit more here: 

https://regex101.com/r/wU8uM7/22

`(?i)(?:\b(Secret)\b)\W+(?:\w+\W+){0,5}?\b(data)\b.*
`. But this will fail if the data key is before the secret key

rgmz:
Although I don't have time to test that right now, it seems to be on the right path.

This type of complex order-dependent pattern would be improved with "multiple secret groups" that I [mentioned here](https://github.com/gitleaks/gitleaks/issues/1274#issuecomment-1806391187).

localleon:
@zricethezav this is ready for review, please take a look!

cmd/generate/config/rules/kubernetes.go
```diff
@@ -0,0 +1,62 @@
...
+// The kubernetes rules are split into two functions to make the complex proximity matching of the data-key and the kind-identifier more readable and testable
+
+// KubernetesSecretWithDataBefore validates if we detected a kubernetes secret which contains data, before the resource identifier!
+func KubernetesSecretWithDataBefore() *config.Rule {
+	// define rule
+	r := config.Rule{
+		RuleID:      "kubernetes-secret-with-data-before",
+		Description: "Possible Kubernetes Secret detected, posing a risk of leaking credentials/tokens from your deployments",
+		// We try to match secrets by looking if we have the keyword
+		Regex: generateUniqueTokenRegex(`(?i)(?:\b(data)\b)\W+(?:\w+\W+){0,200}?\bkind:\s*Secret\b.*`, true),
```
rgmz, line 18: I _think_ this could be improved to explicitly match `data:` instead of just `data` -- unless you have an example of why that's a bad idea.

You could also include the content between `data:` and `kind` in the capture, rather than just data. Rough example: https://regex101.com/r/2Qu5ho/1

cmd/generate/config/rules/kubernetes.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+// KubernetesSecretWithDataAfter validates if we detected a kubernetes secret which contains data, after the resource identifier!
+func KubernetesSecretWithDataAfter() *config.Rule {
+	// define rule
+	r := config.Rule{
+		RuleID:      "kubernetes-secret-with-data-after",
+		Description: "Possible Kubernetes Secret detected, posing a risk of leaking credentials/tokens from your deployments",
+		// We try to match secrets by looking if we have the keyword
+		Regex: generateUniqueTokenRegex(`(?i)(?:\bkind:\s*Secret\b)\W+(?:\w+\W+){0,200}?\b(data)\b.*`, true),
```
rgmz, line 44: It _may_ be possible to update this to capture some of the actual data, similar to above.

Here's a rough attempt: https://regex101.com/r/XbJyfl/1

cmd/generate/config/rules/kubernetes.go
```diff
@@ -0,0 +1,62 @@
...
+		},
+		// Kubernetes secrets are always yaml files, we limit to common yaml-endings to make this rule more safe!
+		Path: regexp.MustCompile(`(?i)\.ya?ml$`),
+	}
+
+	// validate
+	tps := map[string]string{
+		// Sample Kubernetes Secret from https://kubernetes.io/docs/concepts/configuration/secret/
+		// These secrets contain the "data"-key before the actual identifier "kind: Secret"
+		"kubernetes.yaml": "apiVersion: v1 data: extra: YmFyCg== kind: Secret metadata: name: secret-sa-sample annotations: kubernetes.io/service-account.name: 'sa-name'",                                                                                                                // gitleaks:allow
```
rgmz, line 31: It would be good to add test cases that have quoted yaml (`kind: "Secret"`) and use 'actual' newlines between entries instead of spaces.

This can be accomplish by using `\n` inside of double-quotes, or using backticks (`) and formatting text as-is.

rgmz (commented):
Some minor suggestions, but this is looking pretty good.

cmd/generate/config/rules/kubernetes.go
localleon: You are correct about the quoting. Your suggested regex in the above comment will fail if the secret is qouted like kind: 'Secret' 

See here for an example https://regex101.com/r/VThyQ1/1

I've adjusted the regex to match quotes and added the newlines

cmd/generate/config/rules/kubernetes.go
localleon: Yeah that makes it a tiny bit better 

```bash 
$ ./gitleaks detect --source ../gitleaks-test -v

    ○
    │╲
    │ ○
    ○ ░
    ░    gitleaks

Finding:     kind: 'Secret'               
metadata:                    
  name: secret-tls           
type: kubernetes.io/tls      
data:                        
  # values are base64 enco...
  # any useful leve...
Secret:      kind: 'Secret'               
metadata:                    
  name: secret-tls           
type: kubernetes.io/tls      
data:                        
  # values are base64 enco...
RuleID:      kubernetes-secret-with-data-after
Entropy:     4.512694
File:        secret-qouted-2.yaml
Line:        2
Commit:      9ec3e13535504847c6bdeb4739b585e94fb747b6
Author:      localleon
Email:       me@lrau.xyz
Date:        2024-08-04T21:04:00Z
Fingerprint: 9ec3e13535504847c6bdeb4739b585e94fb747b6:secret-qouted-2.yaml:kubernetes-secret-with-data-after:2
```

cmd/generate/config/rules/kubernetes.go
localleon: Implemented the suggestion with minor changes

cmd/generate/config/rules/kubernetes.go
```diff
@@ -0,0 +1,68 @@
...
+// The kubernetes rules are split into two functions to make the complex proximity matching of the data-key and the kind-identifier more readable and testable
+
+// KubernetesSecretWithDataBefore validates if we detected a kubernetes secret which contains data, before the resource identifier!
+func KubernetesSecretWithDataBefore() *config.Rule {
+	// define rule
+	r := config.Rule{
+		RuleID:      "kubernetes-secret-with-data-before",
+		Description: "Possible Kubernetes Secret detected, posing a risk of leaking credentials/tokens from your deployments",
+		// We try to match secrets by looking if we have the keyword
+		Regex: generateUniqueTokenRegex(`(?i)(?:\b(data:))(\W+(?:\w+\W+){0,200}?)\bkind:.*Secret\b`, true),
```
rgmz, line 18: Data doesn't need to be a capturing group.
```suggestion
		Regex: generateUniqueTokenRegex(`(?i)(?:\b(?:data:))(\W+(?:\w+\W+){0,200}?)\bkind:.*Secret\b`, true),
```
-----
Also, I'm apprehensive about using `.*` between `kind:` and `secret`. As far as I'm aware, the only valid characters would be whitespace and an optional quote (`"'`).

I think it would be best to either limit the characters to known valid ones, or put a reasonable limit to reduce any false positives (e.g., `.{0,10}`).

rgmz (commented):
Some minor pattern nitpicks.

localleon:
> Some minor pattern nitpicks.

@rgmz to be honest, i don't have that much time anymore to spend on this PR. I'd like to leave it as is, for an inital version that we can improve on. 

I'd love to now if that's fine for @zricethezav or any other person with merge permissions

cmd/generate/config/rules/kubernetes.go
zricethezav: > Also, I'm apprehensive about using .* between kind: and secret. As far as I'm aware, the only valid characters would be whitespace and an optional quote ("').
> I think it would be best to either limit the characters to known valid ones, or put a reasonable limit to reduce any false positives (e.g., .{0,10}).

@localleon 
I agree with this. A `.{0,10}` should do the trick like @rgmz suggested.

zricethezav:
Thanks for the review @rgmz! Applied your suggestions. If this rule blows up scan times we can always disable and revisit.

rgmz:
Great stuff, @localleon!

localleon:
Please working with you @rgmz

cmd/generate/config/rules/kubernetes.go
```diff
@@ -0,0 +1,68 @@
...
+// The kubernetes rules are split into two functions to make the complex proximity matching of the data-key and the kind-identifier more readable and testable
+
+// KubernetesSecretWithDataBefore validates if we detected a kubernetes secret which contains data, before the resource identifier!
+func KubernetesSecretWithDataBefore() *config.Rule {
+	// define rule
+	r := config.Rule{
+		RuleID:      "kubernetes-secret-with-data-before",
+		Description: "Possible Kubernetes Secret detected, posing a risk of leaking credentials/tokens from your deployments",
+		// We try to match secrets by looking if we have the keyword
+		Regex: generateUniqueTokenRegex(`(?i)(?:\b(?:data:))(\W+(?:\w+\W+){0,200}?)\bkind:.{0,10}Secret\b`, true),
```
brampat, line 18: Does this also test for heuristics? Or just anything in a Kubernetes yaml-file with "kind: Secret"?
As it seems this doesn't exclude templates for those files, so it creates false positives.

rgmz: @brampat any false positive test cases are a welcome addition to #1462.

PR #1459, opened 2024-08-06
Use first non-empty group if `secretGroup` isn't set

### LINKED ISSUE #1457: Make `secretGroup` more resilient

This is based on my suggston in https://github.com/gitleaks/gitleaks/issues/1274#issuecomment-1806391187:

> Support Multiple Secret Groups
> As mentioned https://github.com/gitleaks/gitleaks/pull/1288#issuecomment-1806195451, I was experimenting with changing secretGroup: n to secretGroups: [ n1, n2, ...].
> 
> There are two potential use cases for this:
> 1. combing multiple pieces of input (e.g., username: "john"\npassword: "titor" returning both john and titor rather than just titor)
> 2. returning the first non-empty capture group when you have mutually exclusive ones.

Update the handling of `secretGroup` so that GitLeaks is able to capture secrets from more complex regexes.

Scenarios:

1. `secretGroup` is defined
    -  `secret` is based on the `secretGroup`

2. `secretGroup` is not defined
    - If the pattern has no capture groups, `secret` is the whole match
    - If the pattern has multiple capture groups,`secret` is the first non-empty capture group

This should make it possible to define regexes [like this](https://regex101.com/r/TwtY7J/1) where `secret` is consistently reported as `s3cret` and not (`"s3cret"`, `'s3cret'`, `s3cret`).

> [!NOTE]  
> This design is backwards-compatible with 8.x; it does not handle "multiple" groups mentioned in use case no. 1 of the quoted comment.

cc @zricethezav

rgmz:
### Description:

This fixes #1457.

It updates the `secretGroup` logic to use the first non-empty group if **a)** `secretGroup == 0` and **b)** the pattern has more than one capture . This allows defining patterns with slight variations, such as excluding surrounding quotes, without having to create separate rules.

## Example

Without this change, the rule would either return the entire match OR require three separate rules. Neither are ideal.

**Rule**

```yaml
[[rules]]
id = "password-test"
description = "test."
regex = '''(?i)\bpassword:\s*(?:"([^"]+)"|'([^']+)'|([[:graph:]]+))'''
keywords = [
    "password",
]
```

**File**
```
password: "NZ|u67a6'%Q?"

password: 'iY63;\7,>Tj!'

password: 1t8N9'%!uj}
```

**Output**

```bash
./gitleaks detect -c config/gitleaks.toml --no-git -s /tmp/gltest --verbose

    ○
    │╲
    │ ○
    ○ ░
    ░    gitleaks

Finding:     password: "NZ|u67a6'%Q?"
Secret:      NZ|u67a6'%Q?
RuleID:      password-test
Entropy:     3.418296
File:        /tmp/gltest/gltest.txt
Line:        1
Fingerprint: /tmp/gltest/gltest.txt:password-test:1

Finding:     password: 'iY63;\7,>Tj!'
Secret:      iY63;\7,>Tj!
RuleID:      password-test
Entropy:     3.584963
File:        /tmp/gltest/gltest.txt
Line:        3
Fingerprint: /tmp/gltest/gltest.txt:password-test:3

Finding:     password: 1t8N9'%!uj}
Secret:      1t8N9'%!uj}
RuleID:      password-test
Entropy:     3.459432
File:        /tmp/gltest/gltest.txt
Line:        5
Fingerprint: /tmp/gltest/gltest.txt:password-test:5

10:16AM INF scan completed in 9.99ms
10:16AM WRN leaks found: 3
```

### Checklist:

### DIFF
--- a/detect/detect.go
+++ b/detect/detect.go
@@ -291,24 +291,31 @@ func (d *Detector) detectRule(fragment Fragment, rule config.Rule) []report.Find
 			continue
 		}
 
-		// by default if secret group is not set, we will check to see if there
-		// are any capture groups. If there are, we will use the first capture to start
-		groups := rule.Regex.FindStringSubmatch(secret)
-		if rule.SecretGroup == 0 {
-			// if len(groups) == 2 that means there is only one capture group
-			// the first element in groups is the full match, the second is the
-			// first capture group
-			if len(groups) == 2 {
-				secret = groups[1]
-				finding.Secret = secret
+		// Set the value of |secret|, if the pattern contains at least one capture group.
+		// (The first element is the full match, hence we check >= 2.)
+		groups := rule.Regex.FindStringSubmatch(finding.Secret)
+		if len(groups) >= 2 {
+			if rule.SecretGroup > 0 {
+				if len(groups) <= rule.SecretGroup {
+					// Config validation should prevent this
+					continue
+				}
+				finding.Secret = groups[rule.SecretGroup]
+			} else {
+				// If |secretGroup| is not set, we will use the first suitable capture group.
+				if len(groups) == 2 {
+					// Use the only group.
+					finding.Secret = groups[1]
+				} else {
+					// Use the first non-empty group.
+					for _, s := range groups[1:] {
+						if len(s) > 0 {
+							finding.Secret = s
+							break
+						}
+					}
+				}
 			}
-		} else {
-			if len(groups) <= rule.SecretGroup || len(groups) == 0 {
-				// Config validation should prevent this
-				continue
-			}
-			secret = groups[rule.SecretGroup]
-			finding.Secret = secret
 		}
 
 		// check if the regexTarget is defined in the allowlist "regexes" entry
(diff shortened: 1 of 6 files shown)


detect/detect.go
```diff
@@ -291,24 +291,31 @@ func (d *Detector) detectRule(fragment Fragment, rule config.Rule) []report.Find
...
+						if len(s) > 0 {
+							finding.Secret = s
+							break
+						}
+					}
+				}
 			}
-			secret = groups[rule.SecretGroup]
-			finding.Secret = secret
 		}
```
rgmz, line 319: I don't think a fallback of `finding.Secret = group[0]` is necessary. From my testing, that's already what `secret` is.

rgmz (commented):
It would be prudent to review existing patterns and replace unnecessary capture groups.

report/sarif_test.go
```diff
@@ -76,7 +76,7 @@ func TestWriteSarif(t *testing.T) {
 			}
 			want, err := os.ReadFile(test.expected)
 			require.NoError(t, err)
-			assert.Equal(t, want, got)
+			assert.Equal(t, string(want), string(got))
```
rgmz, line 79: It's quite hard to read a diff of bytes.

rgmz: Tbh I think this works well enough. 

Converting both to string here makes the diff normal JSON instead of a stream of bytes.

zricethezav:
👍🏻 LGTM

PR #1460, opened 2024-08-06
Remove unnecessary capture groups

rgmz:
### Description:

This removes unnecessary capture groups in patterns.

"Unnecessary" meaning:
- a group that wraps the entire pattern: `regex.MustCompile("(foo.bar)")` -> `regex.MustCompile("foo.bar")`
- a group that's solely used for "OR":  `regex.MustCompile("foo(bar|baz)")` ->  `regex.MustCompile("foo(?:bar|baz)")`

### Checklist:

### DIFF
--- a/cmd/generate/config/rules/clojars.go
+++ b/cmd/generate/config/rules/clojars.go
@@ -12,7 +12,7 @@ func Clojars() *config.Rule {
 	r := config.Rule{
 		Description: "Uncovered a possible Clojars API token, risking unauthorized access to Clojure libraries and potential code manipulation.",
 		RuleID:      "clojars-api-token",
-		Regex:       regexp.MustCompile(`(?i)(CLOJARS_)[a-z0-9]{60}`),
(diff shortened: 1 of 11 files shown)


romain-gaillard:
Ah! I just found an issue with this yesterday and was about to create a PR now.

It's very important to fix these because otherwise, it creates issues with the redaction feature. As an example, the GitHub regex `(ghu|ghs)_[0-9a-zA-Z]{36}` will make Gitleaks only redact `gpu`or `ghs` but not the rest of the token (as it works by replacing the first capture group if there's one). Which is problematic!

zricethezav:
Fantastic work as always @rgmz, thank you. I also appreciate the definition of "unnecessary" here.

> "Unnecessary" meaning:
> a group that wraps the entire pattern: regex.MustCompile("(foo.bar)") -> regex.MustCompile("foo.bar")
> a group that's solely used for "OR": regex.MustCompile("foo(bar|baz)") -> regex.MustCompile("foo(?:bar|baz)")

PR #1461, opened 2024-08-06
Return non-0 exit code  from `DetectGit`

### LINKED ISSUE #1450: Invalid --log-opts disregards git log exit code and silently does nothing at all
**Describe the bug**

Gitleaks doesn't seem to consider the git log exit code. When `--log-opts` is set to an invalid value, gitleaks will echo the stderr message but [exit with 0 and claim no leaks were found](https://github.com/gitleaks/gitleaks/blob/master/cmd/detect.go#L90). This is fatal in automated scenarios like CI or precommit hook because users must now rely on spotting the output instead of reacting to the process exit code which is what automation relies on to detect failures.
If gitleaks can't get a valid git log, the entire scan is compromitted and should report an error IMO.

Note: fixing this issue will probably also affect running `gitleaks detect` in a `cwd` that is not a git repository since for some reason this also does `exit 0` despite not scanning anything.

**To Reproduce**

```bash
gitleaks detect --log-opts=--invalid
```

**Expected behavior**
Exit 1, because `git log --invalid` doesn't work and exits with code 128.
It would be fine if there was a `--fail-on-git-error` option or similar, similar to the way curl handles status errors. This would preserve existing behaviour and users that rely on it while allowing automation around gitleaks to become more robust.

**Screenshots**

```bash
» gitleaks detect --log-opts=--invalid

    ○
    │╲
    │ ○
    ○ ░
    ░    gitleaks

7:03AM ERR [git] fatal: unrecognized argument: --invalid
7:03AM ERR  error="stderr is not empty"
7:03AM INF scan completed in 37ms
7:03AM INF no leaks found

» echo $?
0
```

 - OS: OSX
 - Gitleaks Version: 8.18.4

**Additional context**

- In technical terms the issue is that `detector.DetectGit` replaces the previous `err` with `nil` by scanning zero files, which succeeds trivially.
- If we agree on how gitleaks should behave and it helps move things forward, I'm open to creating a pull request to adjust behaviour

cc @zricethezav

rgmz:
### Description:
This fixes #1450.

The issue _appears_ to be that `err` in the outer scope is being shadowed, meaning that `findingSummaryAndExit` always receives a `nil` value.

https://github.com/gitleaks/gitleaks/blob/0334ec10760314034822710507631d0f6d3acecb/cmd/detect.go#L28-L31

https://github.com/gitleaks/gitleaks/blob/0334ec10760314034822710507631d0f6d3acecb/cmd/detect.go#L88

https://github.com/gitleaks/gitleaks/blob/0334ec10760314034822710507631d0f6d3acecb/cmd/detect.go#L99

**New Behaviour**

```sh
$ ./gitleaks detect --log-opts="--invalid"  --verbose -s /tmp/gitleaks

    ○
    │╲
    │ ○
    ○ ░
    ░    gitleaks

7:51PM ERR [git] fatal: unrecognized argument: --invalid
7:51PM ERR failed to scan Git repository error="stderr is not empty"
7:51PM WRN partial scan completed in 17.8ms
7:51PM WRN no leaks found in partial scan

$ echo $?
1
```

### Checklist:

### DIFF
--- a/cmd/detect.go
+++ b/cmd/detect.go
@@ -39,7 +39,7 @@ func runDetect(cmd *cobra.Command, args []string) {
 	// grab source
 	source, err := cmd.Flags().GetString("source")
 	if err != nil {
-		log.Fatal().Err(err).Msg("")
+		log.Fatal().Err(err).Msg("could not get source")
 	}
 	detector := Detector(cmd, cfg, source)
 
@@ -58,7 +58,7 @@ func runDetect(cmd *cobra.Command, args []string) {
 	}
 	fromPipe, err := cmd.Flags().GetBool("pipe")
 	if err != nil {
-		log.Fatal().Err(err)
+		log.Fatal().Err(err).Msg("could not call GetBool() for pipe")
 	}
 
 	// start the detector scan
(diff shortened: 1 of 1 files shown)


zricethezav:
@rgmz great catch!

PR #1462, opened 2024-08-07
Refactor `kubernetes-secret` rule

rgmz:
### Description:
This is a follow-up to #1454.

Changes:
- Leverages #1459 to collapse `kubernetes-secret-with-data-before` and `kubernetes-secret-with-data-after` into a single rule
- Tightens regex to reduce the likelihood of false-positives
- Handles multi-line secret definitions (e.g., https://kubernetes.io/docs/concepts/configuration/secret/#docker-config-secrets, https://kubernetes.io/docs/concepts/configuration/secret/#ssh-authentication-secrets)
- Partially handles cases with comments at the end of lines, or comments between lines.  

**TODO**
- [ ] Add a rule for [`stringData`](https://kubernetes.io/docs/concepts/configuration/secret/#docker-config-secrets) in the future.
- [x] Check for a minimum value length
- [x] More test cases?
  - [x] sopssecret
  - [x] ExternalSecret
  - [x] Templates? (e.g. not base64 encoded value)

### Checklist:


rgmz:
@marcm-ml @brampat @fabio-sv This seems fairly complete to me. If you have a moment, give the new rule a try and let me know if it fixes the false-positives you're seeing.

brampat:
@rgmz Thank you for the quick response. I've tested it as so:
* downloaded the raw toml file from here to my project folder: https://github.com/gitleaks/gitleaks/blob/b9f875c83a732d00c4ce1d796ac092b68eb91c4d/config/gitleaks.toml
* ran gitleaks with -c gitleaks.toml, which gave no leaks found
* ran gitleaks without the -c gitleaks.toml (default config), which again gave the 3 false positives.

So, unless my testing method is off, this seems like it solves the false positives for me.

marcm-ml:
My false-positives are fully resolved by this. Thanks for the quick fix 👍 

Outside the scope of this PR:
It might be interesting to add a rule for special resources such as ExternalSecret or S SopsSecrets and work on an absence of special characters. For example, one could flag an ExternalSecret resource if for each key within the spec.target.template.data field "{{" and "}}" are missing. Similarly, for SopsSecret it would be sufficient to check for the "sops:" field. Of course, you could also add a custom .gitleaks.toml for these personal usecases.

Would like to hear you thoughts about this.

brampat:
If I understand correctly @marcm-ml : See if rules can detect replacement-tags / template usage and in those cases, don't mark those as scanned secrets, since it'll contain a key that references a secret stored somewhere else (eg. in a vault).

rgmz:
> It might be interesting to add a rule for special resources such as ExternalSecret or SopsSecrets and work on an absence of special characters. For example, one could flag an ExternalSecret resource if for each key within the spec.target.template.data field "{{" and "}}" are missing. Similarly, for SopsSecret it would be sufficient to check for the "sops:" field. Of course, you could also add a custom .gitleaks.toml for these personal usecases.
> 
> Would like to hear your thoughts about this.

I created #1513 to track some follow-up ideas. I don't have personal experience with either of these; if you can provide examples of 'true' positives, that would be useful for future reference.

brampat:
I've upgraded to gitleaks 8.9.2 and scanned the repo again. Unfortunately it now triggers the new rule again.

File: 

`
apiVersion: v1
kind: Secret
metadata:
  name: nexus-pull-secret
type: kubernetes.io/dockerconfigjson
data:
  .dockerconfigjson: {{ template "imagePullSecret" . }}
`

Violates (pruned the committer info):
`
Finding:     _kind: Secret
metadata:
  name: nexus-pull-secret
type: kubernetes.io/dockerconfigjson
data:
  .docke..._
Secret:      _kind: Secret
metadata:
  name: nexus-pull-secret
type: kubernetes.io/dockerconfigjson
data:
  .docke..._
RuleID:      kubernetes-secret-with-data-after
Entropy:     4.615095
       helm/petclinic-jboss/templates/secret.yaml
Line:        2
`

rgmz:
I think this is still pending release.

+ @zricethezav

brampat:
Are you sure? I've checked the commit log for that release:
https://github.com/gitleaks/gitleaks/compare/v8.19.2...master

And I see the refactoring stuff is in there.

rgmz:
> Are you sure? I've checked the commit log for that release: [v8.19.2...master](https://github.com/gitleaks/gitleaks/compare/v8.19.2...master)
> 
> And I see the refactoring stuff is in there.

It's a bit confusing. Those are the changes that have happened _since_ that release.
image

Regardless, the example you shared had `"` inside templates `{{ }}` which may have still caused an issue. I opened #1520 to address that.

PR #1463, opened 2024-08-07
Use rule id for config validation error

rgmz:
### Description:
This updates a validation error to use the rule ID instead of the description. It also adds a check to ensure that the ID isn't blank.

### Checklist:

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -82,6 +82,10 @@ func (vc *ViperConfig) Translate() (Config, error) {
 	rulesMap := make(map[string]Rule)
 
 	for _, r := range vc.Rules {
+		if strings.TrimSpace(r.ID) == "" {
+			return Config{}, fmt.Errorf("rule ID is missing or empty, regex: %s", r.Regex)
+		}
+
 		var allowlistRegexes []*regexp.Regexp
 		for _, a := range r.Allowlist.Regexes {
 			allowlistRegexes = append(allowlistRegexes, regexp.MustCompile(a))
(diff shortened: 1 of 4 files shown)


zricethezav:
Excellent change. Thanks!

matiasmoyano:
question about this fix, im deploying something and while using gitleaks it crashes

3:59PM FTL Failed to load config error="rule ID is missing or empty, regex: "
--
100 |  
101 | [Container] 2024/08/12 15:59:16.668010 Command did not exit successfully gitleaks detect --source . -v exit status 1
102 | [Container] 2024/08/12 15:59:16.672966 Phase complete: BUILD State: FAILED
103 | [Container] 2024/08/12 15:59:16.672982 Phase context status code: COMMAND_EXECUTION_ERROR Message: Error while executing command: gitleaks detect --source . -v. Reason: exit status 1

my toml file is
[rules.allowlist] 
description = "ignore commits"
commits = ["commit"]

should i add an extend tag or at the least a rule so it doesnt crash?

rgmz:
What are you trying to accomplish? That configuration does not seem correct.

matiasmoyano:
a config just to ignore a commit id, the reason is, the specific commit has a rotated secret, but if we remove that, gitleaks complains, but that config no longer works, maybe i need to add something else to avoid this new error message, add something else to the tom file:
[rules.allowlist]
description = "ignore commits"
commits = ["commit"]

rgmz:
`rules.allowlist` only works when nested under a specific instance of `[[rules]`. You may want the global `[allowlist]`.

matiasmoyano:
> `rules.allowlist` only works when nested under a specific instance of `[[rules]`. You may want the global `[allowlist]`.

it worked!

thank you sr!

PR #1466, opened 2024-08-14
Update config rule validation logic

rgmz:
### Description:
This adds a check preventing both `regex` and `path` from being undefined.

It also moves the validation logic into its own function. I think that makes more sense but LMK. 

### Checklist:

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -2,7 +2,6 @@ package config
 
 import (
 	_ "embed"
-	"fmt"
 	"regexp"
 	"sort"
 	"strings"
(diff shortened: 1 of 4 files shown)


zricethezav:
This looks great, thanks @rgmz

robertojdelgado:
we are experiencing FTL Failed to load config error="rule |id| is missing or empty, regex: (?i)(?:AccessToken|pat|token).*?[':=\"][a-z0-9/+]{70}==(?:'|\"|\\s|[\\r?\\n]+)"

rgmz:
@robertojdelgado #1507 is related to that. This is a different issue.

PR #1472, opened 2024-08-17
Handle when Read returns both n > 0 bytes and EOF

### LINKED ISSUE #1400: Secret finding is not detected when Read operation returns both n > 0 bytes and EOF
**Describe the bug**
I have observed a scenario where a secret string is not properly detected if the [Read() operation](https://github.com/gitleaks/gitleaks/blob/master/detect/reader.go#L17) returns n > 0 bytes and the EOF flag.

**Additional information:**
According to the golang v1.22.2 [documentation](https://pkg.go.dev/io#Reader):

`a Reader returning a non-zero number of bytes at the end of the input stream may return either err == EOF or err == nil. `

The documentation also states that:
`Callers should always process the n > 0 bytes returned before considering the error err.`

**To Reproduce**
I have implemented the following unit test to help reproducing the issue :
```
package detect

import (
	"fmt"
	"io"
	"strings"
	"testing"

	"github.com/stretchr/testify/assert"
)

type mockRead func(secret []byte) (int, error)

func (m mockRead) Read(secret []byte) (int, error) {
	return m(secret)
}

// TestDetectReader tests the DetectReader function.
func TestDetectReader(t *testing.T) {
	var detector *Detector
	var secret = "AKIAIRYLJVKMPEXAMPLE"

	testString := strings.NewReader(secret)
	tests := []struct {
		name          string
		reader        io.Reader
		bufSize       int
		findingsCount int
	}{
		{
			name:          "Test case - Reader returns n > 0 bytes and nil error",
			bufSize:       10,
			findingsCount: 1,
			reader:        testString,
		},
		{
			name:          "Test case - Reader returns n > 0 bytes and io.EOF error", // this test case is failing with the current implementation
			bufSize:       10,
			findingsCount: 1,
			reader: mockRead(func(secret []byte) (int, error) {
				return 20, io.EOF
			}),
		},
	}

	for _, test := range tests {
		fmt.Printf("Running test case: %v\n", test.name)
		detector, _ = NewDetectorDefaultConfig()
		findings, _ := detector.DetectReader(test.reader, test.bufSize)
		assert.Equal(t, test.findingsCount, len(findings))
	}
}
```

**Expected behavior**
I am expecting the secret to be detected in both of these scenarios:
- n > 0 bytes returned, err == nil
- n > 0 bytes returned, err == io.EOF

**Screenshots**
N/A

 - OS: Sonoma 14.3.1 Darwin Kernel Version 23.3.0
 - Gitleaks Version: v8.18.2

**Additional context**

cc @zricethezav

rgmz:
### Description:
This fixes #1400.

### Checklist:

### DIFF
--- a/detect/reader.go
+++ b/detect/reader.go
@@ -15,22 +15,27 @@ func (d *Detector) DetectReader(r io.Reader, bufSize int) ([]report.Finding, err
 
 	for {
 		n, err := reader.Read(buf[:cap(buf)])
-		buf = buf[:n]
+
+		// "Callers should always process the n > 0 bytes returned before considering the error err."
+		// https://pkg.go.dev/io#Reader
+		if n > 0 {
+			buf = buf[:n]
+			fragment := Fragment{
+				Raw: string(buf),
+			}
+			for _, finding := range d.Detect(fragment) {
+				findings = append(findings, finding)
+				if d.Verbose {
+					printFinding(finding, d.NoColor)
+				}
+			}
+		}
+
 		if err != nil {
 			if err != io.EOF {
 				return findings, err
 			}
 			break
-		}
-
-		fragment := Fragment{
-			Raw: string(buf),
-		}
-		for _, finding := range d.Detect(fragment) {
-			findings = append(findings, finding)
-			if d.Verbose {
-				printFinding(finding, d.NoColor)
-			}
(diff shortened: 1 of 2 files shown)


zricethezav:
💯

PR #1478, opened 2024-08-22
Update logic

rgmz:
### Description:
This updates some of the existing logic.

I've tried to contain each change to a distinct commit, to make review or removal/editing easier.

Tentative TODO list:
- [ ] Fix printing for color/no-colour
- [ ] Put status/reason being printed behind feature flag
- [ ] Put marshaling status/reason behind feature flag
- [ ] Test if verification requests have timeout set (don't want to block forever)
- [ ] Test if verify works with combinations of expect status/body
- [x] Decide whether to keep/remove explicit `requires`
    - [x] Perhaps parse the rule at startup first?
- [ ] Decide whether to keep/change the `report` field
- [ ] General test cases for new feature
- [ ] Filtering for result types: valid, invalid, skipped, error?

### Checklist:

* [ ] ensure nothing related to verification runs unless the flag is enabled


zricethezav:
@rgmz updated a couple small things, lemme know whatcha think

rgmz:
> @rgmz updated a couple small things, lemme know whatcha think

Good catch. :+1: 

I don't use redact so I didn't catch that obvious issue in my testing.

zricethezav:
well that's odd and a little alarming. I did not mean to merge but hey at least it's in a feature branch

PR #1480, opened 2024-08-23
Make config generation utils public

rgmz:
### Description:

I use these to generate my own config from code (vs. using YAML) and can see them being useful to others for the same reason.

These would be "use at your own risk", with no guarantee against breaking changes. Perhaps worth adding a note somewhere?

### Checklist:


zricethezav:
@rgmz this change lgtm.

> These would be "use at your own risk", with no guarantee against breaking changes. Perhaps worth adding a note somewhere?

That makes sense to me. The best place for that note would be in the `cmd/generate/rules/rule.go` file.

rgmz:
> That makes sense to me.

Done. :+1:

PR #1483, opened 2024-08-23
Update Hashicorp Vault token pattern

rgmz:
### Description:
This updates the Hashicorp Vault and Terraform patterns.

1. Adds the old vault `s.` format
2. Tweaks the new vault `hvs.` format, including making it a bit longer as I've found instances in the wild > 100 chars
3. Adds entropy to terraform

### Checklist:

### DIFF
--- a/cmd/generate/config/main.go
+++ b/cmd/generate/config/main.go
@@ -100,7 +100,7 @@ func main() {
 		rules.GrafanaCloudApiToken(),
 		rules.GrafanaServiceAccountToken(),
 		rules.HarnessApiKey(),
-		rules.Hashicorp(),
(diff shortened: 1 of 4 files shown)


zricethezav:
🙇🏻

egibbm:
Can we not use a single character keyword? 

The problem with single character keyword is, it can be generated  by any code minifier/uglifier, and function names with 24 chars can easily be caught using this rule?

In our case, this rule will match the following string in a generated js asset:
`s.multilineTagIndentFactor|`

My workaround is to put that specific string to the global regexes to exclude the line.

Do you have other suggestion for our case?

rgmz:
See #1490

Issue #807, opened 2022-03-18
Discover secrets in base64 encoded strings


oleyv-cl:
**Is your feature request related to a problem? Please describe.**
No. 

**Describe the solution you'd like**
It is currently possible to search for base64 encoded string. After the base64 sting is found, would it be possible to decode the string and pass it through the gitleaks detection logic again in the decoded form to discover potential secrets within base64?

**Describe alternatives you've considered**
Get the base64 string to stdout; pipe the output to base64 -d; pipe the output to gitleaks --no-git. 

**Additional context**
Thank you very much for a great and very flexible tool. 

cc @zricethezav

zricethezav:
> Get the base64 string to stdout; pipe the output to base64 -d; pipe the output to gitleaks --no-git.

This feature request sounds more like you want me to add support for accepting piped input, is that correct?

OleyV:
Re-reading the request I've realized that piped input will definitely help without overcomplicating the code logic.

sergiomarotco:
I vote for this feature

leite08:
It would be super useful for our use case to be able to run a custom configuration against the result of a decoded base64 string.

I'm new to this project so I don't know whether this would make sense to how the underlying engine works, but by looking at the default config it seems we could have a "decodeBase64" flag to indicate the respective config/rule's regex should be applied at the output of base64 decode.

To make it more flexible, instead of a flag we could have a rule setting called `pre-processor` where one would indicate which pre-processor to apply to that rule from a list of supported pre-processors (one being "base64Decoder").

Happy to create a new issue if this comment doesn't apply to the current issue.

rgmz:
> It would be super useful for our use case to be able to run a custom configuration against the result of a decoded base64 string.
> To make it more flexible, instead of a flag we could have a rule setting called pre-processor where one would indicate which pre-processor to apply to that rule from a list of supported pre-processors (one being "base64Decoder").

Pre-processors have many potential use cases — such as decoding base64 (as mentioned), decoding JWTs, or deflating archives — but would the feature would need to be carefully designed and implemented to preserve performance.

@zricethezav what are your thoughts on adding a net-new feature like this? 

----

It's worth noting that [TruffleHog](https://github.com/trufflesecurity/trufflehog) has built-in decoders for Base64 and archives, plus rules are written programmatically (rather than in TOML) which allows for additional flexibility and functionality.

imre-uncinc:
[...]

PR #1488, opened 2024-08-27
Decode Base64

bplaxco:
### Description:

Decode base64 values and run the rules against them (see https://github.com/gitleaks/gitleaks/issues/807)

This is still a WIP. It needs more unittests, performance tests, and a good once over. I'm submitting the draft MR to get some eyes on the progress though.

When finished this should:

- Exclude existing matches (or at least most of them)
- Decode b64 encoded values (both standard and b64url and even values encoded multiple times)
- Report the boundaries of encoded text in the original raw text for the location.

### Checklist:


detect/decoder.go
```diff
@@ -0,0 +1,153 @@
...
+	//
+	relStart     int
+	relEnd       int
+	absStart     int
+	absEnd       int
+	decodedValue string
+}
+
+func init() {
+	b64Regexp = regexp.MustCompile(`[\w\/\-\+]{16,}={0,3}`)
```
rgmz, line 36: Just curious, why do this in an init block instead of directly assigning it?

bplaxco: Ah, didn't realize I could. I'll tweak that. Good catch.

bplaxco: I'll also change the other place where I did that as well.

detect/detect.go
```diff
@@ -247,23 +273,34 @@ func (d *Detector) detectRule(fragment Fragment, rule config.Rule) []report.Find
...
-		// extract secret from match
-		secret := strings.Trim(fragment.Raw[matchIndex[0]:matchIndex[1]], "\n")
+	// use currentRaw instead of fragment.Raw since this represents the current
+	// decoding pass on the text
+	for _, matchIndex := range rule.Regex.FindAllStringIndex(currentRaw, -1) {
+		// Extract secret from match
+		secret := strings.Trim(currentRaw[matchIndex[0]:matchIndex[1]], "\n")
+
+		if len(encodedSegments) > 0 {
+			if segment := segmentContaining(encodedSegments, matchIndex[0], matchIndex[1]); segment != nil {
```
bplaxco, line 292: There is a bug here and also I think this should be overlaps instead of contains.

I didn't account for some of the relative text shifting. I'm working on a fix.

detect/decoder.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+// findEncodedSegments finds the encoded segments in the data and maps them to
+// any parent segments from a previous pass.
+func findEncodedSegments(data string, parentSegments []EncodedSegment) []EncodedSegment {
+	if len(data) == 0 {
+		return []EncodedSegment{}
+	}
+
+	matchIndices := b64Regexp.FindAllStringIndex(data, -1)
```
rgmz, line 63: Would it be paranoid to do a second pass and filter out any values that are all lowercase or uppercase? (Random thought.)

I suspect this would filter a large number of false positives without the overhead of calling base64 decode.

bplaxco: Good note!

I'm planning to do some performance tests to see what I can do to bail out as early as possible with real world code, but I hadn't considered specifically that one.

My current list of ideas to try in different combinations to see if they help speed it up:

- have the `decodedMap` persist across the different passes so that only new things would be decoded
- entropy check (if it won't filter out real b64)
- (newly added) the ideas you just listed
- other misc checks that look at what's highly unlikely in things that aren't base64 (what those are are TBD and I'd only do that if the above didn't get us near the same performance as gitleaks without the changes).

detect/detect.go
```diff
@@ -247,19 +271,32 @@ func (d *Detector) detectRule(fragment Fragment, rule config.Rule) []report.Find
...
+	// decoding pass on the text
+	for _, matchIndex := range rule.Regex.FindAllStringIndex(currentRaw, -1) {
+		// Extract secret from match
+		secret := strings.Trim(currentRaw[matchIndex[0]:matchIndex[1]], "\n")
+
+		// Check if the decoded portions of the segment overlap with the match
+		// to see if its potentially a new match
+		if len(encodedSegments) > 0 {
+			if segment := segmentWithDecodedOverlap(encodedSegments, matchIndex[0], matchIndex[1]); segment != nil {
+				matchIndex[0] = segment.absoluteStart
```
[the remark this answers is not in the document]
bplaxco: :heavy_check_mark: Done

detect/detect.go
```diff
@@ -22,8 +22,11 @@ import (
 const (
 	gitleaksAllowSignature = "gitleaks:allow"
 	chunkSize              = 10 * 1_000 // 10kb
+	maxDetectPasses        = 3          // remove the chance of a decode infinite loop
```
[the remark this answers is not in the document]
bplaxco: :heavy_check_mark: Done

detect/decoder.go
bplaxco: Generating some stats to figure out how likely different situations are.

`test-data` file:

```sh
git clone https://github.com/kubernetes/kubernetes.git
git -C kubernetes/ log --patch --all > test-data
```

Test script:

```python3
from collections import defaultdict
from secrets import randbits
from base64 import b64encode
from string import ascii_lowercase
from string import ascii_uppercase
from string import digits

lowercase = set(ascii_lowercase)
uppercase = set(ascii_uppercase)
digits = set(digits)
special = {'+', '/'}

count = defaultdict(int)
with open("test-data", "rb") as test_file:
    while True:
        data = test_file.read(max(randbits(8), 16))

        if not data:
            break

        enc = set(b64encode(data).decode())

        if enc & lowercase:
            count["lower"] += 1

        if enc & uppercase:
            count["upper"] += 1

        if enc & digits:
            count["number"] += 1

        if enc & special:
            count["special"] += 1

        if enc & (lowercase | uppercase) and not enc & (digits | special):
            count["only_letters"] += 1

        if enc & (lowercase | uppercase | digits) and not enc & special:
            count["no_special"] += 1

        count["total"] += 1

print(count)
```

Totals coming soon.

bplaxco: My hunch is that letters only is going to be rare for base64 data and we can skip that and I can move the decoded map into Detect or something for reuse across recursive scans and then things will be [blazingly fast](https://www.youtube.com/watch?v=Z0GX2mTUtfo).

bplaxco: Results:

```
{'lower': 28142553, 'upper': 28162027, 'number': 27866020, 'no_special': 26757416, 'total': 28164584, 'special': 1407168, 'only_letters': 294527}
```

So if we ignored everything with only letters in it we'd miss about 1% of the base64 data.

I'd be okay with that if there's a significant performance boost out of it and what we have too now is slow.

I'm going to see what my [test script](https://github.com/gitleaks/gitleaks/issues/807#issuecomment-2380045683) spits out for what I have now.

Also looking at the data above, it's basically a given that base64 is going to have upper or lower case letters if we wanted to ignore things like long runs of numbers and stuff.

bplaxco:
Hmm something's failing after a rebase. Looking into that.

bplaxco:
> Hmm something's failing after a rebase. Looking into that.

This fixes it, there was a change on master that adjusted a matchIndex and that conflicted with the adjustments that my code makes.

https://github.com/gitleaks/gitleaks/pull/1488/files#diff-843a307bde5ddd61294c58685e121707f08adfacc34e49f7399cfc6e42e57947R300

bplaxco:
I'll post the performance test results once the tests complete, but the first pass looks really promising (I'm having it do two rounds of several hours of scanning) so I'm going to go ahead and mark this as ready for review.

detect/decoder.go
```diff
@@ -0,0 +1,256 @@
...
+
+	segments := make([]EncodedSegment, 0, len(matchIndices))
+
+	// Keeps up with offsets from the text changing size as things are decoded
+	decodedShift := 0
+
+	for _, matchIndex := range matchIndices {
+		encodedValue := data[matchIndex[0]:matchIndex[1]]
+
+		if !isLikelyB64(encodedValue) {
```
bplaxco, line 163: This may not provide a lot of performance gains and can miss around 1% of base64 under (it might be less than that in real world cases but I dumped `git log --all --patch` in the kubernetes repo for test-data and then encoded it all in varying lengths and and it came out to missing around 1% in my test.

I can take this bit out if we want. `d.decodedMap` also helps speed things up as well so this might not be needed. Thoughts?

bplaxco:
Side quirk I noticed: if a single pattern matches across multiple encoded segments the end column can be a little off (it's a little short). I know how I'd fix that but also its up to y'all if that's a case you expect to happen very often.

rgmz (commented):
This looks solid over all, passed all my base64-encoded test cases.

Perhaps it would be useful to attach metadata (i.e., this was base64-decoded _n_ times)

cmd/root.go
```diff
@@ -56,6 +56,7 @@ func init() {
 	rootCmd.PersistentFlags().Bool("no-banner", false, "suppress banner")
 	rootCmd.PersistentFlags().StringSlice("enable-rule", []string{}, "only enable specific rules by id")
 	rootCmd.PersistentFlags().StringP("gitleaks-ignore-path", "i", ".", "path to .gitleaksignore file or folder containing one")
+	rootCmd.PersistentFlags().Int("max-decode-depth", 0, "allow recursive decoding up to this depth (if depth is 0 no decoding is done)")
```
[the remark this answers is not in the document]
zricethezav: `0` makes sense to me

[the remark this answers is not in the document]
rgmz: Mhm, although `0` currently means disabled.

detect/decoder.go
zricethezav: I like it. Let's leave it in. If it ends up filtering out too much then we can always remove it. Thanks for the test script!

zricethezav:
> This looks solid over all, passed all my base64-encoded test cases.
> Perhaps it would be useful to attach metadata (i.e., this was base64-decoded n times).
> Edit: would a tag suffice?

I agree with this. I think adding to the tag would be sufficient until more robust metadata reporting is implemented.

zricethezav:
@bplaxco Amazing work! This has been a frequently requested feature by the community. Fantastic job delivering!

Would you mind adding a section describing this feature in https://github.com/gitleaks/gitleaks?tab=readme-ov-file#additional-configuration? I think it's worth a call-out. Feel free to link to this PR.

detect/decoder.go
```diff
@@ -0,0 +1,256 @@
...
+	"bytes"
+	"encoding/base64"
+	"regexp"
+	"unicode"
+
+	"github.com/rs/zerolog/log"
+)
+
+var b64LikelyChars [128]byte
+var b64Regexp = regexp.MustCompile(`[\w\/\-\+]{16,}={0,3}`)
```
zricethezav, line 13: **Question**: Where did 16 come from?

rgmz: It's an arbitrary minimum length to filter out low-value secrets, similar to entropy. 16 because base64 segments are multiples of 4.

zricethezav:
@bplaxco 
> Decode b64 encoded values (both standard and b64url and even values encoded multiple times)

I wasn't able to find a test that demonstrates this. Would you mind adding one that demonstrates nested decoding?

detect/decoder.go
bplaxco: And also the average password length is around 9-11 characters based on a google search, after b64 encoding things in that range you end up with about 16 as a decent min.

bplaxco:
> @bplaxco Amazing work! This has been a frequently requested feature by the community. Fantastic job delivering!

Thank you! ^_^

> Would you mind adding a section describing this feature in https://github.com/gitleaks/gitleaks?tab=readme-ov-file#additional-configuration? I think it's worth a call-out. Feel free to link to this PR.

For sure! I can get that added.

detect/decoder.go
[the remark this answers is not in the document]
bplaxco: Sounds good. Heh, just a habit to escape most special characters because I'm usually copy pasting stuff all over the place.

bplaxco:
> I agree with this. I think adding to the tag would be sufficient until more robust metadata reporting is implemented. If specifying n times requires major surgery I think we can get away with doing something just like:

@zricethezav Specifying the depth wouldn't be too hard since we already track it.

One thing I've been doing for meta data tags over in fake-leaks is having `<field>:<value>` so that any tag with a `:` in it means a special field was used.

Thoughts on something like this?

```
type:decoded-base64, decode-depth:5
```

Or

```
decoded:base64, decode-depth:5
```

I kinda like `type` tags since they can be used in other places as a generic way to add class info to things.

Aside, if you ever want to sync up on meta data tagging ideas, let me know.

rgmz:
> Aside, if you ever want to sync up on meta data tagging ideas, let me know.

I use a similar convention to add more detailed metadata (`type:file`, `platform:jvm`, `language:java`, `tool:maven`, `framework:spring`, etc.).

bplaxco:
> > This looks solid over all, passed all my base64-encoded test cases.
> > Perhaps it would be useful to attach metadata (i.e., this was base64-decoded n times).
> > Edit: would a tag suffice?
> 
> I agree with this. I think adding to the tag would be sufficient until more robust metadata reporting is implemented. If specifying `n times` requires major surgery I think we can get away with doing something just like:

@zricethezav how does this look and are you okay with the tags used here?

https://github.com/gitleaks/gitleaks/pull/1488/commits/8ebcc9170dfaf75786a42c7100b210fa6cbe0cc6

bplaxco:
> @bplaxco Amazing work! This has been a frequently requested feature by the community. Fantastic job delivering!
> 
> Would you mind adding a section describing this feature in https://github.com/gitleaks/gitleaks?tab=readme-ov-file#additional-configuration? I think it's worth a call-out. Feel free to link to this PR.

Done ^_^

This look good?

https://github.com/gitleaks/gitleaks/blob/dbb0b51dddb23bd8fce21e5175078de9a79ab122/README.md

bplaxco:
@rgmz && @zricethezav I believe that addresses all the feedback. If I missed anything, please let me know!

README.md
```diff
@@ -360,7 +361,35 @@ class CustomClass:
 
 You can ignore specific findings by creating a `.gitleaksignore` file at the root of your repo. In release v8.10.0 Gitleaks added a `Fingerprint` value to the Gitleaks report. Each leak, or finding, has a Fingerprint that uniquely identifies a secret. Add this fingerprint to the `.gitleaksignore` file to ignore that specific secret. See Gitleaks' [.gitleaksignore](https://github.com/gitleaks/gitleaks/blob/master/.gitleaksignore) for an example. Note: this feature is experimental and is subject to change in the future.
 
+#### Decoding
```
zricethezav, line 364: love it

zricethezav (approved):
let'errip, this looks good to me @bplaxco. Thanks for the help reviewing @rgmz!

PR #1491, opened 2024-08-29
Restrict Etsy keywords

rgmz:
### Description:
Per https://github.com/gitleaks/gitleaks/pull/1484#issuecomment-2318158175, it appears that `etsy` is prone to false-positives from strings such as `SetSystem` or `SetSysctl`. This attempts to resolve that issue by only matching `ETSY`, `etsy`, and `Etsy`, and not things like `etSy`. I feel this is a reasonable limitation.

Another solution would be adding a rule allowlist for `GetSys` and `SetSys` on the line.

### Checklist:


zricethezav:
> This attempts to resolve that issue by only matching ETSY, etsy, and Etsy, and not things like etSy

clever!

Issue #1149, opened 2023-04-14
Improve `sumologic-access-id` pattern


rgmz:
**Is your feature request related to a problem? Please describe.**

Despite the `sumologic-access-id` pattern being quite large and complex-looking, it is prone to false-positives.

https://github.com/gitleaks/gitleaks/blob/839f114f2b2ca9429300e545aa5c93a2caf2e3d8/config/gitleaks.toml#L2643-L2650

For example, this regex will match the following line:
```
sumOfExposures = sumOfExposures;
```

I'd be happy to play around with the regex to tweak performance, however, I don't currently use Sumo Logic nor have access to any IDs/keys that I could use for reference.

cc @zricethezav

PR #1492, opened 2024-08-30
Experiment: add capture group for identifier

rgmz:
### Description:

This is a quickly thrown together experiment for an "identifier" capture group. The motivation is that often false positives are determined by their context and not the secret itself, such as #1491 or #1149. I don't intend to merge this and a merely interested in sharing.

**Likes**

- Full context of the identifier allows you to be more precise with per-rule allowlist.
- Makes it possible to have allowlist for identifier and secret separately

**Dislikes**

- Quite clunky and adds another layer of complexity
- Interferes with existing, straightforward `SecretGroup` logic

---

I think the core of this could be implemented with a `match` allowlist, provided that `[\w.-]{0,10}?` was added to capture context before the identifier.

https://github.com/gitleaks/gitleaks/blob/4e43d1109303568509596ef5ef576fbdc0509891/cmd/generate/config/utils/generate.go#L20

### Checklist:

### DIFF
--- a/cmd/generate/config/rules/etsy.go
+++ b/cmd/generate/config/rules/etsy.go
@@ -9,12 +9,15 @@ import (
 func EtsyAccessToken() *config.Rule {
 	// define rule
 	r := config.Rule{
-		RuleID:      "etsy-access-token",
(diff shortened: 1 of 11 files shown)


rgmz:
Superseded by #1585.

PR #1495, opened 2024-09-02
Add Onfido API Token detector to recognize this type of secrets.

lucasan1:
Add [Onfido](https://onfido.com/) API Token detector to recognize this type of secrets. Documentation on API token can be found here: https://documentation.onfido.com/api/latest/#api-tokens.

Note that Onfido is a GitHub [secret scanning partner](https://docs.github.com/en/code-security/secret-scanning/introduction/supported-secret-scanning-patterns) and those tokens are currently detected by the built-in GitHub scanner.

### DIFF
--- a/cmd/generate/config/rules/onfido.go
+++ b/cmd/generate/config/rules/onfido.go
@@ -1,0 +1,26 @@
+package rules
+
+import (
+    "github.com/zricethezav/gitleaks/v8/cmd/generate/config/utils"
+    "github.com/zricethezav/gitleaks/v8/config"
+)
+
+func OnfidoAPIToken() *config.Rule {
+    // define rule
+    r := config.Rule{
+        RuleID:      "onfido-live-api-token",
+        Description: "Found an Onfido live API Token posing a risk of unauthorized access to identity verification services data, potentially leading to misuse of PII.",
+        Regex:       utils.GenerateUniqueTokenRegex(`(?:api_live(?:_[a-zA-Z]{2})?\.[a-zA-Z0-9-_]{11}\.[-_a-zA-Z0-9]{32})`, true),
+
+        Keywords: []string{
+            "onfido",
+            "api_live",
+            "api_live_ca",
+            "api_live_us",
+        },
+    }
+
+// validate
+    tps := []string{utils.GenerateSampleSecret("onfido-live-api-token", "api_live.abc123ABC-_.abc123ABC-_abc123ABC-_abc123ABC-")}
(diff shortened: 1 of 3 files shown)


cmd/generate/config/rules/onfido.go
```diff
@@ -0,0 +1,26 @@
...
+    "github.com/zricethezav/gitleaks/v8/cmd/generate/config/utils"
+    "github.com/zricethezav/gitleaks/v8/config"
+)
+
+func OnfidoAPIToken() *config.Rule {
+    // define rule
+    r := config.Rule{
+        RuleID:      "onfido-live-api-token",
+        Description: "Found an Onfido live API Token posing a risk of unauthorized access to identity verification services data, potentially leading to misuse of PII.",
+        Regex:       utils.GenerateUniqueTokenRegex(`(?:api_live(?:_[a-zA-Z]{2})?\.[a-zA-Z0-9-_]{11}\.[-_a-zA-Z0-9]{32})`, true),
```
rgmz, line 13: A few thoughts on this:

1. The wrapping `(?: ... )` aren't necessary 
2. This could be simplified using `\w`. 
3. I believe you'd also want to set [`IsCaseInsensitive`](https://github.com/gitleaks/gitleaks/blob/4e43d1109303568509596ef5ef576fbdc0509891/cmd/generate/config/utils/generate.go#L58) to false so it doesn't match `API_LIVE`.
```suggestion
        Regex:       utils.GenerateUniqueTokenRegex(`api_live(?:_[a-zA-Z]{2})?\.[\w-]{11}\.[\w-]{32}`, false),
```

cmd/generate/config/rules/onfido.go
```diff
@@ -0,0 +1,26 @@
...
+
+        Keywords: []string{
+            "onfido",
```
rgmz, line 16: I don't think this is necessary as the token will only match `api_live`.

rgmz (commented):
Cool contribution! I have a few pieces of feedback for your consideration.

cmd/generate/config/rules/onfido.go
lucasan1: Hello Richard, thank you for the review!
I wasn't sure if i needed a capture group or not so i opted for the latter; i agree for all your suggestions

cmd/generate/config/rules/onfido.go
lucasan1: Done, removed the "onfido" keyword 👍

PR #1496, opened 2024-09-07
Define multiple allowlists per rule

rgmz:
### Description:
This is an experimental implementation of two things I described [here](https://github.com/gitleaks/gitleaks/issues/1274#issuecomment-1806391187):

> * **Support Multiple Allowlist Targets**
>   Right now you can only chose a single `regexTarget` per allowlist (global or rule). This makes it awkward when you want to ignore multiple patterns, where one is the secret (e.g., `{{ secrets.MY_SECRET}}`) and another is the match or line (e.g., `passwordFile: ...` or `token_id: ...`). The ability to definite separate regexes for `secret` and `match` or `line` would be a nice quality-of-life improvement.
> 
> * **Support Allowlist matching `path && regex`**
>   Presently, `rule.path` + `rule.regex` and `rule.allowlist.paths` + `rule.allowlist.regexes` behave differently. Unlike `rule` which matches `path && regex`, the `allowlist` will match `path || regex` which can lead to unexpected results.

With this change, a rule can have broad regexes AND ones specific to a certain path. This is a bit of a pain point for more complex configurations, at least in my experience.

A contrived example for the sake of demonstration:
```toml
[[rules]]
id = "sumologic-access-id"
description = "Discovered a SumoLogic Access ID, potentially compromising log management services and data analytics integrity."
regex = '''(?i:(?:sumo)(?:[0-9a-z\-_\t .]{0,20})(?:[\s|']|[\s|"]){0,3})(?:=|>|:{1,3}=|\|\|:|<=|=>|:|\?=)(?:'|\"|\s|=|\x60){0,5}(su[a-zA-Z0-9]{12})(?:['|\"|\n|\r|\s|\x60|;]|$)'''
entropy = 3
keywords = [
    "sumo",
]
# Allow `sumOf` anywhere
[[rules.allowlists]]
regexTarget = "line"
regexes = [
    "sumOf",
]
# Allow `abcdefg` when found under `node_modules/@sumologic/opentelemetry-node`
[[rules.allowlists]]
condition = "AND" # default condition is "OR".
regexes = [ 
 '''abcdefg''',
]
paths = [
    "node_modules/@sumologic/opentelemetry-node/.+$",
]
```

This closes https://github.com/gitleaks/gitleaks/issues/1512.

### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -252,19 +252,16 @@ description = "awesome rule 1"
 # does not support lookaheads.
 regex = '''one-go-style-regex-for-this-rule'''
 
-# Golang regular expression used to match paths. This can be used as a standalone rule or it can be used
-# in conjunction with a valid `regex` entry.
-path = '''a-file-path-regex'''
-
-# Array of strings used for metadata and reporting purposes.
-tags = ["tag","another tag"]
-
 # Int used to extract secret from regex match and used as the group that will have
 # its entropy checked if `entropy` is set.
 secretGroup = 3
 
 # Float representing the minimum shannon entropy a regex group must have to be considered a secret.
 entropy = 3.5
+
+# Golang regular expression used to match paths. This can be used as a standalone rule or it can be used
+# in conjunction with a valid `regex` entry.
+path = '''a-file-path-regex'''
 
 # Keywords are used for pre-regex check filtering. Rules that contain
 # keywords will perform a quick string compare check to make sure the
@@ -277,29 +274,40 @@ keywords = [
   "token",
 ]
 
-# You can include an allowlist table for a single rule to reduce false positives or ignore commits
-# with known/rotated secrets
-[rules.allowlist]
-description = "ignore commit A"
-commits = [ "commit-A", "commit-B"]
-paths = [
-  '''go\.mod''',
-  '''go\.sum'''
-]
-# note: (rule) regexTarget defaults to check the _Secret_ in the finding.
-# if regexTarget is not specified then _Secret_ will be used.
-# Acceptable values for regexTarget are "match" and "line"
-regexTarget = "match"
-regexes = [
-  '''process''',
-  '''getenv''',
-]
-# note: stopwords targets the extracted secret, not the entire regex match
-# like 'regexes' does. (stopwords introduced in 8.8.0)
-stopwords = [
-  '''client''',
-  '''endpoint''',
-]
+# Array of strings used for metadata and reporting purposes.
+tags = ["tag","another tag"]
+
+    # ⚠️ In v8.21.0 `[rules.allowlist]` was replaced with `[[rules.allowlists]]`.
+    # This change was backwards-compatible: instances of `[rules.allowlist]` still  work.  
+    #
+    # You can define multiple allowlists for a rule to reduce false positives.
+    # A finding will be ignored if _ANY_ `[[rules.allowlists]]` matches.
+    [[rules.allowlists]]
+    description = "ignore commit A"
+    # When multiple criteria are defined the default condition is "OR".
+    # e.g., this can match on |commits| OR |paths| OR |stopwords|.
+    condition = "OR"
+    commits = [ "commit-A", "commit-B"]
+    paths = [
+      '''go\.mod''',
+      '''go\.sum'''
+    ]
+    # note: stopwords targets the extracted secret, not the entire regex match
+    # like 'regexes' does. (stopwords introduced in 8.8.0)
+    stopwords = [
+      '''client''',
+      '''endpoint''',
+    ]
+
+    [[rules.allowlists]]
+    # The "AND" condition can be used to make sure all criteria match.
+    # e.g., this matches if |regexes| AND |paths| are satisified.
+    condition = "AND"
+    # note: |regexes| defaults to check the _Secret_ in the finding.
+    # Acceptable values for |regexTarget| are "secret" (default), "match", and "line".
+    regexTarget = "match"
+    regexes = [ '''(?i)parseur[il]''' ]
+    paths = [ '''package-lock\.json''' ]
 
 # You can extend a particular rule from the default config. e.g., gitlab-pat
 # if you have defined a custom token prefix on your GitLab instance
@@ -307,11 +315,9 @@ stopwords = [
 id = "gitlab-pat"
 # all the other attributes from the default rule are inherited
 
-[rules.allowlist]
-regexTarget = "line"
-regexes = [
-    '''MY-glpat-''',
-]
+    [[rules.allowlists]]
+    regexTarget = "line"
+    regexes = [ '''MY-glpat-''' ]
 
 # This is a global allowlist which has a higher order of precedence than rule-specific allowlists.
 # If a commit listed in the `commits` field below is encountered then that commit will be skipped and no
@@ -328,7 +334,6 @@ paths = [
 # if regexTarget is not specified then _Secret_ will be used.
 # Acceptable values for regexTarget are "match" and "line"
 regexTarget = "match"
-
 regexes = [
   '''219-09-9999''',
   '''078-05-1120''',
--- a/config/allowlist.go
+++ b/config/allowlist.go
@@ -1,8 +1,16 @@
 package config
 
 import (
+	"fmt"
 	"regexp"
 	"strings"
+)
+
+type AllowlistMatchCondition string
+
+const (
+	AllowlistMatchOr  AllowlistMatchCondition = "OR"
+	AllowlistMatchAnd                         = "AND"
 )
 
 // Allowlist allows a rule to be ignored for specific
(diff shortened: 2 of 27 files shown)


config/allowlist.go
```diff
@@ -5,12 +5,28 @@ import (
 	"strings"
 )
 
+type AllowlistMatchCondition string
+
+const (
+	AllowlistMatchOr  AllowlistMatchCondition = "OR"
+	AllowlistMatchAnd                         = "AND"
```
rgmz, line 12: I opted to use AND/OR instead of &&/|| for clarity. Another option is ALL/ANY.

config/config.go
```diff
@@ -81,60 +88,93 @@ func (vc *ViperConfig) Translate() (Config, error) {
...
+			if a.RegexTarget != "" {
+				switch a.RegexTarget {
+				case "secret":
+					a.RegexTarget = ""
+				case "match":
+					// do nothing
+				case "line":
+					// do nothing
+				default:
+					return Config{}, fmt.Errorf("%s: unknown allowlist |regexTarget| '%s' (expected 'match', 'line')", rule.RuleID, a.RegexTarget)
```
rgmz, line 155: Quality-of-life improvement, since I'm already touching the allowlist logic.

detect/detect_test.go
```diff
@@ -855,6 +848,225 @@ func TestDetectWithSymlinks(t *testing.T) {
...
+			fragment: Fragment{
+				FilePath: "config.js",
+			},
+			allowlist: config.Allowlist{
+				MatchCondition: config.AllowlistMatchAnd,
+				Commits:        []string{"41edf1f7f612199f401ccfc3144c2ebd0d7aeb48"},
+				Paths:          []*regexp.Regexp{regexp.MustCompile(`config.js`)},
+				Regexes:        []*regexp.Regexp{regexp.MustCompile(`(?i)summer.+`)},
+				StopWords:      []string{"2024"},
+			},
```
rgmz, line 988: If this is a non-git scan, should this match on paths+regexes+stopwords? Or should allowlists that include `commits` it ONLY match git scans?

config/config.go
```diff
@@ -81,60 +88,93 @@ func (vc *ViperConfig) Translate() (Config, error) {
...
-				RegexTarget: r.Allowlist.RegexTarget,
-				Regexes:     allowlistRegexes,
-				Paths:       allowlistPaths,
-				Commits:     r.Allowlist.Commits,
-				StopWords:   r.Allowlist.StopWords,
-			},
+			Keywords:    vr.Keywords,
+			Tags:        vr.Tags,
 		}
+		// Parse the allowlist, including the
```
zricethezav, line 124: nit: incomplete comment

config/config.go
```diff
@@ -269,11 +309,9 @@ func (c *Config) extend(extensionConfig Config) {
 			}
 			baseRule.Tags = append(baseRule.Tags, currentRule.Tags...)
 			baseRule.Keywords = append(baseRule.Keywords, currentRule.Keywords...)
-			baseRule.Allowlist.Commits = append(baseRule.Allowlist.Commits, currentRule.Allowlist.Commits...)
```
zricethezav, line 272: Why is this block getting deleted? The extend feature for a singular rule allowlist should function the same as before

detect/detect_test.go
zricethezav: stopwords have always been independent of `regex` and `path`. I think stopwords should continue to have the highest authority on ignoring a finding. 

> Or should AND allowlists that include commits it ONLY match git scans?

I would say only match git scans since the default config does not include any commit fields in rule allowlists. If youre going through the trouble of configuring a rule you could just copy and paste your existing rule w/ and w/o a commit field in the allowlist. I hope that makes sense.

config/config.go
rgmz: I can't recall what my original comment was. I've filled it in based on what makes sense — this must be how ChatGPT feels.

bplaxco:
At first blush it wasn't 100% clear to me that 

```
[[rules.allowlists]]
condition = "AND" # default condition is "OR".
```

applied to the different checks in this allowlist and not across the different allowlists associated with the rule. I don't know that I have a great suggestion at this moment. It wasn't hard to figure out after looking closer, but just trying to think of something that someone could catch at a glance if they weren't familiar with the behavior.

Maybe: `allowRequires = "ALL/ANY"` or something that highlights the checks are what the condition applies to.

Thoughts?

config/config.go
rgmz: I believe this preserves the prior functionality, while allowing the new stuff to function as expected. (Correct me if I'm wrong.)

Consider the following config:
```toml
[extend]
useDefault = true

[[rules]]
id = "generic-api-key"
[rules.allowlist]
paths = ['''Some_Files\.xaml''']
stopwords = ["foo"]
```

With the append logic this would get translated as the following (pseudo-code):
```
allowlist{
  paths = ["some_files.xaml"]
  stopwords = ["foo", DefaultStopWords... ]
}
```

With the new logic, it would get translated as the following. Despite the difference in structure, the logic is still the same: `paths == "some_files.xaml" || stopwords == "foo" || stopwords == DefaultStopWords...`. 
```
allowlist{ // extended
  stopwords = [ DefaultStopWords... ]
},
allowlist{ // our config
  paths = ["some_files.xaml"]
  stopwords = ["foo" ]
}
```
---

As for the "new stuff", if either your or the extended rule's has an allowlist with `AND` then appending commits/paths/regexes/stopwords will break that.

It would be possible to carve out special handling for `[rules.allowlist]` vs. `[[rules.allowlist]]`, or append for `OR` and not `AND`, but I figured this way is the most straightforward. I can add more tests to be certain.

README.md
```diff
@@ -277,41 +274,47 @@ keywords = [
...
-stopwords = [
-  '''client''',
-  '''endpoint''',
-]
+# Array of strings used for metadata and reporting purposes.
+tags = ["tag","another tag"]
+
+    # You can include allowlists for a rule to reduce false positives.
+    # A finding will be ignored if any allowlist matches.
+    [[rules.allowlists]]
```
zricethezav, line 282: @rgmz might be worthwhile to note here with a ⚠️ that this config update was introduced in v8.21.0. Also a note that you can still use `[[rules.allowlist]]`. WDYT?

config/config.go
zricethezav: > With the new logic, it would get translated as the following. Despite the difference in structure, the logic is still the same: paths == "some_files.xaml" || stopwords == "foo" || stopwords == DefaultStopWords....

I gotcha, that makes sense. Boy this would be much easier to test if I made that `config` command already

zricethezav:
> At first blush it wasn't 100% clear to me that
> [[rules.allowlists]]
condition = "AND" # default condition is "OR".
> applied to the different checks in this allowlist and not across the different allowlists associated with the rule. I don't know that I have a great suggestion at this moment. It wasn't hard to figure out after looking closer, but just trying to think of something that someone could catch at a glance if they weren't familiar with the behavior.

We can make it extra clear in the readme that the `condition` field is for checks in the singular allowlist.

README.md
rgmz: I added a comment. LMK your thoughts.

rgmz:
> We can make it extra clear in the readme that the `condition` field is for checks in the singular allowlist.

Do you have anything in mind for wording?

rgmz:
I discovered an 'interesting' interaction while testing. While it's outside the scope of this PR, I figured it was worth mentioning. 

[Generic matches are ignored by the `filter` method if they were found by a specific rule.](https://github.com/gitleaks/gitleaks/blob/8ea608595ce1c992166d710b217f59965d7a44d7/detect/utils.go#L77):
```
11:33AM TRC skipping generic-api-key finding (token="REDACTED"), github-pat rule takes precedence (REDACTED)
11:33AM TRC skipping generic-api-key finding (token = 'REDACTED'), github-pat rule takes precedence (REDACTED)
...
11:33AM INF scan completed in 3.46ms
11:33AM WRN leaks found: 2
```

However, if you create an allowlist entry for the rule, the secret will now be reported as a `generic-api-key` match. It feels suboptimal to have to create two allowlist entries for any finding you want to ignore.
```
11:10AM TRC Skipping finding due to rule allowlist condition=OR contains-stopword=true regex-allowed=false rule-id=github-pat
11:10AM TRC skipping generic-api-key finding (token = 'REDACTED'), github-pat rule takes precedence (REDACTED)
Finding:     gh_token = 'ghp_...
Secret:      ghp_...
RuleID:      github-pat
Entropy:     4.921928
File:        ../test-secrets/test/github_ignore.txt
Line:        2
Fingerprint: ../test-secrets/test/github_ignore.txt:github-pat:2

Finding:     gh_token="ghp_..."
Secret:      ghp_...
RuleID:      generic-api-key
Entropy:     4.921928
File:        ../test-secrets/test/github_ignore.txt
Line:        1
Fingerprint: ../test-secrets/test/github_ignore.txt:generic-api-key:1

11:10AM INF scan completed in 4.56ms
11:10AM WRN leaks found: 2
```

zricethezav:
> However, if you create an allowlist entry for the rule, the secret will now be reported as a generic-api-key match. It feels suboptimal to potentially have to create two allowlist entries for rule-specific finding you want to ignore.

I agree with that. Let's make a follow up issue to track this

README.md
zricethezav: @rgmz I would also add a note that `[rules.allowlist]` still works

rgmz: Done.

zricethezav:
🥳 nice

PR #1500, opened 2024-09-09
Catch common pattern errors with `\b`

rgmz:
### Description:
The intention of this change is to detect common pattern errors (specifically when using `\b`).

I've discussed this in a few different places, most recently: https://github.com/leaktk/patterns/pull/41#discussion_r1723758978 

### Checklist:


rgmz:
This PR isn't ready to merge but does need some feedback.

zricethezav:
> Should this fail or simply log a warning?

I think this is safe to fail since it only gets called during the generation of the config and not during scans.

> Should something permission like (?:.|\s){0,200} count as a non-word character? I am leaning towards yes because it's a lack of specificity, but am stuck.

That's a tough one, I'm not sure either.

Overall I like this change since it enhances already existing guardrails during the config generation.

bplaxco:
> @bplaxco Some interesting results after testing against [leaktk/patterns](https://github.com/leaktk/patterns/blob/73cdd286c0f8f318904fdf9875e6213a5eab8e80/target/patterns/gitleaks/8.18.2):

Thanks for pointing those out :ok_hand: I'll update the patterns and my pattern compiler to check for that ^_^

PR #1501, opened 2024-09-11
Bugfix/1352 incorrect report multiple lines

### LINKED ISSUE #1352: Finding sometimes is incorrectly reported to span multiple lines
**Describe the bug**
Sometimes, the reported start and end lines are different even though the end line has nothing to do with the finding

**To Reproduce**
1. Create a new directory containing only a single file (`.env.prod`) with the following content:
```
DB_HOST=example.com
DB_PORT=443
DB_USERNAME=postgres
DB_PASSWORD=8ae31cacf141669ddfb5da
DB_NAME=best_db
DB_SSL=true
```
2. Run `gitleaks detect --no-git --report-path o`
3. In `o` , you will find the following report:
```
[
 {
  "Description": "Detected a Generic API Key, potentially exposing access to various services and sensitive operations.",
  "StartLine": 4,
  "EndLine": 5,
  "StartColumn": 5,
  "EndColumn": 1,
  "Match": "PASSWORD=8ae31cacf141669ddfb5da",
  "Secret": "8ae31cacf141669ddfb5da",
  "File": ".env.prod",
  "SymlinkFile": "",
  "Commit": "",
  "Entropy": 3.5383105,
  "Author": "",
  "Email": "",
  "Date": "",
  "Message": "",
  "Tags": [],
  "RuleID": "generic-api-key",
  "Fingerprint": ".env.prod:generic-api-key:4"
 }
]
```

**Expected behavior**
The start line is correct. However, the start column already is wrong because start column 5 means that the finding is

```
SSWORD=8ae31cacf141669ddfb5da
DB
```

if columns are 0-indexed and

```
ASSWORD=8ae31cacf141669ddfb5da
D
```

if columns are 1-indexed.

However, what really seems very wrong is the jump into line 5 because `DB_NAME` has nothing to do with the password.

**Screenshots**

 - OS: Manjaro
 - Gitleaks Version: 8.18.2

cc @zricethezav

recreator66:
### Description:
This fixes #1352 
### Checklist:

### DIFF
--- a/detect/detect_test.go
+++ b/detect/detect_test.go
@@ -603,11 +603,32 @@ func TestFromFiles(t *testing.T) {
 			cfgName:          "simple",
 			expectedFindings: []report.Finding{},
 		},
+		{
+			source:  filepath.Join(repoBasePath, "nogit", ".env.prod"),
+			cfgName: "generic",
+			expectedFindings: []report.Finding{
+				{
+					Description: "Generic API Key",
+					StartLine:   4,
+					EndLine:     4,
+					StartColumn: 5,
+					EndColumn:   35,
+					Match:       "PASSWORD=8ae31cacf141669ddfb5da",
+					Secret:      "8ae31cacf141669ddfb5da",
+					Line:        "\nDB_PASSWORD=8ae31cacf141669ddfb5da\nDB_NAME=best_db",
+					File:        "../testdata/repos/nogit/.env.prod",
+					RuleID:      "generic-api-key",
+					Tags:        []string{},
+					Entropy:     3.5383105,
+					Fingerprint: "../testdata/repos/nogit/.env.prod:generic-api-key:4",
+				},
+			},
+		},
 	}
 
 	for _, tt := range tests {
 		viper.AddConfigPath(configPath)
-		viper.SetConfigName("simple")
+		viper.SetConfigName(tt.cfgName)
 		viper.SetConfigType("toml")
 		err := viper.ReadInConfig()
 		require.NoError(t, err)
--- a/detect/detect_test.go
+++ b/detect/detect_test.go
@@ -615,7 +615,7 @@ func TestFromFiles(t *testing.T) {
 					EndColumn:   35,
 					Match:       "PASSWORD=8ae31cacf141669ddfb5da",
 					Secret:      "8ae31cacf141669ddfb5da",
-					Line:        "\nDB_PASSWORD=8ae31cacf141669ddfb5da\nDB_NAME=best_db",
+					Line:        "\nDB_PASSWORD=8ae31cacf141669ddfb5da",
 					File:        "../testdata/repos/nogit/.env.prod",
 					RuleID:      "generic-api-key",
 					Tags:        []string{},
--- a/detect/location.go
+++ b/detect/location.go
@@ -1,7 +1,5 @@
 package detect
 
-import "fmt"
-
 // Location represents a location in a file
 type Location struct {
 	startLine      int
--- a/detect/location.go
+++ b/detect/location.go
@@ -78,6 +78,5 @@ func location(fragment Fragment, matchIndex []int) Location {
 		}
 		location.endLineIndex = end + i
 	}
-	fmt.Println(location.startLineIndex, location.endLineIndex)
 	return location
 }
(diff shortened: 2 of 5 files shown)


detect/location.go
```diff
@@ -50,6 +50,15 @@ func location(fragment Fragment, matchIndex []int) Location {
...
 			location.endLineIndex = newLineByteIndex
 		}
+
+		// Fixes: https://github.com/gitleaks/gitleaks/issues/1352
+		// fixes the bug that a new line from default secretSuffix
+		// leads in the reporting to a false positive in the following line
+		if location.endColumn == 1 {
+			location.endLine -= 1 // -1 because of newline in secretSuffix
+			location.endColumn = prevNewLine - location.startLineIndex
+		}
```
rgmz, line 60: Is it possible to add a test for this? Even with the comment, the reason why this is necessary isn't immediately obvious.

recreator66: I will try to create a test for this during this week.

recreator66: I have added a test case for this issue. In addition it was necessary to adjust the generic.toml within the testdata directory and to change the viper config in the test function.

@rgmz please have a look at the latest commits and let me know if something needs to be adjusted :)

recreator66:
I have added a test case for this issue. In addition it was necessary to adjust the generic.toml within the testdata directory and to change the viper config in the test function.

@rgmz please have a look at the latest commits and let me know if something needs to be adjusted :)

detect/detect_test.go
```diff
@@ -603,11 +603,32 @@ func TestFromFiles(t *testing.T) {
...
+			expectedFindings: []report.Finding{
+				{
+					Description: "Generic API Key",
+					StartLine:   4,
+					EndLine:     4,
+					StartColumn: 5,
+					EndColumn:   35,
+					Match:       "PASSWORD=8ae31cacf141669ddfb5da",
+					Secret:      "8ae31cacf141669ddfb5da",
+					Line:        "\nDB_PASSWORD=8ae31cacf141669ddfb5da\nDB_NAME=best_db",
```
rgmz, line 618: I _think_ this exposes a deeper issue. The `EndLine` and `EndColumn` are now correct but the `line` information was previously populated with the incorrect coordinates.

\+ @zricethezav Shouldn't `line` be a single line? Or is this expected behaviour when a match spans multiple lines?

rgmz (commented):
Great addition! I'd like to get @zricethezav's thoughts on my question, as there may be something else that needs to be changed,

rgmz:
I think the root cause is that newlines are being trimmed here:
https://github.com/gitleaks/gitleaks/blob/ed19c4e8b417313e7ab6878327167c558f5f98f8/detect/detect.go#L266

But the original match indices — including newlines — are being passed to `location`:
https://github.com/gitleaks/gitleaks/blob/ed19c4e8b417313e7ab6878327167c558f5f98f8/detect/location.go#L21-L22

recreator66:
@rgmz I think the root cause starts even earlier..

https://github.com/gitleaks/gitleaks/blob/ed19c4e8b417313e7ab6878327167c558f5f98f8/detect/detect.go#L263

matchIndices already returns the start of the additional line (in our case `DB_NAME=best_db`). This matches also with the result of regex101

regex101
matchIndices

But the end of the secret is actually matchIndex[1]-1 = 87

![matchIndex[1]](https://github.com/user-attachments/assets/5527d00f-3062-4ff2-9e1e-bdd2a2b74847)

An alternative solution would be to review the regex rules and to avoid `\n` and replace them with  hex char`\x20`. However this would only represent a single whitespace and might have a huge impact on the secret detection..

What's your guys opinion on how to proceed? 😄

recreator66:
Since the `secret` already trims `\n`we could easily use it's length to adjust the `matchIndex[1]`. I think it is not necessary to trim whitespaces at the beginning since the regex rules most likely search for keywords at first.

https://github.com/gitleaks/gitleaks/blob/ed19c4e8b417313e7ab6878327167c558f5f98f8/detect/detect.go#L266

I will do a quick commit and see if it works as expected 👍

recreator66:
I have adjusted matchIndex[1] and changed the expected line in the newly created test case accordingly 👍

detect/detect_test.go
zricethezav: > + @zricethezav Also shouldn't line be a single line? Or is this expected behaviour when a match spans multiple lines?

Line should be a single line, correct

zricethezav:
Amazing work @recreator66 and @rgmz! wrangling the location code is not an easy task. LGTM, merging

Issue #1507, opened 2024-09-16
Issue: Failed to Load Config: RuleID Missing (Starting Sept 15th, 2024)


Raag007:
Hi team - starting from 15th Sept 2024 we are seeing the below exception and couldn't run GitLeaks on GitHub repos. 

Can you please look into it, and fix it?

```
Starting: Detect Secrets
==============================================================================
Task         : Gitleaks scan
Description  : Scan git repos (or files) for secrets using regex and entropy.
Version      : 2.9.0
Author       : Joost Voskuil (joost@foxhole.nl)
Help         : [More information](https://github.com/JoostVoskuil/azure-devops-gitleaks)
==============================================================================
Thanks to Zachary Rice (https://github.com/zricethezav) for creating and maintaining gitleaks.
Thanks to Jesse Houwing (https://github.com/jessehouwing) for providing a gitleaks config that has most of Microsoft's deprecated credscan rules ported to it.

Only commits are scanned that are ommited between this build and the previous build with a max of 1000 commits.
Scanning for 1 Git commit(s) for this build. First commit is e4ca616cb93a1446b9a4a3c266015e8f84acb8cc, Last commit is e4ca616cb93a1446b9a4a3c266015e8f84acb8cc
Querying Gitleaks Latest Release GitHub Page.
The latest version of Gitleaks (8.19.1) must be downloaded...
Downloading: https://github.com/zricethezav/gitleaks/releases/download/v8.19.1/gitleaks_8.19.1_windows_x64.zip
Extracting archive
C:\Windows\system32\chcp.com 65001
Active code page: 65001
C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe -NoLogo -Sta -NoProfile -NonInteractive -ExecutionPolicy Unrestricted -Command "$ErrorActionPreference = 'Stop' ; try { Add-Type -AssemblyName System.IO.Compression.FileSystem } catch { } ; [System.IO.Compression.ZipFile]::ExtractToDirectory('C:\AzureDevOpsAgent\_work\_temp\b9765b7c-574e-41c9-a2ca-d0628a043980', 'C:\AzureDevOpsAgent\_work\_temp\21b69363-87d9-4ce0-8fcc-893949b121f7')"
Caching tool: gitleaks 8.19.1 x64
[...]

PR #1509, opened 2024-09-16
Disable rule regex & path validation

rgmz:
### Description:
This disables the check that |regex| and |path| aren't both nil.

As pointed out in https://github.com/gitleaks/gitleaks/issues/1507#issuecomment-2352550208, it does not seem to work with extended configs.

### Checklist:


VertiJoey:
Seems like this will be the fix for my issue aswell

PR #1515, opened 2024-09-21
Verify feature

zricethezav:
### Description:
experimental verify feature

### Checklist:


zricethezav:
copypasta from @rgmz PR


Tentative TODO list:
- [ ] Fix printing for color/no-colour 
- [ ] Put status/reason being printed behind feature flag
- [ ] Put marshaling status/reason behind feature flag
- [ ] Test if verification requests have timeout set (don't want to block forever)
- [ ] Test if verify works with combinations of expect status/body
- [x] Decide whether to keep/remove explicit `requires`
    - [x] Perhaps parse the rule at startup first?
- [ ] Decide whether to keep/change the `report` field
- [ ] General test cases for new feature
- [ ] Filtering for result types: valid, invalid, skipped, error?

--- 
> - [ ] Fix printing for color/no-colour 

I think this is done

> - [ ] Put status/reason being printed behind feature flag

Not mvp, it's an experimental feature after all

> - [ ] Put marshaling status/reason behind feature flag

Need more context on this

> - [ ] Test if verification requests have timeout set (don't want to block forever)

We set a default timeout to 5 seconds. Future feature is to add a configurable rule based and global timeout.

> - [ ] Test if verify works with combinations of expect status/body

Still need todo

> - [ ] Decide whether to keep/change the `report` field

Keep

> - [ ] General test cases for new feature

Mostly covered

> - [ ] Filtering for result types: valid, invalid, skipped, error?

Not mvp

rgmz:
> >         Fix printing for color/no-colour
> 
> I think this is done

[The logic in `printFinding`](https://github.com/gitleaks/gitleaks/pull/1515/files#diff-32ec270572bdfe82f52565e6904b801d93e64ab7e290a95da9206511c6be7c3dR163-R186) was sloppily added. The output probably shouldn't differ except for colour.

>>         Put status/reason being printed behind feature flag
> 
> Not mvp, it's an experimental feature after all

True. So you don't think it's necessary to tie the output to the use of `--experimental-verification`?

>>  Put marshaling status/reason behind feature flag
>
> Need more context on this

Since new fields were added to `report.Finding` [they will show up in the serialized output](https://github.com/gitleaks/gitleaks/pull/1515/files#diff-766604e7fcfaf0f64c0f3d942d636551caea085fe91517662e48e8446ae8d56c). I'm worried it could inadvertently break something -- somehow.

zricethezav:
> True. So you don't think it's necessary to tie the output to the use of --experimental-verification?

@rgmz I misunderstood, I actually think this is necessary. We don't want to break existing reports by adding a new field unless `--experimental-verification` is set. You're right 👍🏻

config/rule.go
```diff
@@ -42,6 +42,42 @@ type Rule struct {
...
+	// This defaults to 'true' but can be set to 'false' for multi-part secret verification.
+	Report bool
+	Verify Verify
+}
+
+type Verify struct {
+	HTTPVerb             string
+	URL                  string
+	Headers              map[string]string
+	ExpectedStatus       []string
```
rgmz, line 56: Should this be `[]int`?

rgmz:
FYI @baruchiro @bplaxco.

config/rule.go
zricethezav: Initially wrote this as a string to support "2xx", "30x" for example. `[]int` might be better considering the return codes should be known.

rgmz: > might be better considering the return codes should be known.

Yeah. I see the appeal of "2xx" but I think it's too error prone in practice.

baruchiro:
Sorry, but I don't work on 2MS anymore, but I will let them know.

Thank you!

cmd/generate/config/rules/config.tmpl
```diff
@@ -63,4 +63,23 @@ regexes = [
...
 ]{{ end }}
 {{ end }}
+{{- if and $rule.Verify.URL $rule.Verify.HTTPVerb }}[rules.Verify]
+url = "{{ $rule.Verify.URL }}"
+httpVerb = "{{ $rule.Verify.HTTPVerb }}"
+{{- with $rule.Verify.ExpectedStatus }}
+expectedStatus = [{{ range $j, $status := . }}
+    "{{ $status }}",{{ end }}
+]{{ end }}
+{{- with $rule.Verify.ExpectedBodyContains }}
```
bplaxco, line 73: It might be overkill but would it make sense to change this to `ExpectedBodyMatches` and have it be a regex? I figure that'll get ahead of some weird more complicated verification situations in the future and make it more flexible if format of responses change a bit.

bplaxco:
Just curious, why was this closed? I was going to pick through it a bit to see if it'll fit a few use cases and add thoughts but also holding off if it's closed ^_^

zricethezav:
Development on this feature has been paused until further notice

PR #1517, opened 2024-09-21
More verify updates

rgmz:
### Description:
This is another set of updates for #1515.

Changes (so far):
- Don't include empty verification-related fields in serialized output
- Validate verify config with `go generate ./...` 

### Checklist:


rgmz:
The latest commit is non-functional so I'll switch this back to a draft.

detect/verify.go
```diff
@@ -200,9 +216,19 @@ FindingLoop:
 						goto RuleLoopEnd
```
zricethezav, line 216: @rgmz I think this could be `continue FindingLoop`?

PR #1518, opened 2024-09-22
More logic updates

rgmz:
### Description:
Apparently I can't re-open #1517. :thinking: 

### Checklist:


config/gitleaks.toml
```diff
@@ -2154,11 +2161,18 @@ keywords = [
...
-entropy = 1
+regex = '''(?:^|[\\'"\x60>=:\s])(hf_[a-zA-Z0-9]{34})(?:$|[\\'"\x60<\s])'''
+entropy = 3
 keywords = [
     "hf_",
 ]
+[rules.Verify]
+httpVerb = "GET"
+url = "https://huggingface.co/api/whoami-v2"
+headers = {headers = "Bearer ${huggingface-access-token}"}
```
zricethezav, line 2172: @rgmz shouldn't this be `headers = { "authorization" :  "Bearer ${huggingface-access-token}"}`

rgmz: Absolutely. I got distracted by a pattern issue and completely missed this.

rgmz: Fixed.

cmd/generate/config/rules/config.tmpl
```diff
@@ -63,23 +63,23 @@ regexes = [
     "{{ $stopword }}",{{ end }}
 ]{{ end }}
 {{ end }}
-{{- if and $rule.Verify.URL $rule.Verify.HTTPVerb }}[rules.Verify]
-url = "{{ $rule.Verify.URL }}"
+{{- if and $rule.Verify }}[rules.Verify]
```
zricethezav, line 66:
```suggestion
{{- if and $rule.Verify }}[rules.verify]
```

Not sure if this is the best practices, but dropping `verify` to lowercase keeps it consistent with `allowlist`. WDYT?

config/flags/flags.go
```diff
@@ -0,0 +1,5 @@
+package flags
+
+import "sync/atomic"
+
+var EnableExperimentalVerification atomic.Bool
```
zricethezav, line 5: **Thought**: I wonder if it makes sense to move more of the cli flags to this package. Making that change would be beyond the scope of this PR but curious to hear your thoughts.

On the other hand `EnableExperimentalVerification` could just be passed as an argument to the `print` functions

config/rule.go
```diff
@@ -46,67 +49,179 @@ type Rule struct {
...
+		return fmt.Errorf("%s: verify config does not contain a |httpVerb|", ruleID)
+	}
+	v.HTTPVerb = strings.ToUpper(v.HTTPVerb)
+	if _, ok := validMethods[v.HTTPVerb]; !ok {
+		return fmt.Errorf("%s: invalid HTTP verb: %s", ruleID, v.HTTPVerb)
+	}
+
+	// Parse URL.
+	if _, err := url.ParseRequestURI(v.URL); err != nil {
+		return fmt.Errorf("%s: invalid URL: %w", ruleID, err)
```
zricethezav, line 146: **Question**: would this fail for urls with expanded strings? e.g, `example.com/${account-id}`?

cmd/generate/config/rules/config.tmpl
rgmz: Consistency is nice; I can see this driving someone crazy.

config/flags/flags.go
rgmz: > **Thought**: I wonder if it makes sense to move more of the cli flags to this package. Making that change would be beyond the scope of this PR but curious to hear your thoughts.

We'd have to review how flags are being set/accessed to see if it would improve the code quality.

There were a few places where it wasn't possible/practical to pass `d.EnableExperimentalVerification` (either ugly code or cyclical references). Most of those cases no longer exists, however, this still seems like the cleanest way.

config/rule.go
rgmz: It won't fail unless the placeholder is in the hostname. I'm fine to remove it if that seems like a problem.

zricethezav: seems like an unlikely case, probably fine for now

PR #1520, opened 2024-09-23
Tweak Kubernetes pattern to handle Helm variables

rgmz:
### Description:
This is a follow-up to #1462.

[A comment left by](https://github.com/gitleaks/gitleaks/pull/1462#issuecomment-2368049635) @brampat made me realize that I hadn't covered all possible syntax for [Helm variables](https://helm.sh/docs/chart_template_guide/variables/).

### Checklist:

### DIFF
--- a/cmd/generate/config/rules/kubernetes.go
+++ b/cmd/generate/config/rules/kubernetes.go
@@ -17,8 +17,7 @@ func KubernetesSecret() *config.Rule {
 	// - valid base64 characters
 	// - longer than 10 characters (no "YmFyCg==")
 	//language=regexp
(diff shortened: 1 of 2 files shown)


zricethezav:
v nice, thank you!

brampat:
When can we expect this fix to be released in the gitleaks version? I'm currently on the latest (8.21.2) and I stall get false positives on the kubernetes Helm templates

rgmz:
This was included in [v8.19.3](https://github.com/gitleaks/gitleaks/releases/tag/v8.19.3). Mind sharing an example of the false positives you're seeing?

brampat:
Okay, I see what went wrong. I exported the first version of the rule as a toml file. Those still triggered, which I didn't realise was causing the issue. I did a full clean checkout which didn't give any false positives, so that's where I was completely baffled. Then I saw the toml file and thought... wait a minute. I renamed them to filename.toml___ to disable them and that confirmed those caused the issue.

PR #1524, opened 2024-09-23
Extend allowlist & handle extend when validating

### LINKED ISSUE #1523: Fix config extend validation + allowlist
> We use a toml file in our repo to _extend_ the default configuration, like mentioned in the PR I believe to cause this issue: https://github.com/gitleaks/gitleaks/pull/1463
> 
> As an example, our "local" config looks like this:
> 
> ```
> [extend]
> # useDefault will extend the base configuration with the default gitleaks config:
> # https://github.com/zricethezav/gitleaks/blob/master/config/gitleaks.toml
> useDefault = true
> 
> # An array of tables that contain information that define instructions
> # on how to detect secrets
> [[rules]]
> id = "generic-api-key"
> [rules.allowlist]
> 	description = "False positive. Keys used for colors match the rule, and should be excluded."
> 	files = ['''Some_Files\.xaml''']
> ```
> 
> In this case, the `generic-api-key` rule incorrectly flagged some of our hex-defined colors in a XAML file. Since the update, we now get this error:
> 
> `FTL Failed to load config error="generic-api-key: both |regex| and |path| are empty, this rule will have no effect"`
> 
> But... the `[id]` isn't empty?
> 

_Originally posted by @celloza in https://github.com/gitleaks/gitleaks/issues/1507#issuecomment-2352550208_

While researching this, I also noticed that [rule allowlists don't get merged with extend](https://github.com/gitleaks/gitleaks/blob/97dbe1ef0969f6162e061a5418063f19ed729497/config/config.go#L237).

rgmz:
### Description:
This fixes #1523.

### Checklist:

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -131,9 +131,6 @@ func (vc *ViperConfig) Translate() (Config, error) {
 				StopWords:   r.Allowlist.StopWords,
 			},
 		}
-		if err := r.Validate(); err != nil {
-			return Config{}, err
-		}
 
 		orderedRules = append(orderedRules, r.RuleID)
 		rulesMap[r.RuleID] = r
@@ -171,7 +168,15 @@ func (vc *ViperConfig) Translate() (Config, error) {
 		} else if c.Extend.Path != "" {
 			c.extendPath()
 		}
+	}
 
+	// Validate the rules after everything has been assembled (including extended configs).
+	if extendDepth == 0 {
+		for _, rule := range rulesMap {
+			if err := rule.Validate(); err != nil {
+				return Config{}, err
+			}
+		}
 	}
 
 	return c, nil
@@ -235,12 +240,22 @@ func (c *Config) extendURL() {
 }
 
 func (c *Config) extend(extensionConfig Config) {
-	for ruleID, rule := range extensionConfig.Rules {
-		if _, ok := c.Rules[ruleID]; !ok {
-			log.Trace().Msgf("adding %s to base config", ruleID)
-			c.Rules[ruleID] = rule
-			c.Keywords = append(c.Keywords, rule.Keywords...)
+	for ruleID, baseRule := range extensionConfig.Rules {
+		currentRule, ok := c.Rules[ruleID]
+		if !ok {
+			// Rule doesn't exist, add it to the config.
+			c.Rules[ruleID] = baseRule
+			c.Keywords = append(c.Keywords, baseRule.Keywords...)
 			c.OrderedRules = append(c.OrderedRules, ruleID)
+		} else {
+			// Rule exists, merge our changes into the base.
+			baseRule.Allowlist.Commits = append(baseRule.Allowlist.Commits, currentRule.Allowlist.Commits...)
+			baseRule.Allowlist.Paths = append(baseRule.Allowlist.Paths, currentRule.Allowlist.Paths...)
+			baseRule.Allowlist.Regexes = append(baseRule.Allowlist.Regexes, currentRule.Allowlist.Regexes...)
+			baseRule.Allowlist.StopWords = append(baseRule.Allowlist.StopWords, currentRule.Allowlist.StopWords...)
+
+			delete(c.Rules, ruleID)
+			c.Rules[ruleID] = baseRule
 		}
 	}
 
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"fmt"
+	"github.com/google/go-cmp/cmp"
 	"regexp"
 	"testing"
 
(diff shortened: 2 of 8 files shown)


rgmz:
@zricethezav There's a strange issue with this one. The tests pass if the others are commented out, but fails if they are uncommented.

Is there some sort of state being kept between classes?

config/config.go
```diff
@@ -235,12 +240,22 @@ func (c *Config) extendURL() {
...
-			c.Keywords = append(c.Keywords, rule.Keywords...)
+	for ruleID, baseRule := range extensionConfig.Rules {
+		currentRule, ok := c.Rules[ruleID]
+		if !ok {
+			// Rule doesn't exist, add it to the config.
+			c.Rules[ruleID] = baseRule
+			c.Keywords = append(c.Keywords, baseRule.Keywords...)
 			c.OrderedRules = append(c.OrderedRules, ruleID)
+		} else {
+			// Rule exists, merge our changes into the base.
```
zricethezav, line 251: nice, ty for this

rgmz: Do you think there are any other fields we should append or overwrite?

bufferoverflow: I would suggest adding `RegexTarget` as well:

```diff
diff --git a/config/config.go b/config/config.go
index caea22f..34e5f98 100644
--- a/config/config.go
+++ b/config/config.go
@@ -252,6 +252,7 @@ func (c *Config) extend(extensionConfig Config) {
                        baseRule.Allowlist.Commits = append(baseRule.Allowlist.Commits, currentRule.Allowlist.Commits...)
                        baseRule.Allowlist.Paths = append(baseRule.Allowlist.Paths, currentRule.Allowlist.Paths...)
                        baseRule.Allowlist.Regexes = append(baseRule.Allowlist.Regexes, currentRule.Allowlist.Regexes...)
+                       baseRule.Allowlist.RegexTarget = currentRule.Allowlist.RegexTarget
                        baseRule.Allowlist.StopWords = append(baseRule.Allowlist.StopWords, currentRule.Allowlist.StopWords...)
 
                        delete(c.Rules, ruleID)
```

I have a case where I must use line and the change above would enable extend for us.

bufferoverflow: I just made https://github.com/gitleaks/gitleaks/pull/1536 to cover that case

9999years:
Hello, this broke the ability to override the entropy of rules.

With `.gitleaks.toml`:

```toml
[extend]
useDefault = true

[[rules]]
id = "finicity-client-secret"
entropy = 4
```

```
$ echo "newtype FinicityDataResponse = FinicityDataResponse" | gitleaks --config=.gitleaks.toml --redact=0 --no-banner --verbose --log-level trace stdin
11:54AM DBG using gitleaks config .gitleaks.toml from `--config`
11:54AM DBG extending config with default config

Finding:     newtype FinicityDataResponse = FinicityDataResponse
Secret:      FinicityDataResponse
RuleID:      finicity-client-secret
Entropy:     3.584184

11:54AM INF scan completed in 1.72ms
11:54AM WRN leaks found: 1
```

config/config.go
9999years: I'm surprised this didn't get caught earlier; I believe this makes it impossible to use _most_ of the rule attributes documented in the README:

```toml
# An array of tables that contain information that define instructions
# on how to detect secrets
[[rules]]

# Unique identifier for this rule
id = "awesome-rule-1"

# Short human readable description of the rule.
description = "awesome rule 1"

# Golang regular expression used to detect secrets. Note Golang's regex engine
# does not support lookaheads.
regex = '''one-go-style-regex-for-this-rule'''

# Golang regular expression used to match paths. This can be used as a standalone rule or it can be used
# in conjunction with a valid `regex` entry.
path = '''a-file-path-regex'''

# Array of strings used for metadata and reporting purposes.
tags = ["tag","another tag"]

# Int used to extract secret from regex match and used as the group that will have
# its entropy checked if `entropy` is set.
secretGroup = 3

# Float representing the minimum shannon entropy a regex group must have to be considered a secret.
entropy = 3.5
```

https://github.com/gitleaks/gitleaks?tab=readme-ov-file#configuration

rgmz:
> Hello, this broke the ability to override the entropy of rules.

Hey @9999years, can you elaborate on what you expect to happen and which version you upgraded from? As far as I'm aware, that would have never worked.

Prior this this PR, `extend` only did did two things:
1. Merged your custom rules with the extended config, provided there were no ID conflicts
2. Appended values to the global allowlist.

https://github.com/gitleaks/gitleaks/blob/v8.19.2/config/config.go#L237

9999years:
@rgmz Just a minimal example. Our actual configuration looks like this:

```toml
[[rules]]
description = "Finicity Client Secret"
id = "finicity-client-secret"
regex = '''(?i)(?:finicity)(?:[0-9a-z\-_\t .]{0,20})(?:[\s|']|[\s|"]){0,3}(?:=|>|:{1,3}=|\|\|:|<=|=>|:|\?=)(?:'|\"|\s|=|\x60){0,5}([a-z0-9]{20})(?:['|\"|\n|\r|\s|\x60|;]|$)'''
secretGroup = 1
entropy = 4
keywords = ["finicity"]
```

PR #1535, opened 2024-09-28
Disable inherited rule from extended config

### LINKED ISSUE #1531: Disable inherited rule from extended config
> Any way to disable an inherited rule from an extended base config would be very helpful 🙂
> 
> It's great to be able to ignore/exclude secret-fingerprints, path-regex, commits and single lines (via comment), but if there is just one rule in the base config which regularly creates false positives you're pretty much forced to stop using the extension mechanism and lose the ability to be kept "up-to-date" automatically 😔
> 

_Originally posted by @daniel-stockhausen in https://github.com/gitleaks/gitleaks/issues/1274#issuecomment-1959612039_

rgmz:
### Description:
This fixes #1531.

It allows you to specify ~`ignoredRules`~ `disabledRules` under `[extend]` and exclude them from your config. (I'm amenable to changing the name.)

Edit: I switched back to `disabled` because I think it's consistent with the `--enable-rule` flag.

### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,6 +235,8 @@ useDefault = true
 # or you can supply a path to a configuration. Path is relative to where gitleaks
 # was invoked, not the location of the base config.
 path = "common_config.toml"
+# If there are any rules you don't want to inherit, they can be specified here.
+disabledRules = [ "generic-api-key"]
 
 # An array of tables that contain information that define instructions
 # on how to detect secrets
--- a/config/config.go
+++ b/config/config.go
@@ -76,9 +76,10 @@ type Config struct {
 // Extend is a struct that allows users to define how they want their
 // configuration extended by other configuration files.
 type Extend struct {
-	Path       string
-	URL        string
-	UseDefault bool
+	Path          string
+	URL           string
+	UseDefault    bool
+	DisabledRules []string
 }
 
 func (vc *ViperConfig) Translate() (Config, error) {
@@ -215,7 +216,7 @@ func (vc *ViperConfig) Translate() (Config, error) {
 
 	// Validate the rules after everything has been assembled (including extended configs).
 	if extendDepth == 0 {
-		for _, rule := range rulesMap {
+		for _, rule := range c.Rules {
 			if err := rule.Validate(); err != nil {
 				return Config{}, err
 			}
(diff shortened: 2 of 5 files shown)


detect/detect.go
```diff
@@ -194,6 +194,9 @@ func (d *Detector) Detect(fragment Fragment) []report.Finding {
 	}
 
 	for _, rule := range d.Config.Rules {
+		if !rule.Enabled {
+			continue
+		}
```
rgmz, line 199: It seemed easiest to simply skip over the rule, rather than trying to remove it from `Rules`, `OrderedRules`, and `Keywords` in Config.

cmd/generate/config/rules/config.tmpl
```diff
@@ -29,6 +29,8 @@ paths = [
 
 {{ range $i, $rule := .Rules }}[[rules]]
 id = "{{$rule.RuleID}}"
+{{- with $rule.Enabled }}
```
rgmz, line 32: This probably doesn't work as intended. I think we'd want it the other way: include if `false`, exclude if `true`.

config/config.go
```diff
@@ -34,6 +34,7 @@ type ViperConfig struct {
 		Keywords    []string
 		Path        string
 		Tags        []string
+		Enabled     *bool
```
rgmz, line 37: I think the best name is either `enabled` or `disabled`. I'm not sure which the most intuitive.

bufferoverflow:
@rgmz regarding your question within https://github.com/gitleaks/gitleaks/pull/1536#issuecomment-2381440822 , I would rather prefer a list options within extend, such as the following:

```toml
[extend]

disabledRules = [
  "private-key",
  "generic-api-key",
  "jwt"
]
```

Typically you have a bunch of rules to exclude and adding each rule with a `enabled = false` could make a config using extend quite huge, e.g. 
```toml
[extend]

[[rules]]
id= "private-key"
enabled = false

[[rules]]
id= "generic-api-key"
enabled = false

[[rules]]
id= "jwt"
enabled = false
```

rgmz:
> Typically you have a bunch of rules to exclude and adding each rule with a enabled = false could make a config using extend quite huge, e.g.

That's a great point. I've updated the implementation accordingly.

README.md
```diff
@@ -237,6 +237,8 @@ useDefault = true
 # or you can supply a path to a configuration. Path is relative to where gitleaks
 # was invoked, not the location of the base config.
 path = "common_config.toml"
+# If there are any rules you don't want to inherit, they can be specified here.
+ignoredRules = [ "awsome-rule-2"]
```
bufferoverflow, line 241:
```suggestion
ignoredRules = [ "generic-api-key"]
```

I suggest adding `generic-api-key` from the default config here, as a lot of folks might want to disable this due to the amount of false-positives.

zricethezav:
great feature. Sorry for the delayed merge

Issue #1384, opened 2024-04-02
Documentation for the [extend] file is wrong


drclark-dev:
The documentation for [extend] config files states:
```
# Extend the base (this) configuration. When you extend a configuration
# the base rules take precedence over the extended rules. I.e., if there are
# duplicate rules in both the base configuration and the extended configuration
# the base rules will override the extended rules.
```

This seems counterintuitive because extensibility by definition adds to or modifies base functionality, and in fact after testing with an extend file that makes private-key just match on the substring `toto`  I see that it does in fact override the base rules.

```
title = "Test extend"

[extend]
useDefault = true

[[rules]]
description = "test"
id = "private-key"
regex = '''toto'''
```

And using the extend file I get a match based on the extend rule, not the base rule.

```
gitleaks detect --no-git -v -c ..\gitleaks-extend.toml

    ○
    │╲
    │ ○
    ○ ░
    ░    gitleaks

Finding:     totoo
Secret:      toto
RuleID:      private-key
Entropy:     1.000000
File:        test.txt
Line:        1
Fingerprint: test.txt:private-key:1

11:02AM INF scan completed in 13.3ms
11:02AM WRN leaks found: 1
```

In my opinion this is the desired functionality for an extend rule, and the documentation should be corrected to reflect it.

shearn89:
I'm currently battling with this but can't work out what's going on. I think the documentation is severely lacking.

My problem:
* I want to configure gitleaks via pre-commit to scan for secrets
* I want to define a list of files that are ignored (e.g. `.terraform.lock.hcl` which triggers for some reason)
* I want to exempt certain files from particular rules (e.g. we store some access key IDs in `constants.tf` and I want to stop getting nagged about it)

This is probably a common use case - "I've found a thing I need to allow list, how do I set that up" - and yet the documentation doesn't seem to provide a clear example.

bufferoverflow:
@zricethezav I suggest closing this as it is addressed by #1536

drclark-dev:
I see that this PR addresses a functional issue raised by @shearn89 but to my original point, the documentation should also be updated.  It says:

```
# Extend the base (this) configuration. When you extend a configuration
# the base rules take precedence over the extended rules. I.e., if there are
# duplicate rules in both the base configuration and the extended configuration
# the base rules will override the extended rules.
```

But the opposite is true. Extended rules take precedence over the base rules.

bufferoverflow:
[...]

Issue #1027, opened 2022-11-10
Additional documentation for "extending a central config"


weineran:
Add more documentation to the README around how to extend a central config. Currently we have to refer users to a blog post, which is not idea.

_Originally posted by @anuj147256 in https://github.com/zricethezav/gitleaks/discussions/1020_

bufferoverflow:
@zricethezav I suggest closing this as it is addressed by #1536

Issue #999, opened 2022-10-05
How to configure gitleaks


weineran:
Opening an issue for greater visibility for @zricethezav's latest blog post on gitleaks configuration: https://blog.gitleaks.io/stop-leaking-secrets-configuration-2-3-aeed293b1fbf

Read the blog post to learn:
* How to specify a custom config file
* How to pull in a central config file that can be used across multiple repos
* How to extend a central config file with a repo-specific config file
* Shameless plug: How to do all of the above using [Gitleaks-Action](https://gitleaks.io/products.html)

bufferoverflow:
@zricethezav I suggest closing this as it is addressed by #1536

PR #1536, opened 2024-09-28
fix: use regexTarget for extend config

bufferoverflow:
### Description:

Extend config is super useful, but if you have `regexTarget = "line"` within `[rules.allowlist]`, extend does not work.
The basic functionality works thanks to https://github.com/gitleaks/gitleaks/pull/1524 but `regexTarget` was missing.
Beside of this the adding keywords from base and extended rule into global keywords list is required to extend rules.
In addition some documentation to extend a rule has been added.

This should also solve:

- https://github.com/gitleaks/gitleaks/issues/999
- https://github.com/gitleaks/gitleaks/issues/1027
- https://github.com/gitleaks/gitleaks/issues/1384

### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -300,6 +300,17 @@ stopwords = [
   '''endpoint''',
 ]
 
+# You can extend a particular rule from the default config. e.g., gitlab-pat
+# if you have defined a custom token prefix on your GitLab instance
+[[rules]]
+id = "gitlab-pat"
+# all the other attributes from the default rule are inherited
+
+[rules.allowlist]
+regexTarget = "line"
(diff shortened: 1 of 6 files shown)


rgmz:
Makes sense to me. :+1:

Do you think it'd be beneficial to have clear options for "update this inherited rule" versus "completely replace this inherited rule"? The logic for `[extend]` isn't fully fleshed out and likely a common source of confusion.

zricethezav:
Thanks for the catch. This LGTM 👍🏻. Great work @bufferoverflow

PR #1538, opened 2024-09-29
Store `Config` keywords as map

rgmz:
### Description:
This uses a `map` to deduplicate `Config.Keywords`.

### Checklist:

### DIFF
--- a/cmd/generate/config/utils/validate.go
+++ b/cmd/generate/config/utils/validate.go
@@ -11,20 +11,9 @@ import (
 	"strings"
 )
 
-func Validate(r config.Rule, truePositives []string, falsePositives []string) *config.Rule {
-	// normalize keywords like in the config package
-	var keywords []string
-	for _, k := range r.Keywords {
-		keywords = append(keywords, strings.ToLower(k))
-	}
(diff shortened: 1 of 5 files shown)


rgmz:
Hmm, lack of ordered output for the generated config could be more effort than it's worth.

zricethezav:
> Hmm, lack of ordered output for the generated config could be more effort than it's worth.
 
 Is this still a concern?

rgmz:
No. The original code only used a map in validate.go which resulted in random keyword order.

cmd/generate/config/utils/validate.go
```diff
@@ -81,5 +60,29 @@ func ValidateWithPaths(r config.Rule, truePositives map[string]string, falsePosi
 				Msg("Failed to Validate. False positive was detected by regex and/or path.")
 		}
 	}
-	return &r
+	return r
+}
+
+func createDetector(r *config.Rule) *detect.Detector {
```
zricethezav, line 66: **Suggestion**: @rgmz wdyt about renaming this `createSingleRuleDetector`? At first when I read this code I was like "wait why are we building a detector here?"

Issue #1333, opened 2024-01-30
gitleaks detect consistently reports wrong commit information for some secrets


nicholas-arnica:
**Describe the bug**
`gitleaks detect` consistently reports the wrong commit information (`Commit`, `Author`, `Email`, `Date`, `Message`) for many findings.

The `StartLine`, `EndLine`, `StartColumn`, `EndColumn`, `Match`, `Secret`, and `File` are all correct.

We've observed this across several unrelated repositories.

(speculation) It looks like there may be a bug with parsing the `git log`s, as the `Message` field looks like it's appending git log information from the CORRECT commit. For example, in the example below, the commit message `gitleaks detect` reports is:

```
Merge branch 'BLAKE3' of github.com:BoloniniD/ClickHouse into BLAKE3

commit 645e156af6b362145fad82d714f8e70a5b5a55a8
Author: Meena Renganathan <Meena.Renganathan@ibm.com>
Date:   Wed Apr 6 14:50:10 2022 -0700

Updated the boringssl-cmake to match the latest broingssl module update
```

but the message for `f6ded7a51cf917bdb44097066fab608c0facde5b` is just
```
Merge branch 'BLAKE3' of github.com:BoloniniD/ClickHouse into BLAKE3
```

but if we `git log 645e156af6b362145fad82d714f8e70a5b5a55a8` (id taken from the "leaked" Message), we have:
```
commit 645e156af6b362145fad82d714f8e70a5b5a55a8
Author: Meena Renganathan <Meena.Renganathan@ibm.com>
Date:   Wed Apr 6 14:50:10 2022 -0700

    Updated the boringssl-cmake to match the latest broingssl module update
```

**To Reproduce**
1) Clone https://github.com/nicholas-gitgoat/ClickHouse
2) Run `gitleaks detect -r gitleaks.json`

Should have output similar to below
```
9:00AM WRN warning: exhaustive rename detection was skipped due to too many files.
9:00AM WRN warning: you may want to set your diff.renameLimit variable to at least 1128 and retry the command.
9:00AM INF 74007 commits scanned.
9:00AM INF scan completed in 2m5s
9:00AM WRN leaks found: 849
```

Edit: Use `gitleaks detect -s /home/nicholas/Projects/ClickHouse/ -r gitleaks.json --log-opts="^645e156af6^ 645e156af6 f6ded7a51cf917bdb44097066fab608c0facde5b"` for a quicker scan that doesn't take 2 minutes, and also reproduces the issue

[...]

PR #1542, opened 2024-09-30
Scan binary files

### LINKED ISSUE #1541: Scan binary files
Gitleaks does not scan binary files, opting to skip them instead. 

https://github.com/gitleaks/gitleaks/blob/c5b15c9d74036ca0c32134280c3d941b41e4f2d9/detect/git.go#L24-L27

Binary files should be scanned. A simple equivalent of `strings $file` would suffice for the initial implementation. The ability to scan more complex file types, such as archives, would be nice to have eventually.

N/A

**Additional context**
N/A

cc @zricethezav

rgmz:
### Description:
This is a rough implementation that fixes #1541.

General feedback on approach and potential optimizations are welcome.

TODO:
- [ ] Check mimetype?
- [ ] Handle bare repositories
- [ ] Decide whether it's better to copy files to tmp -> read mimetype -> read (scan), versus copy directly to a buffer.
- [x] Add handling for non-git
- [x] Add a limit for file size
- [ ] Add separate handling for binary and text files. Binary files cannot be safely split.
- [ ] Measure performance
- [ ] Add test cases

Anything else?

### Checklist:


detect/git.go
```diff
@@ -71,3 +86,94 @@ func (d *Detector) DetectGit(gitCmd *sources.GitCmd) ([]report.Finding, error) {
...
+	reader, err := file.Reader()
+	if err != nil {
+		return fmt.Errorf("failed to open file reader: %w", err)
+	}
+	defer func(reader io.ReadCloser) {
+		_ = reader.Close()
+	}(reader)
+
+	// Scan the reader.
+	findings, err := d.DetectReader(reader, 1*1024*1024)
```
rgmz, line 136: @zricethezav I'm trying to avoid duplicating code. Is there a meaningful difference between how [`DetectReader`](https://github.com/gitleaks/gitleaks/blob/c5b15c9d74036ca0c32134280c3d941b41e4f2d9/detect/reader.go#L11) and [`DetectFiles`](https://github.com/gitleaks/gitleaks/blob/c5b15c9d74036ca0c32134280c3d941b41e4f2d9/detect/directory.go#L42) work?

i.e., would it be reasonable for both `DetectGit` and `DetectFiles` to call `DetectReader` instead of implementing their own logic?

detect/reader.go
```diff
@@ -8,11 +8,11 @@ import (
 )
 
 // DetectReader accepts an io.Reader and a buffer size for the reader in KB
-func (d *Detector) DetectReader(r io.Reader, bufSize int) ([]report.Finding, error) {
-	reader := bufio.NewReader(r)
-	buf := make([]byte, 0, 1000*bufSize)
-	findings := []report.Finding{}
-
+func (d *Detector) DetectReader(r io.Reader) ([]report.Finding, error) {
```
rgmz, line 11: I think it's safe to remove `bufSize` here? It looks like the value was only used in `reader_test`, but removing it has no effect.

rgmz:
Testing is failing due to an abundance of "file not found" errors. This appears to be caused by #1333.
```
6:04PM ERR failed to scan binary file error="error getting file in tree: file not found" commit=bc26fcd27702806e96991764482078b3f0905f44 path=scripts/.scala/project_940fb43dce-2f45572b79/classes/main/inc$minus73.tasty
```

rgmz:
Superseded by #1841 or #1872.

PR #1546, opened 2024-10-04
Update global config 

rgmz:
### Description:
This PR updates the config generation logic to populate the global config programmatically.

The advantages of this are two-fold:
1. We can generate tests for the global config, e.g. making sure that a pattern is working as expected (not done atm)
2. The rule validate tests now inherit global allowlist values. This allows us to safely add common offenders, e.g. `{{ .Secrets.FOO }}` to the global allowlist.

It also organizes the existing `[allowlist.paths]` entries, and adds a few new ones.

### Checklist:

### DIFF
--- a/cmd/generate/config/base/config.go
+++ b/cmd/generate/config/base/config.go
@@ -1,0 +1,69 @@
+package base
+
+import (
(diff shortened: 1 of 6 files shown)


zricethezav:
I like it

PR #1547, opened 2024-10-04
Add placeholder regexes to global allowlist

rgmz:
### Description:
This leverages #1546 to add `[allowlist.regexes]` entries for common placeholder/variable formats, with tests. 

### Checklist:


zricethezav:
Nice, thanks for providing TP/FP tests. That change might introduce some false negatives to slip through the cracks but it's probably worth it to filter out false positives.

rgmz:
My hope is that the patterns are conservative enough that the chance of FNs is ~0%.

PR #1556, opened 2024-10-09
Fix rule extension

9999years:
### Description:
PR #1524 broke the ability to specify a `[[rules]]` with the same ID as a built-in rule in order to extend the default configuration.

This PR restores this ability, but merges the provided configuration with the default rule instead of overwriting the default rule wholesale.

### Checklist:

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -252,6 +252,27 @@ func (c *Config) extend(extensionConfig Config) {
 			c.OrderedRules = append(c.OrderedRules, ruleID)
 		} else {
 			// Rule exists, merge our changes into the base.
+			if currentRule.Description != "" {
+				baseRule.Description = currentRule.Description
+			}
+			if currentRule.Entropy != 0 {
+				baseRule.Entropy = currentRule.Entropy
+			}
+			if currentRule.SecretGroup != 0 {
+				baseRule.SecretGroup = currentRule.SecretGroup
+			}
+			if currentRule.Regex != nil {
+				baseRule.Regex = currentRule.Regex
+			}
+			if currentRule.Path != nil {
+				baseRule.Path = currentRule.Path
+			}
+			if len(currentRule.Tags) > 0 {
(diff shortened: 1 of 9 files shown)


rgmz:
(Continuing the discussion from #1524.)

> @rgmz Just a minimal example. Our actual configuration looks like this:

Ah, I see you want to completely replace the rule. I haven't seen anyone do that before. The impetus for this change was because quite a few people were doing stuff similar to your minimal example.

The `[extend]` functionality seems to be poorly understood / lack definite clarity on what gets appended versus overwritten. This change makes sense — although it is inconsistent that tags/keywords are overwritten but allowlist is appended.

config/config.go
```diff
@@ -252,6 +252,27 @@ func (c *Config) extend(extensionConfig Config) {
...
+				baseRule.SecretGroup = currentRule.SecretGroup
+			}
+			if currentRule.Regex != nil {
+				baseRule.Regex = currentRule.Regex
+			}
+			if currentRule.Path != nil {
+				baseRule.Path = currentRule.Path
+			}
+			if len(currentRule.Tags) > 0 {
+				baseRule.Tags = currentRule.Tags
```
zricethezav, line 271: @9999years Should this append to the base rule rather than overwrite? Same goes with Keywords?

9999years: Sure, that's fine.

zricethezav:
> The [extend] functionality seems to be poorly understood / lack definite clarity on what gets appended versus overwritten. This change makes sense — although it is inconsistent that tags/keywords are overwritten but allowlist is appended.

I think we need to be consistent across the board for the default case. _All the things_ should be appended rather than overwritten. A feature could be included to configure the behavior of `extend` down the line -- `append` vs `overwrite`, or something.

config/config.go
rgmz: Overwrite does give more control. #1384 points out that the docs explicitly say rules are overridden.

PR #1563, opened 2024-10-14
docs: describe extended rules take precedence over base rules

### LINKED ISSUE #1384: Documentation for the [extend] file is wrong

bufferoverflow:
Closes #1384

### Description:

Document actual behavior

### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,9 +223,10 @@ Gitleaks offers a configuration format you can follow to write your own secret d
 title = "Gitleaks title"
 
 # Extend the base (this) configuration. When you extend a configuration
-# the base rules take precedence over the extended rules. I.e., if there are
-# duplicate rules in both the base configuration and the extended configuration
-# the base rules will override the extended rules.
+# the extended rules take precedence over the base rules. I.e., if there are
+# duplicate rules in both the extended configuration and the base configuration
+# the extended rules or attributes of them will override the base rules.
+#
 # Another thing to know with extending configurations is you can chain together
 # multiple configuration files to a depth of 2. Allowlist arrays are appended
 # and can contain duplicates.


README.md
```diff
@@ -223,9 +223,10 @@ Gitleaks offers a configuration format you can follow to write your own secret d
 title = "Gitleaks title"
 
 # Extend the base (this) configuration. When you extend a configuration
-# the base rules take precedence over the extended rules. I.e., if there are
-# duplicate rules in both the base configuration and the extended configuration
-# the base rules will override the extended rules.
+# the extended rules take precedence over the base rules. I.e., if there are
```
rgmz, line 226: This isn't strictly true, as #1556 explicitly causes our config to overwrite values in the extended config.

I think we need better language to describe the relationship. "Base" is used ambiguously in a few places, e.g., in the code "base" refers to the extended rule [here](https://github.com/gitleaks/gitleaks/blob/8ea608595ce1c992166d710b217f59965d7a44d7/config/config.go#L244).

bufferoverflow: Yes, you are right. I somehow struggle a bit to suggest a proper text here. @rgmz Do you have a good suggestion?

amithm7: @rgmz

- "custom" would be a better word for the user-defined config. "base" can cause confusion, as it maybe misunderstood to refer the default config (gitleaks provided).
- Users would generally want "custom" user-defined config rules to take precedence (override) over the "inherited" rules from the extended / parent / default config.
- If there are exceptions on this overriding, it maybe noted.

I hope that these word suggestions help to form a better text here. Let me know if I can help.

bufferoverflow: @zricethezav @rgmz @amithm7 I rewrote the README.md regarding extending rules, please take a look.

bufferoverflow: ping

bufferoverflow:
@rgmz @zricethezav please close this PR if you see no value.

rgmz:
That's up to @zricethezav: I have no permissions. Personally, I think it's a good change.

PR #1582, opened 2024-10-19
Add entropy & allowlist to AWS

rgmz:
### Description:
This relates to #1577.

### Checklist:


zricethezav:
long overdue. Thanks 👍🏻

cmd/generate/config/rules/aws.go
```diff
@@ -14,15 +14,34 @@ func AWS() *config.Rule {
 		RuleID:      "aws-access-token",
 		Regex: regexp.MustCompile(
 			"(?:A3T[A-Z0-9]|AKIA|ASIA|ABIA|ACCA)[A-Z0-9]{16}"),
+		Entropy: 3,
 		Keywords: []string{
+			"A3T",
```
Ben-grmbl, line 19: While looking into AWS token formats, I could find no reference to A3T anywhere. Is this really an aws string?

https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_identifiers.html#identifiers-unique-ids
```
"AKIA", // Access key
"ASIA", // Temporary (AWS STS) access key
"ABIA", // AWS STS service bearer token
"ACCA", // Context-specific credential
```

rgmz: Ha, I'm not sure. I saw your earlier comment but decided to keep it in as I'm not an expert with AWS.

PR #1584, opened 2024-10-20
fix(aws-access-token): only match if correct length

Ben-grmbl:
### Description:
Previously, the rule `aws-access-token` matched for any long string containing e.g. ASIA, ACCA somewhere
This change checks for word boundaries before and after the secret.

Also the prefix (AKIA|ASIA|...) was moved from the non-capturing group into the secret-matching group,
since it is part of the token. So now the secret reported by gitleaks should be the actual AWS key that needs to be revoked.

Relevant rule change:
```diff
- regex =  '''(?:A3T[A-Z0-9]|AKIA|ASIA|ABIA|ACCA)[A-Z0-9]{16}'''
+ regex = '''\b((A3T[A-Z0-9]|AKIA|ASIA|ABIA|ACCA)[A-Z0-9]{16})(?:['|\"|\n|\r|\s|\x60|;]|$)'''
```

### Checklist:


cmd/generate/config/rules/aws.go
```diff
@@ -12,15 +13,15 @@ func AWS() *config.Rule {
 	r := config.Rule{
 		Description: "Identified a pattern that may indicate AWS credentials, risking unauthorized cloud resource access and data breaches on AWS platforms.",
 		RuleID:      "aws-access-token",
-		Regex: regexp.MustCompile(
-			"(?:A3T[A-Z0-9]|AKIA|ASIA|ABIA|ACCA)[A-Z0-9]{16}"),
-		Entropy: 3,
+		Regex:       utils.GenerateUniqueTokenRegex("(A3T[A-Z0-9]|AKIA|ASIA|ABIA|ACCA)[A-Z0-9]{16}", false),
```
rgmz, line 16:
```suggestion
		Regex:       utils.GenerateUniqueTokenRegex("(?:A3T[A-Z0-9]|AKIA|ASIA|ABIA|ACCA)[A-Z0-9]{16}", false),
```

The prefixes should be a non-capturing group. A capture group is only necessary when around the entire secret.

PR #1576, opened 2024-10-18
Add rule for curl `Authorization:` header

rgmz:
### Description:
This adds a rule to detect [curl header (-H, --header)](https://curl.se/docs/manpage.html#-H) auth on the same line, or up to 5 lines away.

### Checklist:

* [x] Does your PR pass tests?
* [x] Have you written new tests for your changes?
* [x] Have you lint your code locally prior to submission?

PR #1575, opened 2024-10-17
Add rule for curl `--user` flag

rgmz:
### Description:
This adds a rule to detect [curl basic auth (`-u` and `--user`)](https://curl.se/docs/manpage.html#-u) on the same line, or up to 5 lines away.

### Checklist:

* [x] Does your PR pass tests?
* [x] Have you written new tests for your changes?
* [x] Have you lint your code locally prior to submission?

PR #1586, opened 2024-10-20
More rule fixes

rgmz:
### Description:
This adds entropy & test cases to a few more rules.

It also tweaks the curl rules introduced in #1575 and #1576 to fix a pattern issue and reduce false positives.

### Checklist:

### DIFF
--- a/cmd/generate/config/rules/kubernetes.go
+++ b/cmd/generate/config/rules/kubernetes.go
@@ -17,7 +17,7 @@ func KubernetesSecret() *config.Rule {
 	// - valid base64 characters
 	// - longer than 10 characters (no "YmFyCg==")
 	//language=regexp
-	dataPat := `\bdata:(?:.|\s){0,100}?\s+([\w.-]+:(?:[ \t]*(?:\||>[-+]?)\s+)?[ \t]*(?:["']?[a-z0-9]{10,}={0,3}["']?|\{\{[ \t\w"|$:=,.-]+}}|""|''))`
+	dataPat := `\bdata:(?:.|\s){0,100}?\s+([\w.-]+:(?:[ \t]*(?:\||>[-+]?)\s+)?[ \t]*(?:["']?[a-z0-9+/]{10,}={0,3}["']?|\{\{[ \t\w"|$:=,.-]+}}|""|''))`
 
 	// define rule
 	r := config.Rule{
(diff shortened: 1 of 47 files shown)


cmd/generate/config/rules/slack.go
```diff
@@ -44,10 +44,11 @@ func SlackBotToken() *config.Rule {
 func SlackUserToken() *config.Rule {
 	// define rule
 	r := config.Rule{
-		Description: "Found a Slack User token, posing a risk of unauthorized user impersonation and data access within Slack workspaces.",
 		RuleID:      "slack-user-token",
+		Description: "Found a Slack User token, posing a risk of unauthorized user impersonation and data access within Slack workspaces.",
 		// The last segment seems to be consistently 32 characters. I've made it 28-34 just in case.
 		Regex:    regexp.MustCompile(`xox[pe](?:-[0-9]{10,13}){3}-[a-zA-Z0-9-]{28,34}`),
+		Entropy:  2,
```
rgmz, line 51: These are 2 because I haven't personally tested them. They can be bumped up in the future.

Ben-grmbl: These changes look great to me!

sorry, I accidentaly created a conflicting pull request #1593, which can be ignored for now. 
But it contains some code which I think could be useful for calculating a good entropy threshold at some point in the future.

zricethezav:
i dig it

cmd/generate/config/rules/kubernetes.go
DanielViard, line 1: Hello,

this change introduce false positive for us because we use some :
  kind: SecretStore
which are detected in place of 
  kind: Secret
  
 More detail about SecretStore available here : https://external-secrets.io/latest/api/secretstore/ 
 
 I don't know how to help to fix this false positive as I did not know why `+/ `as been added.
 
 Thank you !

rgmz: Hey @DanielViard,  `+/` were added because Kubernetes secrets are [base64 encoded](https://base64.guru/learn/base64-characters).

> this change introduce false positive for us because we use some :
> kind: SecretStore

Are you able to open an issue and provide an example of such a false positive? I wasn't able to reproduce this based on a quick test from the example you linked.

DanielViard: Hello,

Thank you for your answer!
Sorry I misread the detection it is a little more tricky, it is the `kind: SecretStore` in secretStoreRef which is detected:

```
apiVersion: external-secrets.io/v1beta1
kind: ExternalSecret
metadata:
  name: "example"
spec:
  # Optional, SecretStoreRef defines the default SecretStore to use when fetching the secret data.
  secretStoreRef:
    name: example
    kind: SecretStore
```
[externalsecret example](https://external-secrets.io/latest/api/externalsecret/#example)

Anyway I created a detailed issue with the step to reproduce : https://github.com/gitleaks/gitleaks/issues/1644
I hope it will help you to reproduce the issue.

PR #1585, opened 2024-10-20
Match left side of identifier

rgmz:
### Description:
This update the identifier pattern to include more context. This will make it easier to exclude false positives.

```diff
-"Match": "KeyToken=b77a5c561934e089\"",
+"Match": "PublicKeyToken=b77a5c561934e089\"",
 "Secret": "b77a5c561934e089",
```

Related: #1492 and #1578.

### Checklist:

* [ ] Does your PR pass tests?
* [ ] Have you written new tests for your changes?
* [ ] Have you lint your code locally prior to submission?

PR #1587, opened 2024-10-20
feat(generic-api-key): exclude keywords

### LINKED ISSUE #1578: Improving rule 'generic-api-key' to avoid false positives on e.g. "public_key=", "monkey=", ...

I am getting a lot of false positives with the rule 'generic-api-key'.
Currently, the rule searches for `key|api|token|secret|client|passwd|password|auth|access` in close vicinity to a random-looking string

This will currently match e.g. the following false positives:
```
public_key = "Gc3Q0RXJM2CUNbgKpUMnkPtOfOzeFzsW"     (this is intended to be public)
MantleAPI_version=9a038989604e8da62ecddbe2094b16ce1b778be1     (this is a git commit hash, used for versioning)
LLM_SECRET_NAME = "NEXUS-GPT4-API-KEY"         (entropy 3.7, this is merely the name of the environment variable)
keyword	= "Befaehigung_P2"     (entropy 3.5)
<Object mapKey="R648-MF-DEPARTURE" ... >  (entropy 3.5)

// from https://github.com/microsoft/vcpkg/blob/master/scripts/ci.baseline.txt
minisat-master-keying:x64-uwp=fail
monkeys-audio:mx64-uwp=fail
rapidstring:marm64-uwp=fail
```

I would like to exclude certain keywords from the match:
- `public_key, public_token`, because clearly they are meant to be public
- `keyword, monkey, donkey, keyboard, rapid, capital, ...`, because they merely contain the string api/key
- `secret_name, client_name, ...`, because the name of a secret is not a secret
- `api_version, client_version`, the version of an api should not be secret.
- `fake_secret, dummy_password`, ...
- ...

I think a manageable list of ~20 such keywords would cover >90% of false positives of this specific type

In order to do this, I suggest adding some regexes to the current AllowList, e.g.:
```golang
+ MatchCondition: config.AllowlistMatchOr,
+ Regexes: []*regexp.Regexp{
+ 	regexp.MustCompile(`(?i)` +
+ 		`(public.?key|public.?token` + // public key -> not a secret
+		`|api.?version|client.?version` + // version info -> not a secret
+		`|map.?key|key.?word|monkey|donkey|keyboard|keying|keystone` + // common words containing "key"
+		`|rapid|capital` + // common words containing "api"
+		`|secret.?name|client.?name|key.?name` + // name of secret, not the secret itself
+		`|fake|dummy)`),  // 
+ },
+ RegexTarget: "match",
  StopWords:   DefaultStopWords,
```

in order to make RegexTarget: "match" work for the allowList regex, the secret matching regex would have to be slightly modified:
```
Regex: utils.GenerateSemiGenericRegex([]string{
!	`\w{0,10}key`,      // <- to include e.g. "public_key", "rapid" in the capture
!	`\w{0,10}api`,
!	`\w{0,10}token`,
!	"\w{0,10}secret",  // <- to include e.g. "fake_secret, "dummy_secret" in the capture
	"client",
	"passwd",
	"password",
	"auth",
	"access",
}, `[0-9a-z\-_.=]{10,150}`, true),
```
With that, RegexTarget: "match" can be used. RegexTarget: "line" is suboptimal, because a single line could contain e.g. a public_key **and** a secret key.

- With negative look-behind / look-ahead groups this could be accomplished with only the secret matching Regex, without having an allowList regex. I believe those are not supported by the current regex implementation.

- Tinkering with the minimum entropy might also help reduce some of the false positives? The current value of 3.5 seems very low, based on the examples above.

**Additional context**

I'd be happy to create a pull request for this.

If there is anything I missed, or if there is a problem with this approach, please let me know.
I would also be interested in alternative approaches.

cc @zricethezav

Ben-grmbl:
closes #1578

### Description:

Previously, the rule "generic-api-key"  looked for key|api|token|secret|client|passwd|password|auth|access in close vicinity to a random-looking string

With this change, some false-positives are prevented, by adding the following rules to the allowList:
```golang
`public[_.-]?(key|token)` + // public key -> not a secret
`|(api|client)[_.-]?(version|id)` + // version/id -> not a secret
`|(secret|client)[_.-]?name` + // name of client, name of env variable, ...
`|key[_.-]?word|monkey|donkey|keyboard|keying|keystone` + // common words containing "key"
`|rapid|capital` + // common words containing "api"
```

⚠ Previously the **true-positive** examples contained
```
`"client_id" : "0afae57f3ccfd9d7f5767067bc48b30f719e271ba470488056e37ab35d4b6506"`,
```
The new rule explicitly does not match client_id anymore. (with the assumption, that a client_id is usually not intended to be secret)
If this assumption is unwarranted, I can revert that.

### other Changes
- I added "jetbrains" and "snapshot" to the generic-stopwords list, in order to prevent the following false positives:
```
`<entry key="jetbrains.mps.v8_elimination" value="executed" />`, // file generated by Jetbrains IntelliJ
`event-bus-message-api:rc0.15.0_20231217_1420-SNAPSHOT'`, // contains "snapshot" ->a version number, not a secret
```

- I stumbled on a certain flakiness when generating true-positive examples with `secrets.NewSecret`: The list of stopwords is very long, and contains a lot of very short stopwords. The chances of randomly generating a string with one of these words is quite significant (~5%?)

---------

this pull requests uses the new increased context in front of the match from #1585

I wanted to try that new feature out, but if this pull request conflicts with newer changes or someone else is working on the same thing concurrently, feel free to ignore all of this.

### Checklist:


cmd/generate/config/rules/generic.go
```diff
@@ -35,7 +37,20 @@ func GenericCredential() *config.Rule {
 		Entropy: 3.5,
 		Allowlists: []config.Allowlist{
 			{
-				StopWords: DefaultStopWords,
+				Description:    "Allowlist for Generic API Keys",
+				MatchCondition: config.AllowlistMatchOr,
+				Regexes: []*regexp.Regexp{
+					regexp.MustCompile(`(?i)` +
+						`(public.?key|public.?token` + // public key -> not a secret
```
rgmz, line 44: Personally, I'd suggest grouping common prefixes and suffixes. e.g.,
```suggestion
						`(public[_.-]?(key|token)` + // public key -> not a secret
```

cmd/generate/config/rules/generic.go
```diff
@@ -35,7 +37,20 @@ func GenericCredential() *config.Rule {
...
+						`|api.?version|client.?version` + // version of api -> not a secret
+						`|api.?id|client.?id` + // generally, the id is not a secret
+						`|secret.?name|client.?name` + // name of secret, not the secret itself
+						`|map.?key|key.?word|monkey|donkey|keyboard|keying|keystone` + // common words containing "key"
```
rgmz, line 48: I agree with all these except `mapKey`. I have seen some legitimate tokens under such a name.
```suggestion
						`|key.?word|monkey|donkey|keyboard|keying|keystone` + // common words containing "key"
```

cmd/generate/config/rules/generic.go
Ben-grmbl: > more specific separators

makes sense, especially considering the current regex for including context is [\w.-]{0,10}?, so there will be no other separators.

> grouping

also a great idea

config/gitleaks.toml
```diff
@@ -519,7 +519,10 @@ keywords = [
 ]
 # NOTE: Gitleaks >= v8.21.0 should use [[rules.allowlists] instead.
 [rules.allowlist]
-stopwords = [
+regexTarget = "match"
+regexes = [
+    '''(?i)(public[_.-]?(key|token)|(api|client)[_.-]?(version|id)|(secret|client)[_.-]?name|key[_.-]?word|monkey|donkey|keyboard|keying|keystone|rapid|capital)''',
+]stopwords = [
```
rgmz, line 525: Seems like the template is a little broken. 

This should fix it:
```diff
diff --git a/cmd/generate/config/rules/config.tmpl b/cmd/generate/config/rules/config.tmpl
index 9597b8f..2e79aa3 100644
--- a/cmd/generate/config/rules/config.tmpl
+++ b/cmd/generate/config/rules/config.tmpl
@@ -52,20 +52,18 @@ tags = [
 {{- if eq (len $rule.Allowlists) 1 }}{{ with index $rule.Allowlists 0}}{{ if or .Regexes .Paths .Commits .StopWords }}
 # NOTE: Gitleaks >= v8.21.0 should use [[rules.allowlists] instead.
 [rules.allowlist]
-{{ with .MatchCondition }}condition = "{{ . }}"
-{{ end -}}
-{{- with .Commits }}commits = [
+condition = "{{ .MatchCondition.String }}"
+{{- with .Commits -}}{{println}}commits = [
     {{ range $j, $commit := . }}"{{ $commit }}",{{ end }}
 ]{{ end }}
-{{- with .Paths }}paths = [
+{{- with .Paths }}{{println}}paths = [
     {{ range $j, $path := . }}'''{{ $path }}''',{{ end }}
 ]{{ end }}
-{{- with .RegexTarget }}regexTarget = "{{ . }}"
-{{ end -}}
-{{- with .Regexes }}regexes = [{{ range $i, $regex := . }}
+{{- if and .RegexTarget .Regexes }}{{println}}regexTarget = "{{ .RegexTarget }}"{{ end -}}
+{{- with .Regexes }}{{println}}regexes = [{{ range $i, $regex := . }}
     '''{{ $regex }}''',{{ end }}
 ]{{ end }}
-{{- with .StopWords }}stopwords = [{{ range $j, $stopword := . }}
+{{- with .StopWords }}{{println}}stopwords = [{{ range $j, $stopword := . }}
     "{{ $stopword }}",{{ end }}
 ]{{ end }}{{ end }}{{ end }}
 {{ else }}
@@ -84,7 +82,7 @@ tags = [
 {{- with $allowlist.Regexes }}regexes = [{{ range $i, $regex := . }}
     '''{{ $regex }}''',{{ end }}
 ]{{ end }}
-{{- with $allowlist.StopWords }}stopwords = [{{ range $j, $stopword := . }}
+{{ with $allowlist.StopWords }}stopwords = [{{ range $j, $stopword := . }}
     "{{ $stopword }}",{{ end }}
 ]{{ end }}{{ end }}{{ end }}
 {{ end }}{{ end }}
```

Ben-grmbl: Thanks! I have been trying to figure this out for the last hour, and could not make it work.

One thing: after applying this patch, I got the new line `condition = "OR" on 7 rules, which previously had no such setting. Is that intended?

e.g. 
```patch
diff --git a/config/gitleaks.toml b/config/gitleaks.toml
index 4d62cf3..a0370a1 100644
--- a/config/gitleaks.toml
+++ b/config/gitleaks.toml
@@ -157,7 +157,6 @@ keywords = [
 ]
 # NOTE: Gitleaks >= v8.21.0 should use [[rules.allowlists] instead.
 [rules.allowlist]
-condition = "OR"
 regexes = [
     '''.+EXAMPLE$''',
 ]
```

I tried to fix that by thinkering like this:
```patch
diff --git a/cmd/generate/config/rules/config.tmpl b/cmd/generate/config/rules/config.tmpl
index 2e79aa3..a0729d9 100644
--- a/cmd/generate/config/rules/config.tmpl
+++ b/cmd/generate/config/rules/config.tmpl
@@ -52,7 +52,7 @@ tags = [
 {{- if eq (len $rule.Allowlists) 1 }}{{ with index $rule.Allowlists 0}}{{ if or .Regexes .Paths .Commits .StopWords }}
 # NOTE: Gitleaks >= v8.21.0 should use [[rules.allowlists] instead.
 [rules.allowlist]
-condition = "{{ .MatchCondition.String }}"
+{{- with .MatchCondition }}{{println}}condition = "{{ .MatchCondition.String }}"{{ end }}
 {{- with .Commits -}}{{println}}commits = [
     {{ range $j, $commit := . }}"{{ $commit }}",{{ end }}
 ]{{ end }}
```

Now the toml file appears to look normally again.

rgmz: > One thing: after applying this patch, I got the new line `condition = "OR" on 7 rules, which previously had no such setting. Is that intended?

I hadn't really put much thought into it. The default value for it is `OR` so I suppose it doesn't need to be included.

cmd/generate/config/rules/generic.go
```diff
@@ -35,7 +37,19 @@ func GenericCredential() *config.Rule {
...
 			{
-				StopWords: DefaultStopWords,
+				Description:    "Allowlist for Generic API Keys",
+				MatchCondition: config.AllowlistMatchOr,
+				Regexes: []*regexp.Regexp{
+					regexp.MustCompile(`(?i)(` +
+						`public[_.-]?(key|token)` + // public key -> not a secret
+						`|(api|client)[_.-]?(version|id)` + // version/id -> not a secret
+						`|(secret|client)[_.-]?name` + // name of client, name of env variable, ...
+						`|key[_.-]?word|monkey|donkey|keyboard|keying|keystone` + // common words containing "key"
```
rgmz, line 47: Is it possible that `monkey` and `donkey` could inadvertently match something like `...monKey`? Perhaps we should require a specific case.

```suggestion
						`|key[_.-]?word|(?-i:[DdMm]onkey|[DM]ONKEY)|keyboard|keying|keystone` + // common words containing "key"
```

Ben-grmbl: Good point. And I was not even aware you could switch case-invariant off again.

Ben-grmbl: I am wondering, whether it would be better to limit the keywords:
instead of matching case-invariant for "key", only match for 
- "Key"
- "key", not preceded or followed by lowercase letter, 
- "KEY", not preceded or followed by uppercase letter
something like this:
```regex
(?-i:(?:Key[^a-z])|(?:(?:\b|[^a-z\s])key[^a-z])|(?:(?:\b|[^A-Z\s])KEY[^A-Z]))
```
Then monkey, donkey would automatically be excluded. But it seems a little convoluted.

rgmz: > ```regexp
> (?-i:(?:Key[^a-z])|(?:(?:\b|[^a-z\s])key[^a-z])|(?:(?:\b|[^A-Z\s])KEY[^A-Z]))
> ```
> 
> Then monkey, donkey would automatically be excluded. But it seems a little convoluted.

That is a clever pattern; unfortunately, I think this has to be explicitly tweaked on a case-by-case basis.

Otherwise, there's the risk of false negatives such as:
```yaml
snyk:
  apikey: 04def1de-680f-4340-bdad-78fabf18380f
```

zricethezav:
> The new rule explicitly does not match client_id anymore. (with the assumption, that a client_id is usually not intended to be secret)

I'm not sure I 100% agree with this. While `client_id` may not be a secret, it's likely required to authenticate a secret. But, because this is change is specifically targeting the generic-api-key rule, I think excluding `client_id` is acceptable.

cmd/generate/config/rules/config.tmpl
```diff
@@ -52,20 +52,18 @@ tags = [
...
 [rules.allowlist]
-{{ with .MatchCondition }}condition = "{{ . }}"
-{{ end -}}
-{{- with .Commits }}commits = [
+{{- with .MatchCondition }}{{println}}condition = "{{ .MatchCondition.String }}"{{ end }}
+{{- with .Commits -}}{{println}}commits = [
     {{ range $j, $commit := . }}"{{ $commit }}",{{ end }}
 ]{{ end }}
-{{- with .Paths }}paths = [
+{{- with .Paths }}{{println}}paths = [
```
zricethezav, line 59: TIL `println`

rgmz:
> I'm not sure I 100% agree with this. While `client_id` may not be a secret, it's likely required to authenticate a secret. But, because this is change is specifically targeting the generic-api-key rule, I think excluding `client_id` is acceptable.

In my experience `client` is >90% false positives. The remaining 10% almost entirely overlap with other keywords like `auth` and `secret`.

IMO we should swap `client` for `cred(ential)?s?`.

zricethezav:
@rgmz

> In my experience client is >90% false positives. The remaining 10% almost entirely overlap with other keywords like auth and secret.
> IMO we should swap client for cred(ential)?s?.

Sounds reasonable and like a followup PR unless @Ben-grmbl wants to include it in this PR

zricethezav (approved):

Ben-grmbl:
> In my experience `client` is >90% false positives. The remaining 10% almost entirely overlap with other keywords like `auth` and `secret`.
> 
> IMO we should swap `client` for `cred(ential)?s?`.

I agree that in my experience client.* has too many false positives.

"credential..." should work nicely, "creds" seems also like a good specific abbreviation.

Regarding the abbreviation "cred": might have to exclude e.g. "credit", "credible", "incredulity", "accreditation". Is "cred" as abbreviation common enough for secrets?

rgmz:
> Regarding the abbreviation "cred": might have to exclude e.g. "credit", "credible", "incredulity", "accreditation", .... Is "cred" as abbreviation common enough for secrets?

It is in my experience. However, it also introduces a number of false positives as you've mentioned.

I'm undecided whether it should be included. Thoughts, @bplaxco?

Ben-grmbl:
I updated this branch to 


- remove "client" from the list of keywords, as suggested
- add "credential", "creds", to the list of keywords
- (cred* was not yet added, due to the need to exclude more false positives)
- add more suggestions by @rgmz  from https://github.com/gitleaks/gitleaks/issues/1578#issuecomment-2432882552 to reduce key-related false positives
  - primary-key, keyStoreType, issuerKeyHash, KEY_UP, ...
- include all recent changes from master

 
Unfortunately I made a mistake yesterday, and did not push the newest gitleaks.toml; this has been fixed now.

rgmz (approved):

zricethezav:
🙇🏻 great work @Ben-grmbl @rgmz

PR #1592, opened 2024-10-21
Fix config validate rules when [extend] is used

rgmz:
### Description:
This fixes an issue where rules were not being validated when the config contained `[extend]`.

Because `extendDepth` is globally mutable, `if extendDepth == 0 {` was never true.

### Checklist:


rgmz:
@zricethezav Thoughts?

PR #1593, opened 2024-10-22
slightly flaky build - adjust entropy, tps, for sumologic-access-id, plaidAccessId, vault-service-token

Ben-grmbl:
### Description:
`go generate` was slightly flaky, failing sometimes for the rules "sumologic-access-id", "plaid-client-id", "vault-service-token"

There was a small but significant chance that the true positive samples generated with `secrets.NewSecret()` would not match the minimum entropy.

I calculated a reasonably minimum threshold for the entropy, so that ~99.999% of random strings should have higher entropy.

#### changes:
- plaid-client-id: entropy 3.5 -> 3.4
- sumologic-access-id: entropy 3 -> 2.5
- hashicorp: entropy kept at 3.5, only adjusted the tps sample to include uppercase

The entropy of the generated true positives was also slightly lower than expected, since `secrets.NewSecret()` will only generate uppercase letters if they are explicitely included in the regex, so no uppercase for e.g. `secrets.NewSecret(utils.AlphaNumeric("24")`

I explicitely added the uppercase letters to the NewSecret calls for now. Not sure if there is a more elegant solution:
```diff
- utils.GenerateSampleSecret("vault", "s."+secrets.NewSecret(utils.AlphaNumeric("24"))),
+ utils.GenerateSampleSecret("vault", secrets.NewSecret(`s\.[a-zA-Z0-9]{24}`)),
```

Here are my calculated entropy thresholds:

| rule | regex | min entropy (matching 99.999% of strings) |
|------|-------|-------------|
|sumologic-access-id           | `su[a-zA-Z0-9]{12}`      | 2.638460          |
|                                        | `su[a-z0-9]{12}`      | 2.403677          |
| plaidAccessId                   | `[a-zA-Z0-9]{24}`    |  3.516151          |
|                                       | `[a-z0-9]{24}`    |  3.184561         |
| vault-service-token            | `s\.[a-zA-Z0-9]{24}`    |  3.642371       |
|                                       | `s\.[a-z0-9]{24}`  |  3.348084      |
|                                       | `hvs\.[\w-]{90}`  |  5.055993      |
|                                       | `hvs\.[a-z0-9_\-]{90}`  |  4.586675      |

I calculated these values with this code (not commited):
```golang
g, err := reggen.NewGenerator(regex)
if err != nil {
	panic(err)
}

entropy_samples := []float64{}
for i := 0; i < 1000000; i++ {
	secret := g.Generate(1)
	entropy := shannonEntropy(secret)
	entropy_samples = append(entropy_samples, entropy)
}
sort.Float64s(entropy_samples)
return entropy_samples[10]
```

### Checklist:

### DIFF
--- a/cmd/generate/config/rules/hashicorp_vault.go
+++ b/cmd/generate/config/rules/hashicorp_vault.go
@@ -28,10 +28,10 @@ func VaultServiceToken() *config.Rule {
 	// validate
 	tps := []string{
 		// Old
-		utils.GenerateSampleSecret("vault", "s."+secrets.NewSecret(utils.AlphaNumeric("24"))),
+		utils.GenerateSampleSecret("vault", secrets.NewSecret(`s\.[a-zA-Z0-9]{24}`)),
 		`token: s.ZC9Ecf4M5g9o34Q6RkzGsj0z`,
 		// New
-		utils.GenerateSampleSecret("vault", "hvs."+secrets.NewSecret(utils.AlphaNumericExtendedShort("90"))),
+		utils.GenerateSampleSecret("vault", secrets.NewSecret(`hvs\.[\w\-]{90}`)),
 		`-vaultToken hvs.CAESIP2jTxc9S2K7Z6CtcFWQv7-044m_oSsxnPE1H3nF89l3GiYKHGh2cy5sQmlIZVNyTWJNcDRsYWJpQjlhYjVlb1cQh6PL8wEYAg"`, // longer than 100 chars
 	}
 	fps := []string{
--- a/cmd/generate/config/rules/plaid.go
+++ b/cmd/generate/config/rules/plaid.go
@@ -23,7 +23,7 @@ func PlaidAccessID() *config.Rule {
 
 	// validate
 	tps := []string{
-		utils.GenerateSampleSecret("plaid", secrets.NewSecret(utils.AlphaNumeric("24"))),
+		utils.GenerateSampleSecret("plaid", secrets.NewSecret(`[a-zA-Z0-9]{24}`)),
 	}
 	return utils.Validate(r, tps, nil)
 }
--- a/cmd/generate/config/rules/sumologic.go
+++ b/cmd/generate/config/rules/sumologic.go
@@ -25,7 +25,7 @@ func SumoLogicAccessID() *config.Rule {
 		`sumologic_access_id = "sug5XpdpaoxtOH"`,     // gitleaks:allow
 		`export SUMOLOGIC_ACCESSID="suDbJw97o9WVo0"`, // gitleaks:allow
 		`SUMO_ACCESS_ID = "suGyI5imvADdvU"`,          // gitleaks:allow
-		utils.GenerateSampleSecret("sumo", "su"+secrets.NewSecret(utils.AlphaNumeric("12"))),
+		utils.GenerateSampleSecret("sumo", secrets.NewSecret(`su[a-zA-Z0-9]{12}`)),
 	}
 	fps := []string{
 		`- (NSNumber *)sumOfProperty:(NSString *)property;`,


rgmz:
Another option is to tweak the generate function to loop until it generates a secret with sufficient entropy. It feels sub-optimal to lower the threshold for rules because of an issue with our tests.

Ben-grmbl:
> @Ben-grmbl Oh good catch! I think it makes sense to throw `[a-zA-Z0-9]{24}` into a function, wdyt?

It certainly makes sense. 

We could even have some functions like utils.AlphaNumericCaseSensitive(24), with the slight drawback that it will be much longer than "[a-zA-Z0-9]{24}", and only usefull for generating secrets, not for capturing.

Another option would be to have a parameter caseInvariant=true/false for generating secrets, but to properly implement that I think it would be neccessary to rewrite the lucasjones/reggen generator.

rgmz:
> I think it would be neccessary to rewrite the lucasjones/reggen generator.

That's probably not worth it. I'm hoping we can switch to a different approach (#1499).

Ben-grmbl:
> Another option is to tweak the generate function to loop until it generates a secret with sufficient entropy. It feels sub-optimal to lower the threshold for rules because of an issue with our tests.

To make builds reproducible, a fixed seed would also work. (Possibly configurable with an environment variable?)

But I would not necessarily say this is lowering the threshold for the sake of the tests. The tests just show the real chance of rejecting a valid token as invalid.

e.g. for "su[a-zA-Z0-9]{12}"
- if 99.999% of all valid tokens should be detected, min entropy needs to be below ~2.6
- if 99.99% of all valid tokens should be detected, min entropy needs to be below ~2.8
- if 99.9% of all valid tokens should be detected, min entropy needs to be below ~3.0
- if 99% of all valid tokens should be detected, min entropy needs to be below ~3.2

So, the current value of 3.0 should be sufficient if 99.9% is an acceptable detection rate. If the detection rate needs to be higher, the entropy threshold needs to be a bit lower.

rgmz:
> So, the current value of 3.0 for sumologic-access-id should be sufficient if 99.9% is an acceptable detection rate.
> In that case I can remove that change from the pull request.

I think that's fine. Possible values for `su[a-zA-Z0-9]{12}` don't necessarily equate possible values for a SumoLogic access ID (e.g.,`suAAAAAAAAAAAA`). None of the generated IDs that I've tested have had an entropy below 3.5.

zricethezav:
> if 99.9% of all valid tokens should be detected, min entropy needs to be below ~3.0

@Ben-grmbl Let's bump sumo back up to 3.0

Ben-grmbl:
Sorry for taking a while for this. 
I removed the entropy changes, leaving only the changed true positive patterns, and ran some tests in a loop. 

The chance of a failing go generate with the sumologic-access-id rule with entropy at 3.0 is now fairly low, about 1/1000.

However, I ran into another problem: the chance of the "vault-service-token" rule failing is quite high, about 1/50 builds fail.

This is however not due to the entropy, as I first assumed. The rule contains the following allowList in hashicorp.go:
```golang
			{
				Regexes: []*regexp.Regexp{
					// https://github.com/gitleaks/gitleaks/issues/1490#issuecomment-2334166357
					regexp.MustCompile(`s\.[A-Za-z]{24}`),
				},
			},
```

The linked comment says:
> If the tokens are randomly generated, I'd guess the chances of it only being letters is fairly low.
> 
> pow(24, 26 + 26) / pow(24, 26 + 26 + 10)
> 
> So if the math is right, a 0.000000000002% chance that `s\.[a-zA-Z]{24}\b` will miss a true positive if the keys are random.

However, this does not seem to be correct.
The chance of generating only letters, when picking from A-Z,a-z,0-9, are 

P = (52 / 62) ^ 24 = 1.46 %

So 1.46% of go generate will currently fail.

Previously, before changing the true positive to contain upper- and lowercase letters, the chance of the build failing was only P = (26 / 36) ^ 24 = 0.04 %

PR #1598, opened 2024-10-23
Detect base64 secrets with generic-api-key

rgmz:
### Description:
This fixes an inconsistency where a baes64 encoded secret value (such as `abdcefg==`) would be detected, but other ones with valid characters `+` and `/` weren't.

### Checklist:

### DIFF
--- a/cmd/generate/config/rules/generic.go
+++ b/cmd/generate/config/rules/generic.go
@@ -19,11 +19,10 @@ func GenericCredential() *config.Rule {
 			"credential",
 			"creds",
 			"key",
-			"passwd",
-			"password",
+			"passw(?:or)?d",
 			"secret",
 			"token",
-		}, `[\w.=-]{10,150}`, true),
+		}, `[\w.=-]{10,150}|[a-z0-9][a-z0-9+/]{11,}={0,3}`, true),
 		Keywords: []string{
 			"access",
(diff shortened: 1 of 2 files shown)


cmd/generate/config/rules/generic.go
```diff
@@ -19,11 +19,10 @@ func GenericCredential() *config.Rule {
 			"credential",
 			"creds",
 			"key",
-			"passwd",
-			"password",
+			"passw(or)?d",
 			"secret",
 			"token",
-		}, `[\w.=-]{10,150}`, true),
+		}, `(?:[0-9a-z\-_.=]{10,150}|[a-z0-9][a-z0-9+/]{11,}={0,3})`, true),
```
zricethezav, line 25:
```suggestion
		}, `[\w.=-]{10,150}|[a-z0-9][a-z0-9+/]{11,}={0,3}`, true),
```

Since this gets wrapped in a capture group I don't think we need `(?:` and I think we should use `\w` for consistency.

zricethezav: I was just referring to the non b64 pattern: `[0-9a-z\-_.=]{10,150}`

rgmz: Oh, duh. Done!

Issue #1402, opened 2024-05-07
GitLeaks does not detect Jira api keys in source code


Maksym-Hrabar:
During the use of gitleaks, it was found that it does not detect Jira API keys.

We found a rule that should check these key types, but it doesn't work:
```
[[rules]]
id = "atlassian-api-token"
description = "Detected an Atlassian API token, posing a threat to project management and collaboration tool security and data confidentiality."
regex = '''(?i)(?:atlassian|confluence|jira)(?:[0-9a-z\-_\t .]{0,20})(?:[\s|']|[\s|"]){0,3}(?:=|>|:{1,3}=|\|\|:|<=|=>|:|\?=)(?:'|\"|\s|=|\x60){0,5}([a-z0-9]{24})(?:['|\"|\n|\r|\s|\x60|;]|$)'''
keywords = [
    "atlassian","confluence","jira",
]
```

For the test, we used this key: "ATATT3xFfGF0K3irG5tKKi-6u-wwaXQFeGwZ-IHR-hQ3CulkKtMSuteRQFfLZ6jihHThzZCg_UjnDt-4Wl_gIRf4zrZJs5JqaeuBhsfJ4W5GD6yGg3W7903gbvaxZPBjxIQQ7BgFDSkPS8oPispw4KLz56mdK-G6CIvLO6hHRrZHY0Q3tvJ6JxE=C63992E6"

This key in the source code wasn't detected.

PR #1601, opened 2024-10-25
Add new jira api token pattern

oscarbc96:
### Description:

This PR adds support for Jira’s new API token pattern by introducing a new regular expression to detect it. Additionally, it includes a MergeRegexps function that enables the merging of two or more regular expressions, ensuring backward compatibility with previous patterns. This allows the system to flexibly detect both the old and new patterns.

Fixes https://github.com/gitleaks/gitleaks/issues/1402

### Checklist:

### DIFF
--- a/cmd/generate/config/rules/atlassian.go
+++ b/cmd/generate/config/rules/atlassian.go
@@ -14,13 +14,13 @@ func Atlassian() *config.Rule {
 		Regex: utils.MergeRegexps(
 			utils.GenerateSemiGenericRegex(
 				[]string{"atlassian", "confluence", "jira"},
-				utils.AlphaNumeric("24"),
-				true,
+				`[a-zA-Z0-9{24}`,
+				false,
 			),
-			utils.GenerateUniqueTokenRegex(`ATATT3[A-Za-z0-9_\-=]{186}`, false),
+			utils.GenerateUniqueTokenRegex(`ATATT3[\w-]{177}=[A-F0-9]{8}`, false),
 		),
 		Entropy:  3.5,
-		Keywords: []string{"atlassian", "confluence", "jira"},
+		Keywords: []string{"atlassian", "confluence", "jira", "atatt3"},
 	}
 
 	// validate
--- a/cmd/generate/config/rules/atlassian.go
+++ b/cmd/generate/config/rules/atlassian.go
@@ -14,10 +14,10 @@ func Atlassian() *config.Rule {
 		Regex: utils.MergeRegexps(
 			utils.GenerateSemiGenericRegex(
 				[]string{"atlassian", "confluence", "jira"},
-				`[a-zA-Z0-9{24}`,
+				`[a-zA-Z0-9]{24}`,
 				false,
 			),
-			utils.GenerateUniqueTokenRegex(`ATATT3[\w-]{177}=[A-F0-9]{8}`, false),
+			utils.GenerateUniqueTokenRegex(`ATATT3[A-Za-z0-9_\-=]{186}`, false),
 		),
 		Entropy:  3.5,
 		Keywords: []string{"atlassian", "confluence", "jira", "atatt3"},
--- a/cmd/generate/config/rules/atlassian.go
+++ b/cmd/generate/config/rules/atlassian.go
@@ -11,8 +11,18 @@ func Atlassian() *config.Rule {
 	r := config.Rule{
 		Description: "Detected an Atlassian API token, posing a threat to project management and collaboration tool security and data confidentiality.",
 		RuleID:      "atlassian-api-token",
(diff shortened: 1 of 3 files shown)


cmd/generate/config/rules/atlassian.go
```diff
@@ -11,15 +11,26 @@ func Atlassian() *config.Rule {
...
+				[]string{"atlassian", "confluence", "jira"},
+				utils.AlphaNumeric("24"),
+				true,
+			),
+			utils.GenerateSemiGenericRegex(
+				[]string{"atlassian", "confluence", "jira"},
+				`[A-Za-z0-9_\-=]{192}`,
+				false,
+			),
+		),
```
rgmz, line 25: Might as well add entropy while we're changing this. Every old token I've found is _exactly_ 4.00, newer ones are > 5. I think 3.5 is a good option.
```suggestion
		),
		Entropy: 3.5,
```

cmd/generate/config/rules/atlassian.go
```diff
@@ -11,15 +11,26 @@ func Atlassian() *config.Rule {
...
 		RuleID:      "atlassian-api-token",
-		Regex: utils.GenerateSemiGenericRegex([]string{
-			"atlassian", "confluence", "jira"}, utils.AlphaNumeric("24"), true),
+		Regex: utils.MergeRegexps(
+			utils.GenerateSemiGenericRegex(
+				[]string{"atlassian", "confluence", "jira"},
+				utils.AlphaNumeric("24"),
+				true,
+			),
+			utils.GenerateSemiGenericRegex(
```
rgmz, line 20: Every token I've found has the prefix `ATATT3xFfGF0`. It seems like this could instead be
```suggestion
			utils.GenerateUniqueTokenRegex(`ATATT3[A-Za-z0-9_\-=]{186}`, false),
```

rgmz (changes requested):

cmd/generate/config/rules/atlassian.go
rgmz: Moreover, they all seem to have the suffix `=[a-fA-F0-9]{8}`. `=` doesn't seem to be valid outside of that.

cmd/generate/config/rules/atlassian.go
oscarbc96: Added. Thank you!

cmd/generate/config/rules/atlassian.go
```diff
@@ -11,15 +11,23 @@ func Atlassian() *config.Rule {
 	r := config.Rule{
 		Description: "Detected an Atlassian API token, posing a threat to project management and collaboration tool security and data confidentiality.",
 		RuleID:      "atlassian-api-token",
-		Regex: utils.GenerateSemiGenericRegex([]string{
-			"atlassian", "confluence", "jira"}, utils.AlphaNumeric("24"), true),
+		Regex: utils.MergeRegexps(
+			utils.GenerateSemiGenericRegex(
+				[]string{"atlassian", "confluence", "jira"},
+				utils.AlphaNumeric("24"),
+				true,
```
rgmz, line 18: Testing this a bit more, I think the first one needs to be changed as follows:
```suggestion
				`[a-zA-Z0-9{24}`,
				false,
```

Otherwise, the `(?i)` at the start of the pattern also makes `ATATT3` case-insensitive. I would have thought `(?:(?i)a|b)` would only apply to `a` but it applies to both.

cmd/generate/config/rules/atlassian.go
```diff
@@ -11,15 +11,23 @@ func Atlassian() *config.Rule {
...
+			),
+			utils.GenerateUniqueTokenRegex(`ATATT3[A-Za-z0-9_\-=]{186}`, false),
```
rgmz, line 20: Based on my testing, this can be more specific:
```suggestion
			utils.GenerateUniqueTokenRegex(`ATATT3[\w-]{177}=[A-F0-9]{8}`, false),
```

cmd/generate/config/rules/atlassian.go
```diff
@@ -11,15 +11,23 @@ func Atlassian() *config.Rule {
...
+		),
+		Entropy:  3.5,
 		Keywords: []string{"atlassian", "confluence", "jira"},
```
rgmz, line 23: I think the newer token is unique enough that we can add the prefix to the list of keywords.
```suggestion
		Keywords: []string{"atlassian", "confluence", "jira", "atatt3"},
```

rgmz (commented):
A few suggestions after further testing. Looks good otherwise!

cmd/generate/config/rules/atlassian.go
oscarbc96: I didn't know that. I will change it

PR #1602, opened 2024-10-25
Create Octopus Deploy api key

rgmz:
### Description:
Detects API keys for [Octopus Deploy](https://octopus.com/docs/octopus-rest-api/examples/users-and-teams/create-api-key#script).

### Checklist:

### DIFF
--- a/cmd/generate/config/rules/octopusdeploy.go
+++ b/cmd/generate/config/rules/octopusdeploy.go
@@ -1,0 +1,32 @@
+package rules
+
+import (
+	"github.com/zricethezav/gitleaks/v8/cmd/generate/config/utils"
+	"github.com/zricethezav/gitleaks/v8/cmd/generate/secrets"
+	"github.com/zricethezav/gitleaks/v8/config"
+)
+
+func OctopusDeployApiKey() *config.Rule {
+	// define rule
+	r := config.Rule{
+		RuleID:      "octopus-deploy-api-key",
+		Description: "Discovered a potential Octopus Deploy API key, risking application deployments and operational security.",
+		Regex:       utils.GenerateUniqueTokenRegex(`API-[A-Z0-9]{26}`, false),
+		Entropy:     3,
+		Keywords:    []string{"api-"},
+	}
+
+	// validate
+	tps := []string{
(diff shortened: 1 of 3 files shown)


cmd/generate/config/rules/octopusdeploy.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+func OctopusDeployApiKey() *config.Rule {
+	// define rule
+	r := config.Rule{
+		RuleID:      "octopus-deploy-api-key",
+		Description: "Discovered a potential Octopus Deploy API key, risking application deployments and operational security.",
+		Regex:       utils.GenerateUniqueTokenRegex(`API-[A-Z0-9]{26}`, false),
+		Entropy:     3,
+		Keywords:    []string{"api-"},
```
zricethezav, line 16: @rgmz Should `octopus` be a keyword as well?

rgmz: I'm undecided on that. Using `API-` as the keyword could either be high confidence (more so than "octopus"), or result in false positives.

I don't know of any other providers that have the specific format `API-[A-Z0-9]{26}`.

cmd/generate/config/rules/octopusdeploy.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/zricethezav/gitleaks/v8/cmd/generate/secrets"
+	"github.com/zricethezav/gitleaks/v8/config"
+)
+
+func OctopusDeployApiKey() *config.Rule {
+	// define rule
+	r := config.Rule{
+		RuleID:      "octopus-deploy-api-key",
+		Description: "Discovered a potential Octopus Deploy API key, risking application deployments and operational security.",
+		Regex:       utils.GenerateUniqueTokenRegex(`API-[A-Z0-9]{26}`, false),
```
zricethezav, line 14: @rgmz I'm concerned this would lead to some false positives since `API-` could be fairly common. We can keep an eye on this rule. Probably fine as is though.

cmd/generate/config/rules/octopusdeploy.go
zricethezav: 👍🏻 lets just keep an eye on this rule

PR #1603, opened 2024-10-25
Further improvements to curl rules

### LINKED ISSUE #1589: False positives for curl-auth rule if variables are quoted
**Describe the bug**
The new rule `curl-auth-user` is supposed to find leaks of basic auth credentials used with curl by determining if the curl command was called with the `-u` or `--user` flag, where the value contains a leaked secret (i.e. the credentials of an account).

However, the rule also matches if the flag is used in a script with variables. This is not a leak, as the environment variables are retrieved through other means and never committed to source code or any artifacts.

**To Reproduce**

Create a file `test.sh` (example AQL call to Artifactory):

```sh
curl -sSf -u ${USER}:${PASSWORD} \
    -H "content-type: text/plain" \
    -X POST 'https://xxx.com/artifactory/api/search/aql' \
    -d @query.aql \
    > output.json
```

Run gitleaks and inspect the finding.

> ...
Finding:     curl -sSf -u REDACTED \
Secret:      REDACTED
RuleID:      curl-auth-user
Entropy:     4.169890
...

If the quotes around the value passed using `-u` are removed, the false positive disappears. This means the rule must be fixed to not simply match if quotes are given.

**Expected behavior**
Variables do not match the rule, even if they are quoted, as they are a valid way to pass the credentials without leaks.
One wants to quote variables to prevent word-splitting and globbing.

**Screenshots**

 - OS: MacOs
 - Gitleaks Version: v8.21.1

**Additional context**
You can also verify the behavior with regex101:
https://regex101.com/r/LcccLw/1

cc @zricethezav

rgmz:
### Description:
This adds the following improvements to the CURL rule. It closes #1589.

1. `curl-auth-header` matches provider-specific headers like `X-AIO-Key` or `X-Octopus-ApiKey`.
2. `curl-auth-user` no longer matches placeholders where individual segments are quoted (e.g., `"${USR}":"${PWD}"`).

### Checklist:


zricethezav:
nice

PR #1611, opened 2024-10-29
Add freemius secret key rule

naveen17797:
### Description:
This PR adds the freemius secret key detection rule.

### Checklist:

### DIFF
--- a/cmd/generate/config/rules/freemius.go
+++ b/cmd/generate/config/rules/freemius.go
@@ -1,0 +1,47 @@
+package rules
+
+import (
+	"github.com/zricethezav/gitleaks/v8/cmd/generate/config/utils"
+	"github.com/zricethezav/gitleaks/v8/config"
+	"regexp"
+)
+
+func Freemius() *config.Rule {
+	// define rule
+	r := config.Rule{
+		RuleID:      "freemius-secret-key",
+		Description: "Detected a Freemius secret key, potentially exposing sensitive information.",
+		Regex:       regexp.MustCompile(`(?i)["']secret_key["']\s*=>\s*["'](sk_[\S]{29})["']`),
+		Keywords:    []string{"secret_key"},
+		Path:        regexp.MustCompile(`(?i)\.php$`),
+	}
+
+	// validate
+	tps := map[string]string{
+		"file.php": `$config = array(
+			"secret_key" => "sk_ubb4yN3mzqGR2x8#P7r5&@*xC$utE",
+		);`,
+	}
+	// It's only used in PHP SDK snippet.
+	// see https://freemius.com/help/documentation/wordpress-sdk/integrating-freemius-sdk/
+	fps := map[string]string{
+		// Invalid format: missing quotes around `secret_key`.
+		"foo.php": `$config = array(
+			secret_key => "sk_abcdefghijklmnopqrstuvwxyz123",
+		);`,
+		// Invalid format: missing quotes around the key value.
+		"bar.php": `$config = array(
+			"secret_key" => sk_abcdefghijklmnopqrstuvwxyz123,
+		);`,
+		// Invalid: different key name.
+		"baz.php": `$config = array(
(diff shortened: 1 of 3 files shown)


cmd/generate/config/rules/freemius.go
```diff
@@ -0,0 +1,44 @@
...
+	"regexp"
+)
+
+func Freemius() *config.Rule {
+	// define rule
+	r := config.Rule{
+		RuleID:      "freemius-secret-key",
+		Description: "Detected a Freemius secret key, potentially exposing sensitive information.",
+		Regex:       regexp.MustCompile(`(?i)["']secret_key["']\s*=>\s*["'](sk_[\S]{29})["']`),
+		Keywords:    []string{"secret_key", "sk_"},
```
Ben-grmbl, line 16: only one of the keywords needs to match, and if I understand the samples, both are always present, so probably just "secret_key" would do?

cmd/generate/config/rules/freemius.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/zricethezav/gitleaks/v8/config"
+	"regexp"
+)
+
+func Freemius() *config.Rule {
+	// define rule
+	r := config.Rule{
+		RuleID:      "freemius-secret-key",
+		Description: "Detected a Freemius secret key, potentially exposing sensitive information.",
+		Regex:       regexp.MustCompile(`(?i)["']secret_key["']\s*=>\s*["'](sk_[\S]{29})["']`),
```
Ben-grmbl, line 15: `[\S]{29}` seems a little too broad, this should probably be `[a-zA-Z0-9]{29}` or `\w{29}` ?

cmd/generate/config/rules/freemius.go
naveen17797: You're right, i removed the sk_ identifier

cmd/generate/config/rules/freemius.go
naveen17797: but \w{29} wouldn't match this key because it contains characters like `$,@`

`sk_ubb4yN3mzqGR2x8#P7r5&@*xC$utE` ( sample key from their repo https://github.com/Freemius/wordpress-sdk?tab=readme-ov-file#integrating--initializing-the-sdk)

naveen17797: i have also updated the tps with this example key

Ben-grmbl: I see, I was wondering why this key would currently not be covered by the "generic-api-key" rule. 

Any freemius key that only contains [\w.=-] should already be detected by the generic rule

e.g. ` "secret_key" => "sk_ubb4yN3mzqGR2x8sP7r5dfgxCfutE",` is correctly detected,
but anything containing `# & @ * $` will not be

Unfortunately I could not find any hint in the freemius codebase about whether  `# & @ * $` are the only possible special characters .


Perhaps, since this is a very specific php pattern, to avoid unforeseen false positives, would it make sense to limit this to *.php files?

```golang
r := config.Rule{
...
	Path:        regexp.MustCompile(`(?i)\.php$`),
```

naveen17797: > Perhaps, since this is a very specific php pattern, to avoid unforeseen false positives, would it make sense to limit this to *.php files?

You're right, this is only found in php file, i made the changes required.

> side note, the code example in the freemius documentation seems highly irresponsible:

They claim to remove this during deployment (https://freemius.com/help/documentation/wordpress-sdk/integrating-freemius-sdk/)

> Note: The secret key is automatically included in the SDK snippet for convenience. As per the comments in the snippet, this should be removed from production code, but don’t worry if you forget as the PHP processor removes it automatically in the Deployment mechanism

AFAIK they use a PHP parser that relies on a specific AST pattern, so there's always a chance for something to slip through if the snippet structure is changed or if the source code is distributed through another platform.

zricethezav:
LGTM, thanks for the review @Ben-grmbl

PR #1614, opened 2024-10-29
fix(vault-service-token): ensure that TPS contains digits

### LINKED ISSUE #1613: Slight regression in hashicorp true-positive examples
While testing the code for my recent pull request #1593 , I stumbled upon something. I might have introduced a slight regression.

The rule "vault-service-token" in  hashicorp.go rejects hashicorp keys, if the first 24 chars contain only letters, without digits:

```golang
	{
		Regexes: []*regexp.Regexp{
			// https://github.com/gitleaks/gitleaks/issues/1490#issuecomment-2334166357
			regexp.MustCompile(`s\.[A-Za-z]{24}`),
		},
	},
```

The linked comment says:
> If the tokens are randomly generated, I'd guess the chances of it only being letters is fairly low.
> 
> pow(24, 26 + 26) / pow(24, 26 + 26 + 10)
> 
> So if the math is right, a 0.000000000002% chance that `s\.[a-zA-Z]{24}\b` will miss a true positive if the keys are random.

However, this does not seem to be correct.
The chance of generating only letters, when picking from A-Z,a-z,0-9, are, I believe

`P = ( (26+26) / (26+26+10) ) ^ 24 = 1.46 %`

Unfortunately, since I changed the true positive generation to include uppercase letters, this chance has increased. Before that, with only lowercase letters, it was (26 / (26+10) ) ^ 24 = 0.04 % 

Since there are 2 TPS examples with randomly generated letters, I think that about 2.91% of `go generate` will currently fail. `1 - (1 - ((26+26) / (26+26+10)) ^ 24) ^ 2`

With the previous generation of only lowercase letters, this chance was lower. I thought the occasional failures were related to the entropy threshold, but actually by fixing the example secret to include uppercase, the problem got worse instead. Sorry about that, I'll try to fix it asap.

**To Reproduce**

Run `go generate` in a loop

**Expected behavior**

As far as I understand, this allowlist was neccessary to filter some real false positives. So I am not sure that can be tweaked, but I can fix the false positive generation

cc @zricethezav

Ben-grmbl:
closes #1613

### Description:
Ensures that the true positives for both the old and the new hashicorp token formats contain a digit in the first 24 characters.

Chances of randomly generating a 24 letter string without digits: P = ( (26+26) / (26+26+10) ) ^ 24 = 1.46 %

### Checklist:

### DIFF
--- a/cmd/generate/config/rules/hashicorp_vault.go
+++ b/cmd/generate/config/rules/hashicorp_vault.go
@@ -28,10 +28,10 @@ func VaultServiceToken() *config.Rule {
 	// validate
 	tps := []string{
 		// Old
-		utils.GenerateSampleSecret("vault", secrets.NewSecret(`s\.[a-zA-Z0-9]{24}`)),
+		utils.GenerateSampleSecret("vault", secrets.NewSecret(`s\.[0-9][a-zA-Z0-9]{23}`)),
 		`token: s.ZC9Ecf4M5g9o34Q6RkzGsj0z`,
 		// New
-		utils.GenerateSampleSecret("vault", secrets.NewSecret(`hvs\.[\w\-]{90}`)),
+		utils.GenerateSampleSecret("vault", secrets.NewSecret(`hvs\.[0-9][\w\-]{89}`)),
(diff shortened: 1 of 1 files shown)


zricethezav:
gah man, f github's conflict UI

PR #1616, opened 2024-10-30
Add option to include `Line` field in JSON format report

taiki45:
## Description
We propose adding an option to include the `Line` field in the JSON output of gitleaks.

Rather than simply adding a flag like `--include-line`, we suggest introducing a new `jsonextra` format that extends the standard `json` format. This approach ensures future extensibility, as we can easily add additional fields to `jsonextra` that are not part of the default output.

## Background
We are currently using gitleaks in conjunction with our custom "allowlist" extension: https://github.com/Finatext/gls

This extension supports multiple allow-rules for both global and per detection-rules. We developed this as separate tool because integrating such an extension into the upstream version of gitleaks seemed challenging without sufficient confidence.

We have been successfully running gitleaks with our extension on every GitHub PR in our organization for some time. While it may be difficult to incorporate our extension directly into the upstream version, we need to modify gitleaks to add the `Line` field to the JSON output to support this extension.

## Impact
No changes on the existing `json` format.

### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,7 +162,7 @@ Flags:
       --no-banner                     suppress banner
       --no-color                      turn off color for verbose output
       --redact uint[=100]             redact secrets from logs and stdout. To redact only parts of the secret just apply a percent value from 0..100. For example --redact=20 (default 100%)
-  -f, --report-format string          output format (json, csv, junit, sarif) (default "json")
+  -f, --report-format string          output format (json, jsonextra, csv, junit, sarif) (default "json")
   -r, --report-path string            report file
   -v, --verbose                       show verbose output from scan
       --version                       version for gitleaks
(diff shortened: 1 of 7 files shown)


rgmz:
I like the ability to include `Line` since `Match` doesn't always include the full context (e.g., #1585). Idk how I feel about creating a new output option — curious what @zricethezav thinks.

> This extension supports multiple allow-rules for both global and per detection-rules. We developed this as separate tool because integrating such an extension into the upstream version of gitleaks seemed challenging without sufficient confidence.

As an aside, [Gitleaks v8.21.0 introduced multiple per-rule allowlists](https://github.com/gitleaks/gitleaks?tab=readme-ov-file#configuration). I can see value in officially supporting multiple global allowlists as well.

taiki45:
I'm also in favor of implementing an option like `--include=Line` that would apply to all output formats, not just JSON.

>As an aside, Gitleaks v8.21.0 introduced multiple per-rule allowlists

That's great to hear. We'll look into migrating our extension to use this official feature.

zricethezav:
> Rather than simply adding a flag like --include-line, we suggest introducing a new jsonextra format that extends the standard json format. This approach ensures future extensibility, as we can easily add additional fields to jsonextra that are not part of the default output.

@taiki45 Thank you for considering cli flag bloat 🙇🏻. I have no problem including `jsonextra` as a report format option with the caveat that this should be considered an experimental feature. In the future we may add an option that gives the user complete control on what elements get reported. It could look something like https://github.com/brittonhayes/pillager?tab=readme-ov-file#custom-templates.

PR #1620, opened 2024-11-01
simplify regex for secretPrefix, secretSuffix

Ben-grmbl:
### Description:

This pull request should not change any functionality, it only simplifies the regex generated by GenerateSemiGenericRegex and GenerateUniqueTokenRegex a bit, in order to make them easier to understand, (and to make gitleaks.toml 4 KB smaller&nbsp;:-)

based on the sidenote in #1467

Changes:
```diff
// in identifierSuffix:
- (?:[\s|']|[\s|"]){0,3}
+ [\s'"|]{0,3}

// in secretPrefix
- (?:'|\"|\s|=|\x60){0,5}
+ [`'"\s=]{0,5}

// in secretSuffix
- (?:['|\"|\n|\r|\s|\x60|;]|$)
+ (?:[`'"\s;|]|$)
```

I am not quite certain whether `"|"` was intended to be included in the character ranges, or whether the intention was that `[a|b]` should match only a or b, like `[ab]`.

But since the previous regex for identifierSuffix and secretSuffix match for `"|"`, this patch does not change that.
 
I removed \r\n from the secretSuffix, since it is already included in \s

I also replaced `\x60` with `  
As far as I could tell, the backtick does not need to be escaped in a toml file, and it was only slightly inconvenient to escape it in code: 

```diff
- secretPrefix       = `(?:'|\"|\s|=|\x60){0,5}(`
+ secretPrefix       = "[`" + `'"\s=]{0,5}(`
```

However, I may have misunderstood the ramifications, and can of course undo that if desired.
 

I have also written some unit tests covering GenerateSemiGenericRegex and GenerateUniqueTokenRegex, in a separate pull request #1623 

### Checklist:

* [x] Does your PR pass tests?
* [x] Have you written new tests for your changes?
* [x] Have you lint your code locally prior to submission?

PR #1623, opened 2024-11-03
test cases for GenerateSemiGenericRegex, GenerateUniqueTokenRegex

Ben-grmbl:
### Description:

I created some test cases for utils.GenerateSemiGenericRegex() and utils.GenerateUniqueTokenRegex()

This was mainly done to improve my understanding of these functions, and to check whether the generated regex could be simplified, see #1620

It can also serve as documentation of the behaviour, and to make sure future changes to these methods are easier to verify.

(e.g. In a future issue I would like to suggest that `&` be added to the delimiters, to allow detecting secrets in the form `example.com?api_key=abc&otherparam=def`, and these tests help document what would change. )

I am not entirely certain whether the benefits of adding this test case  are large enough to warrant inclusion; if not, I can withdraw it.

### Checklist:

### DIFF
--- a/cmd/generate/config/utils/generate_test.go
+++ b/cmd/generate/config/utils/generate_test.go
@@ -1,0 +1,235 @@
+package utils
+
+import (
+	"testing"
+)
+
+func TestGenerateSemiGenericRegex(t *testing.T) {
+	tests := []struct {
+		name              string
+		identifiers       []string
+		secretRegex       string
+		isCaseInsensitive []bool
+		validStrings      []string
+		invalidStrings    []string
+	}{
+		{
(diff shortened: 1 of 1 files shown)


rgmz:
> (e.g. In a future issue I would like to suggest that `&` be added to the delimiters, to allow detecting secrets in the form `example.com?api_key=abc&otherparam=def`, and these tests help document what would change. )

This was initially the intention for #1484, however the regex changes were removed.

In addition to tests for the specific generate functions, I'd like to merge the `GenerateSampleSecrets` + property testing from #1499 to catch pattern-specific issues.

zricethezav:
This is very helpful. Thank you!

PR #1627, opened 2024-11-04
refactor: pre-commit hooks

adam-moss:
### Description:

Migrates the pre-commit hooks to the new syntax.

### Checklist:

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,16 +1,16 @@
 - id: gitleaks
   name: Detect hardcoded secrets
   description: Detect hardcoded secrets using Gitleaks
-  entry: gitleaks protect --verbose --redact --staged
+  entry: gitleaks git --pre-commit --redact --staged --verbose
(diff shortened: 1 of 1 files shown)


.pre-commit-hooks.yaml
```diff
@@ -1,16 +1,16 @@
 - id: gitleaks
   name: Detect hardcoded secrets
   description: Detect hardcoded secrets using Gitleaks
-  entry: gitleaks protect --verbose --redact --staged
```
zricethezav, line 4: @adam-moss thanks for the PR. I believe the `--verbose` and `--redact` flags should be present still.

adam-moss: I can add them for explicitness if you prefer, but `--redact` would be the same as `--redact=100` which is the default I think?

adam-moss: Updated @zricethezav

zricethezav (approved):

PR #1628, opened 2024-11-05
fix(easypost): only detect tokens of correct length

Ben-grmbl:
### Description:
The rules easypost-api-token and easypost-test-api-token were not checking for word boundaries at the end of token.

This caused some false positives: given enough BASE64 encoded text, the string "+EZAK" or "/EZAK" is bound to occur at some point, e.g.
"...6wqX6fNUXA/rYqRvfQ+**EZAK**GqQRiRyqAFRQshGPWOIAwNWGORfKHSBnVNFtVmWYoW6PH23lkqbbDWep95C/3VmWq/edti6..."

The new regex also checks for word boundaries at the end of the token. 

(This is not perfect, a `+` or `/` could still randomly occur 54 characters after a `+EZAK` in a base64 string, but now the chances are 64 times lower. 
Long-term, I would like to migrate this to utils.GenerateUniqueTokenRegexp(), so I did not want to complicate this too much. Just wanted to avoid something like #1612 happening again)

### Checklist:

### DIFF
--- a/cmd/generate/config/rules/easypost.go
+++ b/cmd/generate/config/rules/easypost.go
@@ -13,14 +13,22 @@ func EasyPost() *config.Rule {
 	r := config.Rule{
 		RuleID:      "easypost-api-token",
 		Description: "Identified an EasyPost API token, which could lead to unauthorized postal and shipment service access and data exposure.",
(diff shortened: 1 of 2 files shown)


zricethezav:
great catch

Issue #1557, opened 2024-10-10
Managing custom config file


cduggn:
Is there a defined pattern in the gitleaks community on how to manage custom rules?

The `--config` flag doesn't support remote download so there is an expectation that the rules file is already present in the local dir, or some path you provide, or it will fall back to the default. This requires a little more effort to make sure teammates are using the latest custom rules.

Extending the `--config` flag to support remote downloads is not efficient but would be interested to hear if it was considered? (Obvious reasons to not include this is It adds a runtime dependency, increases latency , and probably a redundant call the vast majority of executions.)

Potentially the CLI could include a new `--download` flag that could run independent of any scan and overwrite the local rules file if changes were detected. I don't see too many down sides with this and it saves having to maintain a separate script ( like the one I'm currently using included below) to periodically update the rules

```python

import requests
import sys
import os
import hashlib

def calculate_hash(file_path):
    b= hashlib.blake2b()
    with open(file_path, 'rb') as f:
        for chunk in iter(lambda: f.read(4096), b""):
            b.update(chunk)
    return b.hexdigest()

def download_file(url, local_path):
    try:
        response = requests.get(url)
        response.raise_for_status()

        remote_hash = hashlib.blake2b(response.content).hexdigest()

        if os.path.exists(local_path):
            local_hash = calculate_hash(local_path)
            if remote_hash == local_hash:
                print(f"No new version available. Local file is up to date.")
                return

        with open(local_path, 'wb') as file:
            file.write(response.content)
        print(f"Downloaded config file to {local_path}")
    except requests.RequestException as e:
        print(f"Failed to download config file: {e}")
        sys.exit(1)

if __name__ == "__main__":
    if len(sys.argv) < 2:
        print("Specify URL to the gitleaks.toml config")
        sys.exit(2)

    gitleaks_config_url = sys.argv[1]
    gitleaks_config_path = "gitleaks.toml"

    download_file(gitleaks_config_url, gitleaks_config_path)

```

cc @zricethezav

zricethezav:
I know @bplaxco has some great ideas on how to manage configs. Some of which are features that should be added to gitleaks.

> Extending the --config flag to support remote downloads is not efficient but would be interested to hear if it was considered? 

It has been considered and actually there is a placeholder function that needs to be filled in to support this feature. I just haven't gotten to it yet https://github.com/gitleaks/gitleaks/blob/9988e5224c05eac1cce8501c8951e67566bbf0a8/config/config.go#L239

[...]

PR #1630, opened 2024-11-08
cmd: adding config and fetch commands to load custom config from remote URL 

cduggn:
### Description:

Relates to [issue](https://github.com/gitleaks/gitleaks/issues/1557) 

The `--config` persistent flag enables users to provide a path or directory which `gitleaks` can use to load custom rules. This PR enhances that behaviour by adding a new `config` command with a `fetch` sub-command to load a custom config file from a remote location and write that file to disk. `gitleaks` can detect this file on subsequent scans following the config file chain of checks. 

```go
const configDescription = `config file path
order of precedence:
1. --config/-c
2. env var GITLEAKS_CONFIG
3. ~/.config/.gitleaks/config.toml
4. (target path)/.gitleaks.toml
If none of the four options are used, then gitleaks will use the default config`
```

### Detail

- New `config` command with `fetch` subcommand to download and write custom config to `~/.config/gitleaks/config.toml` in `cmd/config.go`
- `--url` flag for fetch command to specify remote URL
- updates config in `cmd/root.go` chain of checks to look for custom file in `~/.config/gitleaks/config.toml` 
- supports existing extend.Default and extend.Path functionality

### Usage
Fetch remote config

```shell
# Validates URL and downloads to ~/.config/gitleaks/config.toml
gitleaks config fetch --url https://mycustomconfig.com/config.toml
```

### Checklist:


manage/config.go
```diff
@@ -0,0 +1,126 @@
...
+	var (
+		resp []byte
+		err  error
+	)
+
+	resp, err = c.fetch(rawURL)
+	if err != nil {
+		return fmt.Errorf("could not download config from remote url: %w", err)
+	}
+
```
bplaxco, line 118: I think it'd be a good idea to validate the config before it's written. 

Here's an example of something similar to validate config:

https://github.com/leaktk/scanner/blob/d1372418a8dfd77fdff22d1e335a2eb956ef1651/pkg/scanner/patterns.go#L157-L191

The reference to the single allowlist is a bit stale in that example, but it helps guard against things like:

- Invalid TOML
- Empty responses
- etc

cmd/config.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/spf13/cobra"
+	"github.com/zricethezav/gitleaks/v8/manage"
+	"os"
+	"path/filepath"
+)
+
+func init() {
+	rootCmd.AddCommand(configCmd)
+	configCmd.AddCommand(configFetchCmd)
+	configFetchCmd.Flags().StringP("url", "u", "", "URL of remote configuration file.")
```
bplaxco, line 14: Not a gripe, just a question ^_^:

Why make it a flag instead of a positional argument if it's required?  (e.g. `gitleaks config fetch <url>`)

cmd/config.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/rs/zerolog/log"
+	"github.com/spf13/cobra"
+	"github.com/zricethezav/gitleaks/v8/manage"
+	"os"
+	"path/filepath"
+)
+
+func init() {
+	rootCmd.AddCommand(configCmd)
+	configCmd.AddCommand(configFetchCmd)
```
bplaxco, line 13: Thoughts on adding [gitleaks config reset](https://github.com/gitleaks/gitleaks/issues/1557#issuecomment-2408765315)?

All it would do is delete the config file if it exists:

With output looking something like:
```
$ gitleaks config reset
Config file at /home/username/.config/gitleaks/config.toml removed

$ gitleaks config reset
No config file at /home/username/.config/gitleaks/config.toml found
Nothing to do
```
or something similar.


Rationale:

- Helps a user reset their config if it gets wonky (gitleaks version upgrade borks something, etc)
- Keeps them from needing to remember where their config file is or even if they have one. They can just do this to make sure an environment is reset back to default

bplaxco (commented):
Added some thoughts and nitpicks here and there. Overall LGTM. I think the config reset cmd would be handy though.

I also noticed this didn't end up including any refresh features. Did you want that included or left out?

The main reason I could see it being worth including is for keeping the patterns fresh for people.

manage/config.go
```diff
@@ -0,0 +1,126 @@
...
+	return true
+}
+
+func (c *ConfigManager) fetch(rawURL string) ([]byte, error) {
+	if !isValidURL(rawURL) {
+		return nil, fmt.Errorf("invalid URL")
+	}
+
+	resp, err := http.Get(rawURL)
```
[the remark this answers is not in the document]
cduggn: @bplaxco retry is good addition, just requires a small change to retry anonymous function to return []byte instead of *http.Response

[backoff](https://github.com/cenkalti/backoff/blob/v5/retry.go) can achieve similar outcome if you are happy to forgo use of crypto/rand to generate jitter?

cmd/config.go
```diff
@@ -0,0 +1,57 @@
...
+		log.Fatal().Msg("URL flag is required")
+	}
+
+	// build file path
+	homeDir, err := os.UserHomeDir()
+	if err != nil {
+		log.Fatal().Err(err).Msg("unable to get user home directory")
+	}
+
+	targetPath := filepath.Join(homeDir, gitleaksHomeConfigRelPath)
```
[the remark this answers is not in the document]
cduggn: will include

cmd/config.go
cduggn: an oversight!

manage/config.go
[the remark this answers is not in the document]
bplaxco: Yeah something out of the box sounds good to me ^_^. That'd be more of a call for @zricethezav

rgmz:
Tangentially related to #1654, as they both introduce a 'config' subcommand.

cduggn:
Is there a reason that was added ahead of this change? And is this PR no longer required, thanks @rgmz

rgmz:
It isn't 'ahead', I'm simply linking them for posterity. 

I think there's still strong interest in this; @bplaxco's left some good comments, idk if they've been resolved. Only @zricethezav has merge rights.

bplaxco:
> @bplaxco's left some good comments, idk if they've been resolved

It looks like some have. I'll try go re-review ^_^ (Would have done it earlier but GitHub notifications weren't getting through for a bit)

cduggn:
PTAL

config/remote_config.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+
+	return result, fmt.Errorf("failed after %d attempts: %w", maxAttempts, err) // Wrap the final error.
+}
+
+func writeToDisk(content []byte, filePath string) (int, error) {
+	// create directory path if not already created
+	dirPath := filepath.Dir(filePath)
+	if err := os.MkdirAll(dirPath, 0700); err != nil {
+		return 0, fmt.Errorf("failed to create directory ~/.config/gitleaks : %w", err)
```
bplaxco, line 152:
```suggestion
		return 0, fmt.Errorf("failed to create directory %s : %w", filePath, err)
```

since it just happens to be ~/.config/gitleaks in the default case but could differ depending on the xdg env var.

config/remote_config.go
```diff
@@ -0,0 +1,189 @@
...
+			err = fmt.Errorf("failed to remove temporary file: %w (original error: %w)", removeErr, err)
+		} // Clean up the temp file in case of failure
+	}()
+
+	// Write to the temporary file
+	writer := bufio.NewWriter(tempFile)
+	bytesWritten, err := writer.Write(content)
+	if err != nil {
+		return 0, fmt.Errorf("failed to write to temporary file: %w", err)
+	}
```
bplaxco, line 172: nit: if you pass in the content as a io.Reader this could be a call to `io.Copy`. Only downside is you wouldn't get the empty file check.

config/remote_config.go
```diff
@@ -0,0 +1,189 @@
...
+	return &RemoteConfig{}
+}
+
+func (c *RemoteConfig) WriteTo(rawURL, targetPath string) error {
+	var (
+		resp []byte
+		err  error
+	)
+
+	resp, err = retry(context.Background(), 3, func(ctx context.Context) ([]byte, error) {
```
bplaxco, line 41:
```suggestion
func (c *RemoteConfig) WriteTo(ctx context.Context, rawURL, targetPath string) error {
	var (
		resp []byte
		err  error
	)

	resp, err = retry(ctx, 3, func(ctx context.Context) ([]byte, error) {
```

Lets the caller define the timeout and such :ok_hand:

bplaxco (changes requested):
Looking pretty good. Just a few minor things :ok_hand:

PR #1632, opened 2024-11-11
Fix empty capture groups being used as secret

### LINKED ISSUE #1631: Regex works doesn't correct
**Describe the bug**
Gitleaks does not find all the secrets, although the regex is correct and everything is ok here + field "Secret" is not filled: 
https://regex101.com/r/zAaY2U/1
https://regex101.com/r/oSNiUu/1

**To Reproduce**
1. Add my private-rule like in url above: 
(?i)-----BEGIN[ A-Z0-9_-]{0,100}PRIVATE KEY[a-z ]{0,6}-----[\s\S-]{20,}?KEY[a-z ]{0,6}-----
2. Generate private keys or download my file
3. Run gitleaks latest on this file with my rule:
gitleaks detect -c current.toml -s keys.txt --no-git -v
**Expected behavior**
all 10 secrets are found + field "Secret" is filled

**Screenshots**
изображение
изображение

[keys.txt](https://github.com/user-attachments/files/17699029/keys.txt)

 - OS: MacOS, Linux
 - Gitleaks Version:
 8.21.0

**Additional context**
I noticed that problem started since 8.18.1 
cc @zricethezav

rgmz:
### Description:
This fixes #1631.

### Checklist:

### DIFF
--- a/detect/detect.go
+++ b/detect/detect.go
@@ -397,16 +397,10 @@ MatchLoop:
 				finding.Secret = groups[rule.SecretGroup]
 			} else {
 				// If |secretGroup| is not set, we will use the first suitable capture group.
-				if len(groups) == 2 {
-					// Use the only group.
-					finding.Secret = groups[1]
-				} else {
-					// Use the first non-empty group.
-					for _, s := range groups[1:] {
-						if len(s) > 0 {
-							finding.Secret = s
-							break
-						}
+				for _, s := range groups[1:] {
+					if len(s) > 0 {
+						finding.Secret = s
+						break
 					}
 				}
 			}


detect/detect.go
```diff
@@ -398,8 +398,10 @@ MatchLoop:
...
 				if len(groups) == 2 {
-					// Use the only group.
-					finding.Secret = groups[1]
+					if s := groups[1]; s != "" {
+						// Use the only group.
+						finding.Secret = s
+					}
 				} else {
 					// Use the first non-empty group.
 					for _, s := range groups[1:] {
```
Ben-grmbl, line 407: This is a good change and should fix the empty secret, but unless I am mistaken, now both branches of the `if len(groups)==2` do the same thing?

So this should be equivalent
```patch
diff --git a/detect/detect.go b/detect/detect.go
index ac72498..8e9131e 100644
--- a/detect/detect.go
+++ b/detect/detect.go
@@ -396,22 +396,15 @@ MatchLoop:
 				}
 				finding.Secret = groups[rule.SecretGroup]
 			} else {
 				// If |secretGroup| is not set, we will use the first suitable capture group.
-				if len(groups) == 2 {
-					if s := groups[1]; s != "" {
-						// Use the only group.
-						finding.Secret = s
-					}
-				} else {
 					// Use the first non-empty group.
 					for _, s := range groups[1:] {
 						if len(s) > 0 {
 							finding.Secret = s
 							break
 						}
 					}
-				}
 			}
 		}
 
 		// check if the regexTarget is defined in the allowlist "regexes" entry

```

rgmz: Hah, good point.

PR #1633, opened 2024-11-12
Ensure keywords are downcased

craigmarksmith:
### Description:

This PR addresses a bug where rules with uppercase keywords can not be found when those rules are defined in a TOML that is extended. 

When the TOML is consumed into the Config object, the keywords for each rule in the TOML are converted to lowercase and then added to a summary of keywords on the Config object. https://github.com/gitleaks/gitleaks/blob/48a2e0e3184d8d07275a22f24d3c96019df8cd4c/config/config.go#L97

Then, during the [detect phase](https://github.com/gitleaks/gitleaks/blob/48a2e0e3184d8d07275a22f24d3c96019df8cd4c/detect/detect.go#L196), these lowercase keywords are used to detect secrets in the the raw fragment which has also been converted to lowercase.

This works as intended for simple configurations, but when one configuration extends from another, the keywords of the base configuration are not converted to lowercase. 

Take for example the configuration

```toml
title = "Extend Config"

[extend]
path="base.toml"

[[rules]]
    id = "aws-secret-key-again-again"
    description = "A new description"
```

```toml
# base.toml
title = "Base Configuration"

[[rules]]
    id = "aws-secret-key-again-again"
    description = "AWS Secret Key"
    regex = '''(?i)aws_(.{0,20})?=?.[\'\"0-9a-zA-Z\/+]{40}'''
    tags = ["key", "AWS"]
    keywords = ["AWS"]
[[rules.allowlists]]
    stopwords = ["fake"]
```

Once parsed, the above configuration results in the keyword `AWS`, rather than `aws` which is required for detect to work as expected.

### Related issue:

https://github.com/gitleaks/gitleaks/issues/1634

### Checklist:


config/config.go
```diff
@@ -314,10 +316,8 @@ func (c *Config) extend(extensionConfig Config) {
 			}
 			// The keywords from the base rule and the extended rule must be merged into the global keywords list
 			for _, k := range baseRule.Keywords {
```
craigmarksmith, line 318: This code is covered by https://github.com/gitleaks/gitleaks/pull/1633/files#diff-373f5578b8f362a364b50b0d01919a57da64c128547ca10d49df8ae422237c82R405

config/config.go
```diff
@@ -287,7 +288,8 @@ func (c *Config) extend(extensionConfig Config) {
 			// Rule doesn't exist, add it to the config.
 			c.Rules[ruleID] = baseRule
 			for _, k := range baseRule.Keywords {
-				c.Keywords[k] = struct{}{}
+				// Craig test this - extend_with_new_rule
+				c.Keywords[strings.ToLower(k)] = struct{}{}
```
craigmarksmith, line 292: This change is covered by https://github.com/gitleaks/gitleaks/pull/1633/files#diff-373f5578b8f362a364b50b0d01919a57da64c128547ca10d49df8ae422237c82R410

config/config.go
```diff
@@ -314,10 +314,7 @@ func (c *Config) extend(extensionConfig Config) {
 			}
 			// The keywords from the base rule and the extended rule must be merged into the global keywords list
 			for _, k := range baseRule.Keywords {
-				c.Keywords[k] = struct{}{}
-			}
-			for _, k := range currentRule.Keywords {
```
craigmarksmith, line 319: The `baseRule.Keywords` includes both the `baseRule.Keywords` and `currentRule.Keywords` [1](https://github.com/gitleaks/gitleaks/pull/1633/files#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17R311), so when assigning the rule keywords to `config.Keywords`, we only need to iterate over the base keywords.

That means these two `for` loops essentially do the same thing, so I've removed the second loop over `currentRule.Keywords`.

craigmarksmith:
Hey @zricethezav 👋 

I found https://github.com/gitleaks/gitleaks/issues/1634 and have created this PR to address the issue. Would you be able to take a look?

zricethezav:
> I think the bug is actually here. The global keyword map (used to populate ahocorasick) is lowercased, however, the rule keywords remain unchanged.

Sorry for the delay in review. @craigmarksmith wdyt about @rgmz's solution? Forcing the _rule_ keyword to lowercase in one spot seems more maintainable (in theory).

craigmarksmith:
@rgmz @zricethezav thank you so much for this. I've updated this PR with @rgmz suggestion and the tests I've added continue to pass! This solution is definitely more maintainable. 
Would you take another look.

zricethezav:
Thanks @craigmarksmith!

Issue #1312, opened 2023-12-09
adding support for writing formatted reports to stdout when --report-path = "" and --report-format is != ""


jfgsilva:
**Is your feature request related to a problem? Please describe.**
yes: please check #1021. When using the containerized solution, which is awesome btw, you can't write the report unless you change fs permissions before.

**Describe the solution you'd like**
It would be great if by specifying a report format without report path, normal output is silenced (all fmt.Println...) and replaced by the written report so we can then redirect it into a file, thus having no need to change file permissions.
By doing so, it would also allow to put these commands and shell scripts and loop it through different repositories.

by running something like
```bash
docker run --rm --network none \
-v $PWD:/path:ro \
zricethezav/gitleaks:latest detect --report-format json > $HOME/findings.json
``

**Describe alternatives you've considered**
Debbuging the code and alpine containers to see where the file permission error is.

**Additional context**
n/a

cc @zricethezav

**Comment by ryboe** [comment]
I would also like the ability to print to stdout. I'm currently doing this to search for secrets in files, but it's not the same as true `stdout` because I can't pipe the output anywhere.

```sh
gitleaks detect --no-git --source=$HOME/.zsh_history --report-path=/dev/stdout 2>/dev/null
```

PR #1642, opened 2024-11-17
feat: use STDOUT when report file not specified

bthuilot:
### Description:

Addresses #1312 to write report to `/dev/stdout` when `report-path` is set to `""` or `"-"`.

### Checklist:


bthuilot:
@rgmz I updated it so now only `-` specifies `STDOUT`, since I think that is the most intuitive for an end user but happy to change if you think only when `report-format` is set

zricethezav:
@bthuilot great work! I like this feature a lot

PR #1646, opened 2024-11-19
test: ci token permissions

bthuilot:
### Description:

check permissions of token


bthuilot:
was checking permissions of the token, i was under the impression it still had access to upstream repsoitory. I was luckily wrong

PR #1650, opened 2024-11-22
Allow user-defined report templates

rgmz:
### Description:
This is a follow-up to #1616. It allows users to customize the Gitleaks report format using [templates](https://www.digitalocean.com/community/tutorials/how-to-use-templates-in-go).

For instance, the change required by that PR could now be done with:
```sh
./gitleaks dir /tmp/ --config config/gitleaks.toml --report-path "report.json" --report-format template --report-template testdata/report/jsonextra.tmpl
```

**Prior art:** This is _heavily inspired_ by https://github.com/brittonhayes/pillager.

### Checklist:


cmd/generate/config/rules/1password.go
rgmz, line 1: There is a lot of noise due to the fact I moved `config/Rule.go` -> `config/rule/Rule.go` to prevent circular imports.

cmd/root.go
```diff
@@ -163,9 +167,78 @@ func Config(cmd *cobra.Command) config.Config {
...
+	reportPath := getReportPath()
+	if reportPath != "" {
+		// TODO: improve this
+		//reportDir := filepath.Dir(reportPath)
+		//if d, err := os.OpenFile(reportDir, os.O_RDWR, 0); err != nil {
+		//	defer func() {
+		//		_ = d.Close()
+		//	}()
+		//	log.Fatal().Err(err).Str("report-path", reportPath).Msg("Invalid report path")
+		//}
```
rgmz, line 180: Ideally, we'd validate the report path at startup.

cmd/root.go
```diff
@@ -163,9 +167,78 @@ func Config(cmd *cobra.Command) config.Config {
...
+		}
+
+		cfg.ReportPath = reportPath
+		cfg.Reporter = reporter
+	}
+
 	return cfg
 }
 
+func getReportPath() string {
```
rgmz, line 218: I don't know if these are necessary. In what conditions can `rootCmd.Flags().GetX(...)` fail?

report/report.go
```diff
@@ -13,24 +10,6 @@ const (
...
-		err = writeCsv(findings, file)
-	case ".xml", "junit":
-		err = writeJunit(findings, file)
-	case ".sarif", "sarif":
-		err = writeSarif(cfg, findings, file)
-	}
-
-	return err
+type Reporter interface {
+	Write(w io.WriteCloser, findings []Finding) error
```
rgmz, line 14: Pillager uses a common interface, rather than having disparate format-specific functions. I think this makes a lot of sense.

report/template.go
```diff
@@ -0,0 +1,50 @@
...
+
+	file, err := os.ReadFile(templatePath)
+	if err != nil {
+		return nil, fmt.Errorf("error reading file: %w", err)
+	}
+	templateText := string(file)
+
+	// TODO: Add helper functions like escaping for JSON, XML, etc.
+	t := template.New("custom")
+	t = t.Funcs(sprig.TxtFuncMap())
```
rgmz, line 31: This library provides numerous useful functions.
https://github.com/Masterminds/sprig?tab=readme-ov-file

config/config.go
```diff
@@ -65,12 +68,16 @@ type Config struct {
...
+	Rules       map[string]rule.Rule
+	Allowlist   rule.Allowlist
 	Keywords    map[string]struct{}
 
 	// used to keep sarif results consistent
 	OrderedRules []string
+
+	// report-related settings.
+	ReportPath string
+	Reporter   report.Reporter
```
zricethezav, line 80: @rgmz was this the line causing the circular dependency? Couldn't the reporter be added to the `Detector` struct like any other cli flag? Or hell, better yet, keep reporting compare logic in the `cmd` package. I'm not a fan of moving `allowlist.go` into a `rule` package since allowlists can be globally set as well.

rgmz: I believe the issue was that one of the reporters expected a reference to Config. Maybe it'd work under detector — I'll check.

rgmz: Okay, I've moved the reporter to the `Detector` struct.

report/json.go
```diff
@@ -5,22 +5,15 @@ import (
...
-	for i := range findings {
-		// Remove `Line` from JSON output
-		findings[i].Line = ""
-	}
-	return writeJsonExtra(findings, w)
+type JsonReporter struct {
 }
 
-func writeJsonExtra(findings []Finding, w io.WriteCloser) error {
+var _ Reporter = (*JsonReporter)(nil)
```
zricethezav, line 11: @rgmz is this injection necessary?

rgmz: This provides a compile-time check that `JsonReporter` satisfies `Reporter`, right? Or are you referring to something different?

zricethezav: nah, that's what I'm referring to. It's not necessary but I can see the value it provides

zricethezav:
@rgmz LGTM, last thing that would be useful is an update in the README. We probably should add a section on `reporting`

rgmz:
Absolutely. I've updated the README to include usage information, LMK your thoughts.

rgmz:
BTW @taiki45, I've used #1616 as an example use case.

Would this (specifying a custom template file) work for you, or would removing `jsonextra` as an option for `--report-format` be disruptive?

zricethezav:
> Would this (specifying a custom template file) work for you, or would removing jsonextra as an option for --report-format be disruptive?

imo let's leave `jsonextra` in place. Doing so avoids breakage in case anyone else uses `jsonextra`. A little bit of redundancy never hurt anyone (I think).

taiki45:
@rgmz Thank you so much for your work on this!

Specifying a custom template file works perfectly for us, and removing the `jsonextra` option for `--report-format` wouldn't cause any issues on our side. +1

sailro:
I don't know if its related to this change but starting with v8.21.3, gitleaks is not creating an empty csv file anymore when no secrets are found. (previously an empty file was created).

So not a big change but was enough to break my script :)

mikelolasagasti:
Is there a reason to use [sprig 2.22.0 v2](https://github.com/Masterminds/sprig/releases/tag/v2.22.0) from 2019 instead maintained `v3`?

https://github.com/Masterminds/sprig/releases

rgmz:
> So not a big change but was enough to break my script :)

Whoops. Looks like #1673 has already been created to address that.

> Is there a reason to use [sprig 2.22.0 v2](https://github.com/Masterminds/sprig/releases/tag/v2.22.0) from 2019 instead maintained v3?

Seems like an oversight on my part. [The latest version in `pkg.go.dev/github.com/Masterminds/sprig` is `v2.22.0`](https://pkg.go.dev/github.com/Masterminds/sprig), you have to specify `v3` separately.

Issue #1657, opened 2024-11-25
Do entropy check first


almaz045:
**Is your feature request related to a problem? Please describe.**
When I was trying to debug the rules, I encountered that the find was ignored due to low entropy, but -l trace didn't have this information. I decided to add it. Additionally, it seemed more appropriate to me to check for entropy first, and only then for regexallow. Maybe I'm wrong.

**Describe the solution you'd like**
I would like to raise the entropy check above the other checks and add logging to the trace

**Describe alternatives you've considered**
If I was wrong about the entropy order, then at least logging on it should be left in the trace

**Additional context**
I apologize in advance, as I have no experience on GO

cc @zricethezav

Ben-grmbl:
A trace output when a finding is rejected due to low entropy totally makes sense, I stumbled upon that problem a few times.

Changing the order does not really make sense in my opinion. If a finding is rejected due to matching the allowlist, its entropy does not matter.

for example: 
- api_key = "aFw#1Gr6nd3%2SDe2"  -> valid finding with high entropy
- monkey_sound = "OOK-OOK-OOK-OOK" -> invalid finding, but the fact that it has low entropy should not matter, More important is that it starts with "monkey", instead of just "key". Even if the monkey said something more complicated, it would remain an invalid finding

rgmz:
> If a finding is rejected due to matching the allowlist, its entropy does not matter.

The idea is that checking the entropy is faster than iterating over the allowlist conditions (regexes and string.Contains). 

The two conditions are mostly independent, however, if something is going to fail the entropy check then there's no point running it through the allowlists first.

Ben-grmbl:
Valid, if that affects the performance in a significant way, then that is a more important consideration than something that is only barely perceivable when analyzing the logs.

almaz045:
I ran 5 scans in one of the directories locally with regular gitleaks and after raising the entropy check above the regex check.
The average result is next.
Regular scan: 91.38 seconds
After raising the entropy check: 89.86 seconds
Perhaps such a test does not reflect all the nuances, but still, the speed becomes 1.5-2% faster

PR #1658, opened 2024-11-25
I do an entropy check earlier than the others and add the info to the…

almaz045:
… trace

Maybe I'm wrong, but it seems to me that if the regex allowlist is large, it's easier to first check for entropy, since we won't have to go through all these regexes. Plus, we need to check entropy anyway. Additionally, when using -l trace, there was no information that the find was missed precisely because of entropy. Now it will be shown

### Description:

### Checklist:

#1657

### DIFF
--- a/detect/detect.go
+++ b/detect/detect.go
@@ -412,6 +412,33 @@ MatchLoop:
 			}
 		}
 
+		// check entropy
+		entropy := shannonEntropy(finding.Secret)
+
+		finding.Entropy = float32(entropy)
+		if rule.Entropy != 0.0 {
+			if entropy <= rule.Entropy {				
+				logger.Trace().
+					//Here you can add REDACTED finding or finding.line or leave as is (at your discretion)				
+					Float32("entropy", finding.Entropy).
+					Str("rule-id", rule.RuleID).
+					Msg("Skipping finding due to low entropy")
+				// entropy is too low, skip this finding
+				continue
+			}
+			// NOTE: this is a goofy hack to get around the fact there golang's regex engine
+			// does not support positive lookaheads. Ideally we would want to add a
+			// restriction on generic rules regex that requires the secret match group
+			// contains both numbers and alphabetical characters, not just alphabetical characters.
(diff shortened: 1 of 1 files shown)


detect/detect.go
```diff
@@ -412,6 +412,33 @@ MatchLoop:
...
+		// check entropy
+		entropy := shannonEntropy(finding.Secret)
+
+		finding.Entropy = float32(entropy)
+		if rule.Entropy != 0.0 {
+			if entropy <= rule.Entropy {				
+				logger.Trace().
+					//Here you can add REDACTED finding or finding.line or leave as is (at your discretion)				
+					Float32("entropy", finding.Entropy).
+					Str("rule-id", rule.RuleID).
```
rgmz, line 424: This is already included in the rule's context.
```suggestion
					Str("rule-id", rule.RuleID).
```

detect/detect.go
```diff
@@ -412,6 +412,33 @@ MatchLoop:
 			}
 		}
 
+		// check entropy
+		entropy := shannonEntropy(finding.Secret)
+
```
rgmz, line 417: Style nitpick.
```suggestion
```

rgmz (commented):
This makes sense to me.

The CI is failing I guess because of inconsistent tabs/spaces?
```
-			if entropy <= rule.Entropy {				
+			if entropy <= rule.Entropy {
 				logger.Trace().
-					//Here you can add REDACTED finding or finding.line or leave as is (at your discretion)				
+					//Here you can add REDACTED finding or finding.line or leave as is (at your discretion)
 					Float32("entropy", finding.Entropy).
 					Str("rule-id", rule.RuleID).
 					Msg("Skipping finding due to low entropy")
@@ -507,7 +507,7 @@ MatchLoop:
 					Msg("Skipping finding due to rule allowlist")
 				continue MatchLoop
 			}
-		}		
+		}
```

almaz045:
@rgmz Should I close this PR if I created new with success tests?

PR #1663, opened 2024-12-09
feat: add settlemint api keys detection

Kemosabert:
### Description:

This PR adds detection for SettleMint API tokens

### Checklist:


cmd/generate/config/rules/settlemint.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/zricethezav/gitleaks/v8/cmd/generate/secrets"
+	"github.com/zricethezav/gitleaks/v8/config"
+)
+
+func SettlemintPersonalAccessToken() *config.Rule {
+	// define rule
+	r := config.Rule{
+		Description: "Found a Settlemint Personal Access Token.",
+		RuleID:      "settlemint-personal-access-token",
+		Regex:       utils.GenerateUniqueTokenRegex(`(sm_pat_)[a-zA-Z0-9]+`, false),
```
rgmz, line 14: Do the tokens have a fixed or minimum length? This will match `sm_pat_a` which likely isn't a real token.

cmd/generate/config/rules/settlemint.go
```diff
@@ -0,0 +1,58 @@
...
+		Keywords: []string{
```
rgmz, line 15: I'd recommend adding entropy to each rule. 3 is usually a good baseline but YMMV.
```suggestion
                Entropy: 3,
		Keywords: []string{
```

cmd/generate/config/rules/settlemint.go
Kemosabert: ah yes, looks like the tokens I was able to generate had a length of 12 characters after the prefix. Added this. 👍

cmd/generate/config/rules/settlemint.go
Kemosabert: Ok, was not sure, thanks 👍

cmd/generate/config/rules/settlemint.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/zricethezav/gitleaks/v8/cmd/generate/secrets"
+	"github.com/zricethezav/gitleaks/v8/config"
+)
+
+func SettlemintPersonalAccessToken() *config.Rule {
+	// define rule
+	r := config.Rule{
+		Description: "Found a Settlemint Personal Access Token.",
+		RuleID:      "settlemint-personal-access-token",
+		Regex:       utils.GenerateUniqueTokenRegex(`(sm_pat_)[a-zA-Z0-9]{12,}`, false),
```
zricethezav, line 14: @Kemosabert in an earlier comment you said secrets have 12 characters after `sm_pat_`. Only 12? If so can we limit the `[a-zA-Z0-9]` range to `{12}` instead of `{12,}`

Kemosabert: I meant 12 characters after the prefix, sorry for the confusion. I'll adjust the regexes 👍

zricethezav:
lgtm, thanks @Kemosabert!

PR #1652, opened 2024-11-22
Increase chunk size 10kb -> 100kb

rgmz:
### Description:
This is a band-aid solution to #1651.

### Checklist:

* [x] Does your PR pass tests?
* [x] Have you written new tests for your changes?
* [x] Have you lint your code locally prior to submission?

PR #1665, opened 2024-12-11
A better chunking algorithm?

rgmz:
### Description:
This is a follow-up to #1652.

It attempts to 'intelligently' chunk inputs on subsequent newlines `\n\n`, rather than indiscriminately breaking text apart. The assumption is that it's safe(r) for a chunk to end at (1), rather than (2) or (3). Of course, the buffer cannot grow indefinitely or the original issue would regress; it will only search for consecutive newlines up to a `maxPeekSize`.

```
  // TODO: optimization could be introduced here
  if mimetype, err := filetype.Match(buf[:n]); err != nil {
    return nil
  } else if mimetype.MIME.Type == "application" {
    return nil // skip binary files
  }

>>> (1) <<<  // If the chunk doesn't end in a newli>>> (2) <<<ne, peek|maxPeekSize| until we find one.
  // This hopefully avoids splitting
>>> (3) <<< // See: https://github.com/gitleaks/gitleaks/issues/1651
  var (
```

The original issue with 60 private keys is solved by this change:
```sh
$ ./gitleaks dir /tmp/60keys.txt --no-color

10:22PM INF scan completed in 88ms
10:22PM WRN leaks found: 60
```

### Thoughts

- Needs to be tested on Windows or files with CRLF.
- Reading one byte at a time might be slow (idk, seems pretty fast)
- It might make sense for whitespace not to reset the counter. That way empty lines can match the criteria.
  ```
    }
     
  // Foo
  ```

### Checklist:

### DIFF
--- a/detect/directory.go
+++ b/detect/directory.go
@@ -1,15 +1,19 @@
 package detect
 
 import (
+	"bytes"
 	"io"
 	"os"
 	"strings"
 
 	"github.com/h2non/filetype"
 	"github.com/rs/zerolog/log"
+
 	"github.com/zricethezav/gitleaks/v8/report"
 	"github.com/zricethezav/gitleaks/v8/sources"
 )
+
+const maxPeekSize = 25 * 1_000 // 10kb
 
 func (d *Detector) DetectFiles(paths <-chan sources.ScanTarget) ([]report.Finding, error) {
 	for pa := range paths {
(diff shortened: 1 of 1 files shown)


zricethezav:
@rgmz 
> Needs to be tested on Windows or files with CRLF.

Ideally I would have a windows machine to test on. We could just merge without testing on windows and if an issue arrises patch it real quick.

> It might make sense for whitespace not to reset the counter. That way empty lines can match the criteria.

Could you elaborate on this a little bit?

rgmz:
> Could you elaborate on this a little bit?

It's common to have blank lines only consisting of whitespace.
```
Foo\n
<tab>\n
Bar
```

The current logic resets the counter if `\n` isn't followed by another `\n`. In my mind, the above is as good of a place for the chunk to end as an empty line.
``` 
Foo\n
\n
Bar
```

zricethezav:
Gotcha, I agree with you that the condition for the counter should be expanded. It doesn't have to be in this PR, we can always follow up.

zricethezav (approved):

PR #1669, opened 2024-12-19
[feat] - replace std library regex engine with go-re2

ahrav:
### Description:
This PR replaces the standard library regex with [go-re2](https://github.com/wasilibs/go-re2) in all instances except for [newLineRegex](https://github.com/gitleaks/gitleaks/blob/a9e6d8c9a73c848e0e0dcbbc62eed0ebe851be3f/detect/detect.go#L28). The standard library regex engine is more performant than go-re2 for small regex patterns.

### Profile
Screenshot 2024-12-19 at 10 45 12 AM

### Benchmarks
1. `git log -p` output for this repo. [https://github.com/gitleaks/gitleaks]

Before ~58s:
Screenshot 2024-12-19 at 10 37 02 AM

After ~9s:
Screenshot 2024-12-19 at 11 16 30 AM

2. gitlab repo [https://gitlab.com/gitlab-org/gitlab]

Before ~49s:
Screenshot 2024-12-19 at 11 26 41 AM

After ~14s:
Screenshot 2024-12-19 at 11 30 07 AM

3. linux repo [https://github.com/torvalds/linux]

Before ~1m52s:
Screenshot 2024-12-19 at 11 28 54 AM

After ~1m:
Screenshot 2024-12-19 at 11 31 45 AM

### Checklist:


rgmz:
Is there a memory profile available? When TruffleHog switched, a few people complained about the large jump in memory usage.

ahrav:
> Is there a memory profile available? When TruffleHog switched, a few people complained about the large jump in memory usage.

I shared one with Zach but can add it here as well. I agree—the cost is noticeable, and the tradeoff might not be worth it. From what I understand, the WASM runtime module has an initial cost of 4MB, which increases with each rule, though the per-rule growth is significantly smaller.

I mentioned this to Zach, but feel free to close if the tradeoff isn't worth it.

rgmz:
It's a shame that Go doesn't have `#if` directives. Then you could easily toggle between both.

zricethezav:
christ almighty these numbers are impressive

ahrav:
> It's a shame that Go doesn't have `#if` directives. Then you could easily toggle between both.

I was reading some C code earlier and thought the exact same thing... 😢

zricethezav:
letterip

rgmz:
> It's a shame that Go doesn't have `#if` directives. Then you could easily toggle between both.

Seems it's technically possible: #1706

Issue #1432, opened 2024-06-27
Gitleaks detection is inconsistent in large files


adamcohen2:
**Describe the bug**

Gitleaks detection is inconsistent in large files. It doesn't find a password on line N, but is able to find the password on line N-1.

**To Reproduce**

1. Run `golang:1.19` Docker image:

   ```shell
   docker run -it --rm golang:1.19 bash
   ```

1. Clone and build Gitleaks:

   ```shell
   git clone https://github.com/gitleaks/gitleaks.git && cd gitleaks
   git checkout v8.18.4
   make build
   ```
1. Create `gitleaks.toml` with a single pattern:

   ```shell
   cat << HEREDOC > gitleaks.toml
   title = "gitleaks config"
   
   [[rules]]
   id = "test"
   description = "test"
   regex = '''\bpassword-[0-9a-zA-Z]{20}\b'''
   HEREDOC
   ```

1. Generate `100MB` of random data and save to a file called `random.txt`:

   ```shell
   python3 <<HEREDOC
   import random
   import string
   
   # Set the seed for reproducibility
   seed_value = 12345
   random.seed(seed_value)
   
   # Define the file size (100 MB)
   file_size = 100 * 1024 * 1024  # 100 megabytes in bytes
   
   # Define the characters to use (ASCII characters including punctuation)
   characters = string.ascii_letters + string.digits + string.punctuation + '\n'
   
   # Open the file in write mode
   with open("random.txt", "w") as f:
       # Write random characters to the file
       for _ in range(file_size):
           f.write(random.choice(characters))
   HEREDOC
   ```

1. Output `md5sum` of `random.txt` for debugging purposes:
   
   ```shell
   md5sum random.txt
   0172d80bbb74fe774fdd90197ef9f622  random.txt
   ```

1. Make a backup of `random.txt`:

   ```shell
   cp random.txt random.txt.bak
   ```

1. Insert a password matching the regex from the `gitleaks.toml` into any line greater than or equal to `62190` of `random.txt`:

   ```shell
   sed -i '62190 i password-0Vf5HAxLrkAQiWWJEWKM' random.txt
   ```

1. Run `gitleaks` on `random.txt`, passing `max-target-megabytes=1000`, and notice that password _is not detected_:

   ```shell
   ./gitleaks detect --source random.txt --config ./gitleaks.toml --no-git --max-target-megabytes=1000

   5:05AM INF scan completed in 187ms
   5:05AM INF no leaks found   
   ```

[...]

PR #1676, opened 2024-12-22
Refactor chunking algorithm changes

rgmz:
### Description:
This is a follow-up to #1665. 

Changes include:
- Moving the logic into a separate function and adding test cases
- Also implementing the changes in `reader.go`.
- Using `bufio.Reader` in `directory.go` to speed up execution

### Checklist:

### DIFF
--- a/detect/directory.go
+++ b/detect/directory.go
@@ -1,6 +1,7 @@
 package detect
 
 import (
+	"bufio"
 	"bytes"
 	"io"
 	"os"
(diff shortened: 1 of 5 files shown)


detect/directory.go
```diff
@@ -49,64 +50,33 @@ func (d *Detector) DetectFiles(paths <-chan sources.ScanTarget) ([]report.Findin
 				}
 			}
 
-			// Buffer to hold file chunks
-			buf := make([]byte, chunkSize)
-			totalLines := 0
+			var (
+				// Buffer to hold file chunks
+				reader     = bufio.NewReader(f)
```
rgmz, line 55: Using #1432 as a test case. @ahrav @zricethezav any obvious errors with my use of `bufio.Reader`?

**Before**
```
12:33PM INF scanned ~104857630 bytes (104.86 MB) in 5.63s
12:33PM WRN leaks found: 1
```
**After**
```
1:05PM INF scanned ~104857630 bytes (104.86 MB) in 1.42s
1:05PM WRN leaks found: 1
```

detect/directory.go
```diff
@@ -49,64 +50,32 @@ func (d *Detector) DetectFiles(paths <-chan sources.ScanTarget) ([]report.Findin
...
 
-						// Read additional data into a temporary buffer
-						m, readErr := f.Read(tempBuf)
-						if m > 0 {
-							peekBuf.Write(tempBuf[:m])
+				// "Callers should always process the n > 0 bytes returned before considering the error err."
+				// https://pkg.go.dev/io#Reader
+				if n > 0 {
+					// Only check the filetype at the start of file.
+					if totalLines == 0 {
```
rgmz, line 66: This doesn't need to be run after the first chunk.

detect/directory.go
```diff
@@ -145,3 +114,49 @@ func (d *Detector) DetectFiles(paths <-chan sources.ScanTarget) ([]report.Findin
...
+		if lastChar == '\n' {
+			newlineCount++
+
+			// Stop if two consecutive newlines are found
+			if newlineCount >= 2 {
+				break
+			}
+		} else if lastChar == '\r' || lastChar == ' ' || lastChar == '\t' {
+			// The presence of other whitespace characters (`\r`, ` `, `\t`) shouldn't reset the count.
+			// (Intentionally do nothing.)
```
rgmz, line 141: This handles https://github.com/gitleaks/gitleaks/pull/1665#issuecomment-2539410608

detect/directory.go
ahrav: This looks correct to me. I agree—it does feel a bit odd to use all three functions, but if I’m understanding this correctly, they each serve different purposes, so it actually makes sense.

rgmz: I'm concerned that `f.Read` and `r.ReadByte` may not place nicely. E.g., after the first iteration does calling `f.Read` also progress the internal reader used by bufio?

The number of bytes read (from #1670) appears to be consistent. 🤷

PR #1679, opened 2024-12-24
Add comma as operator for GenerateSemiGenericRegex

rmdwe:
### Description:
In some cases we may accidentally add secrets in configuration like this:
```
System.setProperty("ARTIFACTORY_TOKEN", "<SECRET>")
```
Which caused a security incident our enterprise codebases.

That's why we want to add the comma `,` , which haven't included as an operator, this pr aims to add comma as an operator for secret detection.

### Checklist:


zricethezav:
Good catch, I can see this being fairly common. Thanks for the PR!

PR #1731, opened 2025-01-26
Add experimental allowlist optimizations

rgmz:
### Description:
These are experimental changes to allowlist to improve efficiency.

Ideas include the following -- not all of them are implemented yet, or obviously better:

- [x] Using a map instead of slices for `Commits`
- [x] Collapsing `Regexes` into a single pattern
- [x] Collapsing `Paths` into a single pattern
- [x] Using ahocorasick for `Stopwords` instead of a slice
- [x] Creating benchmarks of varying sizes to get a more accurate portrayal of performance

**Commits**
```
# Before
BenchmarkCommitAllowed
BenchmarkCommitAllowed-4      	 3539479	       335.0 ns/op
BenchmarkCommitNotAllowed
BenchmarkCommitNotAllowed-4   	 2950525	       403.2 ns/op

# After
BenchmarkCommitAllowed
BenchmarkCommitAllowed-4      	16353297	        79.87 ns/op
BenchmarkCommitNotAllowed
BenchmarkCommitNotAllowed-4   	17393956	        69.91 ns/op
```

**Paths**
```
# Before
BenchmarkPathAllowed
BenchmarkPathAllowed-4        	   68570	     15523 ns/op
BenchmarkPathNotAllowed
BenchmarkPathNotAllowed-4     	   41542	     27972 ns/op

# After
BenchmarkPathAllowed
BenchmarkPathAllowed-4        	  922988	      1256 ns/op
BenchmarkPathNotAllowed
BenchmarkPathNotAllowed-4     	  767101	      1474 ns/op
```

**Regexes**
```
# Before
BenchmarkRegexAllowed
BenchmarkRegexAllowed-4       	  100308	     10873 ns/op
BenchmarkRegexNotAllowed
BenchmarkRegexNotAllowed-4    	   38066	     31017 ns/op

# After
BenchmarkRegexAllowed
BenchmarkRegexAllowed-4       	 1243234	       947.4 ns/op
BenchmarkRegexNotAllowed
BenchmarkRegexNotAllowed-4    	  835924	      1572 ns/op
```

cc @zricethezav @bplaxco 

### Potential Drawbacks

Off the top of my head, potential drawbacks could be:

- Slower startup speed to build an ahocorasick trie for every allowlist
- Faster performance for larger allowlists, at the cost of slower or equivalent performance for smaller allowlists
- Changes the semantics of the `Allowlist` struct, could mess with how some people use Gitleaks as a library

### Checklist:

### DIFF
--- a/config/allowlist.go
+++ b/config/allowlist.go
@@ -7,7 +7,6 @@ import (
 	ahocorasick "github.com/BobuSumisu/aho-corasick"
 	"golang.org/x/exp/maps"
 
-	"github.com/zricethezav/gitleaks/v8/config/flags"
 	"github.com/zricethezav/gitleaks/v8/regexp"
 )
 
@@ -59,6 +58,12 @@ type Allowlist struct {
 
 	// validated is an internal flag to track whether `Validate()` has been called.
 	validated bool
+
+	// EnableExperimentalOptimizations must be set prior to calling `Validate()`.
+	// See: https://github.com/gitleaks/gitleaks/pull/1731
+	//
+	// NOTE: This flag may be removed in the future.
+	EnableExperimentalOptimizations bool
 	// commitMap is a normalized version of Commits, used for efficiency purposes.
 	// TODO: possible optimizations so that both short and long hashes work.
 	commitMap    map[string]struct{}
@@ -80,8 +85,6 @@ func (a *Allowlist) Validate() error {
 		return errors.New("must contain at least one check for: commits, paths, regexes, or stopwords")
 	}
 
-	useOptimizations := flags.EnableExperimentalAllowlistOptimizations.Load()
-
 	// Deduplicate commits and stopwords.
 	if len(a.Commits) > 0 {
 		uniqueCommits := make(map[string]struct{})
@@ -89,7 +92,7 @@ func (a *Allowlist) Validate() error {
 			// Commits are case-insensitive.
 			uniqueCommits[strings.TrimSpace(strings.ToLower(commit))] = struct{}{}
 		}
-		if useOptimizations {
+		if a.EnableExperimentalOptimizations {
 			a.commitMap = uniqueCommits
 		} else {
 			a.Commits = maps.Keys(uniqueCommits)
@@ -102,7 +105,7 @@ func (a *Allowlist) Validate() error {
 		}
 
 		values := maps.Keys(uniqueStopwords)
-		if useOptimizations {
+		if a.EnableExperimentalOptimizations {
 			a.stopwordTrie = ahocorasick.NewTrieBuilder().AddStrings(values).Build()
 		} else {
 			a.StopWords = values
@@ -110,11 +113,13 @@ func (a *Allowlist) Validate() error {
 	}
 
 	// Combine patterns into a single expression.
-	if len(a.Paths) > 0 && useOptimizations {
-		a.pathPat = joinRegexOr(a.Paths)
-	}
-	if len(a.Regexes) > 0 && useOptimizations {
-		a.regexPat = joinRegexOr(a.Regexes)
+	if a.EnableExperimentalOptimizations {
+		if len(a.Paths) > 0 {
+			a.pathPat = joinRegexOr(a.Paths)
+		}
+		if len(a.Regexes) > 0 {
+			a.regexPat = joinRegexOr(a.Regexes)
+		}
 	}
 
 	a.validated = true
--- a/config/allowlist.go
+++ b/config/allowlist.go
@@ -3,7 +3,6 @@ package config
 import (
 	"errors"
 	"strings"
-	"sync"
 
 	ahocorasick "github.com/BobuSumisu/aho-corasick"
 	"golang.org/x/exp/maps"
@@ -68,18 +67,10 @@ type Allowlist struct {
 	stopwordTrie *ahocorasick.Trie
 }
 
-var (
-	flagOnce         sync.Once
-	useOptimizations bool
-)
-
 func (a *Allowlist) Validate() error {
 	if a.validated {
 		return nil
 	}
-	flagOnce.Do(func() {
-		useOptimizations = flags.EnableExperimentalAllowlistOptimizations.Load()
-	})
 
 	// Disallow empty allowlists.
 	if len(a.Commits) == 0 &&
@@ -88,6 +79,8 @@ func (a *Allowlist) Validate() error {
 		len(a.StopWords) == 0 {
 		return errors.New("must contain at least one check for: commits, paths, regexes, or stopwords")
 	}
+
+	useOptimizations := flags.EnableExperimentalAllowlistOptimizations.Load()
 
 	// Deduplicate commits and stopwords.
 	if len(a.Commits) > 0 {
@@ -152,14 +145,11 @@ func (a *Allowlist) CommitAllowed(c string) (bool, string) {
 	if a == nil || c == "" {
 		return false, ""
 	}
-	if useOptimizations {
-		if a.commitMap == nil {
-			return false, ""
-		}
+	if a.commitMap != nil {
 		if _, ok := a.commitMap[strings.ToLower(c)]; ok {
 			return true, ""
 		}
-	} else {
+	} else if len(a.Commits) > 0 {
 		for _, commit := range a.Commits {
 			if commit == c {
 				return true, c
@@ -174,15 +164,12 @@ func (a *Allowlist) PathAllowed(path string) bool {
 	if a == nil || path == "" {
 		return false
 	}
-
-	if useOptimizations {
-		if a.pathPat == nil {
-			return false
-		}
+	if a.pathPat != nil {
 		return a.pathPat.MatchString(path)
-	} else {
+	} else if len(a.Paths) > 0 {
 		return anyRegexMatch(path, a.Paths)
 	}
+	return false
 }
 
 // RegexAllowed returns true if the regex is allowed to be ignored.
@@ -190,15 +177,12 @@ func (a *Allowlist) RegexAllowed(secret string) bool {
 	if a == nil || secret == "" {
 		return false
 	}
-
-	if useOptimizations {
-		if a.regexPat == nil {
-			return false
-		}
+	if a.regexPat != nil {
 		return a.regexPat.MatchString(secret)
-	} else {
+	} else if len(a.Regexes) > 0 {
 		return anyRegexMatch(secret, a.Regexes)
 	}
+	return false
 }
 
 func (a *Allowlist) ContainsStopWord(s string) (bool, string) {
@@ -207,14 +191,11 @@ func (a *Allowlist) ContainsStopWord(s string) (bool, string) {
 	}
 
 	s = strings.ToLower(s)
-	if useOptimizations {
-		if a.stopwordTrie == nil {
-			return false, ""
-		}
+	if a.stopwordTrie != nil {
 		if m := a.stopwordTrie.MatchFirstString(s); m != nil {
 			return true, m.MatchString()
 		}
-	} else {
+	} else if len(a.StopWords) > 0 {
 		for _, stopWord := range a.StopWords {
 			if strings.Contains(s, stopWord) {
 				return true, stopWord
--- a/config/allowlist.go
+++ b/config/allowlist.go
@@ -3,9 +3,12 @@ package config
 import (
 	"errors"
 	"strings"
+	"sync"
 
+	ahocorasick "github.com/BobuSumisu/aho-corasick"
 	"golang.org/x/exp/maps"
 
+	"github.com/zricethezav/gitleaks/v8/config/flags"
 	"github.com/zricethezav/gitleaks/v8/regexp"
 )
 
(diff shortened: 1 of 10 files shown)


rgmz:
Also + @ahrav

config/allowlist.go
```diff
@@ -54,39 +55,49 @@ type Allowlist struct {
...
 func (a *Allowlist) CommitAllowed(c string) (bool, string) {
 	if c == "" {
 		return false, ""
 	}
-
-	for _, commit := range a.Commits {
-		if commit == c {
-			return true, c
-		}
+	if _, ok := a.commitMap[strings.ToLower(c)]; ok {
```
bplaxco, line 71: @rgmz are there places today where it double scans commits?

config/allowlist.go
```diff
@@ -103,17 +114,49 @@ func (a *Allowlist) Validate() error {
...
+			// Commits are case-insensitive.
+			uniqueCommits[strings.TrimSpace(strings.ToLower(commit))] = struct{}{}
 		}
-		a.Commits = maps.Keys(uniqueCommits)
+
+		//a.Commits = maps.Keys(uniqueCommits)
+		a.commitMap = uniqueCommits
+	}
+
+	if len(a.Paths) > 0 {
```
bplaxco, line 125: @rgmz I was thinking about doing something similar in my upstream patterns when they're "compiled". It'd be neat to have it baked in too.

So this does give a good performance boost?

config/allowlist.go
rgmz: Not that I'm aware of. Using a map make sense here because it's `o(1)`[^1] versus iterating the slice `o(n)`.

[^1]: or something like this.

bplaxco:
> Allowlist struct, could mess with how some people use Gitleaks as a library

Yeah anything that we can do to keep state/caches on the detector would be handy for my use case but I can always tweak things. I load the config once and then have a process that continually scans using gitleaks as a lib. I do create a fresh detector each scan though.

config/allowlist.go
rgmz: > So this does give a good performance boost?

The `BenchmarkPathAllowed` and `BenchmarkPathNotAllowed` benchmarks look to be 12-18x faster. We'd need to do some real-world testing to see if that carries over in practice.

Do you know of any large repositories with an excessive number of ignore paths? Node modules, Python deps, media files, etc.

ahrav:
> Also + @ahrav

This is really nifty! Based on my limited understanding of Gitleaks, this seems like a solid use of both the cache and Aho-Corasick. One small consideration—aside from the build-time drawback—is the additional memory overhead required to store the structure. However, from my testing of Aho-Corasick (specifically the library used here), the tradeoff is well worth it.

For my own understanding—apologies if I’m missing something about this codebase or the scanning approach—how do these optimizations affect hot paths? Does a typical scan run through all the optimized sections, or are there specific hotspots? Also, out of curiosity, do you have a CPU or trace profile that guided these optimizations?

P.S: ❤️ the benchmarks

rgmz:
> For my own understanding—apologies if I’m missing something about this codebase or the scanning approach—how do these optimizations affect hot paths? Does a typical scan run through all the optimized sections, or are there specific hotspots?

I suspect it runs through most, however, I don't have any data to support that. It would be interesting to gather.

Some users maintain their own config, [others (likely most) use the default config](https://open.substack.com/pub/lookingatcomputer/p/regex-is-almost-all-you-need?selection=8ddbbd08-f43c-45d6-b099-f0b48846f921). Moreover, the same config can have wildly different performance depending on what it's scanning. Every fragment is evaluated against [the global `[allowlist]`](https://github.com/gitleaks/gitleaks/blob/3188ad60055da792f5dd2bba21a88f4d5be80f96/config/gitleaks.toml#L15-L57). Every finding is evaluated against the rule's allowlist, if present (e.g., [`curl-auth-user`](https://github.com/gitleaks/gitleaks/blob/3188ad60055da792f5dd2bba21a88f4d5be80f96/config/gitleaks.toml#L299)), but no rule is guaranteed to hit. [`generic-api-key`](https://github.com/gitleaks/gitleaks/blob/3188ad60055da792f5dd2bba21a88f4d5be80f96/config/gitleaks.toml#L566) is most likely to hit and also has the most complicated allowlist.

> Also, out of curiosity, do you have a CPU or trace profile that guided these optimizations?

I do not. Are there any resources you recommend to get started with that?

ahrav:
> > For my own understanding—apologies if I’m missing something about this codebase or the scanning approach—how do these optimizations affect hot paths? Does a typical scan run through all the optimized sections, or are there specific hotspots?
> 
> I suspect it runs through most, however, I don't have any data to support that. It would be interesting to gather.
> 
> Some users maintain their own config, [others (likely most) use the default config](https://open.substack.com/pub/lookingatcomputer/p/regex-is-almost-all-you-need?selection=8ddbbd08-f43c-45d6-b099-f0b48846f921). Moreover, the same config can have wildly different performance depending on what it's scanning. Every fragment is evaluated against [the global `[allowlist]`](https://github.com/gitleaks/gitleaks/blob/3188ad60055da792f5dd2bba21a88f4d5be80f96/config/gitleaks.toml#L15-L57). Every finding is evaluated against the rule's allowlist, if present (e.g., [`curl-auth-user`](https://github.com/gitleaks/gitleaks/blob/3188ad60055da792f5dd2bba21a88f4d5be80f96/config/gitleaks.toml#L299)), but no rule is guaranteed to hit. Naturally, [`generic-api-key`](https://github.com/gitleaks/gitleaks/blob/3188ad60055da792f5dd2bba21a88f4d5be80f96/config/gitleaks.toml#L566) is most likely to hit, and also has the most complicated allowlist.
> 
> > Also, out of curiosity, do you have a CPU or trace profile that guided these optimizations?
> 
> I do not. Are there any resources you recommend to get started with that?

Thanks for the quick explanation. 🙇  Based on that, I’d expect these changes to have a noticeable impact. The [Go Blog](https://go.dev/doc/diagnostics#profiling) is a great starting point, as is [this guide](https://stackademic.com/blog/profiling-go-applications-in-the-right-way-with-examples).

config/allowlist.go
bplaxco: I don't off the top of my head, but it probably wouldn't be to hard to make one through automation.

You could generate a list of ignore paths that you'd like to make and then do something like:

```
target_number_of_commits = 1000
for target_number_of_commits:
   # the files to ignore
   allowlist_paths_for_commit = paths.choose(random_int(len(allow_list_example_paths) - 1)
   for path in allowlist_paths_for_commit:
        prefix = random_path(max_depth=30)
        create_file_with_parent_dirs(join_path(prefix, path), filesize=random_int(1024))

   # other random files
   for i in range(pow(len(allowlist_paths_for_commit), 2)):
       prefix = random_path(max_depth=30)
       create_file_with_parent_dirs(join_path(prefix, path), filesize=random_int(1024))

   commit()
```

rgmz:
@zricethezav I've moved the new logic behind a feature-flag, in case we want to test this "in the field" and compare results.

config/allowlist.go
rgmz: One potential candidate is [mattermost/mattermost-data-warehouse](https://github.com/mattermost/mattermost-data-warehouse.git) which takes ~45s-1m to scan. (It isn't consistent, and my laptop is generally quite slow.)

rgmz:
On a large (15GB) repository, this reduced the scanning time from ~46m to ~32m.

**Before**
```
(1) 2:53PM INF scanned ~15450035936 bytes (15.45 GB) in 46m53s
...
(2) 7:27PM INF scanned ~15450035936 bytes (15.45 GB) in 46m7s
```

**After**
```
(1) 5:25PM INF scanned ~15450035936 bytes (15.45 GB) in 32m58s
...
(2) 6:15PM INF scanned ~15450035936 bytes (15.45 GB) in 32m41s
```

config/allowlist.go
```diff
@@ -54,43 +57,96 @@ type Allowlist struct {
...
+			if strings.Contains(s, stopWord) {
+				return true, stopWord
+			}
 		}
 	}
 	return false, ""
 }
 
 func (a *Allowlist) Validate() error {
+	flagOnce.Do(func() {
```
rgmz, line 146: Relying on an 'optional' call to `Validate` to set internal state is bad. A better idea is probably to make a `NewAllowlist(...opts)` function, but I didn't want to add noise before this change got reviewed.

rgmz (commented):
@zricethezav This isn't _perfect_ but it's ready for review.

config/allowlist.go
```diff
@@ -76,16 +93,54 @@ func (a *Allowlist) Validate() error {
 	if len(a.Commits) > 0 {
 		uniqueCommits := make(map[string]struct{})
 		for _, commit := range a.Commits {
-			uniqueCommits[commit] = struct{}{}
+			// Commits are case-insensitive.
+			uniqueCommits[strings.TrimSpace(strings.ToLower(commit))] = struct{}{}
+		}
+		if useOptimizations {
```
bplaxco, line 99: @rgmz instead of using a flag in the allowlist.go, what are your thoughts on:

Adding a `Config.Optimize()` method that called Optimize on the sub items like:

```go
func (c *Config) Optimize() error {
	for _, allowlist := range c.Allowlists {
		if err := allowlist.Optimize(); err != nil {
			return fmt.Errorf("Allowlist optimization error: %w", err)
		}
	}

	for _, rule := range c.Rules {
		if err := a.Optimize(); err != nil {
			return fmt.Errorf("Rule %s optimization error: %w", rule.RuleID, err)
		}
	}

	return nil
}
```

Reasons for doing that:

- Fewer conditionals in the allowlist
- Lib users could do things like update their allowlist and call optimize again if needed
- Wouldn't have to add global flags; in cmd you could just call `cfg.Optimize()` if the cli flag was set for the cmd. Which might be trickier for lib users to interact with.

Downsides:

- Maybe slower than doing it inline here?


Thoughts?

bplaxco: > and an optional .Optimize() method would become redundant.

Fair point :ok_hand: 

>  I also hadn't considered the use case of modifying the allowlist after calling Validate();

And this was just a hypothetical.

> Admittedly flags.EnableExperimentalAllowlistOptimizations was designed with the CLI in mind

There's already at least one place where I do some pre setup for gitleaks for intercepting the log messages. Honestly it probably wouldn't be too awkward to do that too.

Though I do kinda feel like it would be nice to scope it somehow to the config/allowlist itself rather than making it a global flag :thinking: 

Maybe something like this?

```go
func Config(cmd *cobra.Command) config.Config {
	var vc config.ViperConfig
	if err := viper.Unmarshal(&vc); err != nil {
		logging.Fatal().Err(err).Msg("Failed to load config")
	}
	// set experimental feature flag(s)
	if mustGetBoolFlag(cmd, "experimental-optimizations") {
		logging.Warn().Msgf("using experimental allowlist optimizations, updates may contain breaking changes!")
		// This get passed through to the config during Translate()
		vc.EnableExprimentalOptimizations = true
	}
	
	cfg, err := vc.Translate()
	if err != nil {
		logging.Fatal().Err(err).Msg("Failed to load config")
	}
	cfg.Path, _ = cmd.Flags().GetString("config")

	return cfg
}
```

It'd make it easier to test that two versions of the allowlist (optimized and unoptimized) are the same without having to toggle global state.

Thoughts?

zricethezav: Trying to catch up here but 

> I'd be hesitant to introduce an Optimize feature because the intent isn't "improve the peformance of allowlist, if possible"

shouldn't the description be changed? I was under the assumption that this PR was to optimize the allowlist

rgmz: It is. By "Optimize feature" I'm referring to adding a new method to config/rule/allowlist called "Optimize()" which can be called at-will by lib users.

config/allowlist.go
```diff
@@ -97,10 +152,18 @@ func (a *Allowlist) CommitAllowed(c string) (bool, string) {
...
+		if _, ok := a.commitMap[strings.ToLower(c)]; ok {
+			return true, ""
+		}
+	} else {
+		for _, commit := range a.Commits {
+			if commit == c {
+				return true, c
+			}
 		}
 	}
```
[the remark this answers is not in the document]
rgmz: The bool was mainly for readability. I will remove it if you think the flow is clear enough.

config/allowlist.go
```diff
@@ -76,16 +93,54 @@ func (a *Allowlist) Validate() error {
...
+		sb.WriteString("(?:")
+		for i, path := range a.Paths {
+			sb.WriteString(path.String())
+			if i != len(a.Paths)-1 {
+				sb.WriteString("|")
+			}
+		}
+		sb.WriteString(")")
+		a.pathPat = regexp.MustCompile(sb.String())
+	}
```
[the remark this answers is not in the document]
rgmz: Done.

zricethezav:
I think this is good to go. @rgmz wdyt about giving this 2-3 weeks to bake as experimental then making this optimization the default behavior?

zricethezav:
> That sounds good to me! What do you think about https://github.com/gitleaks/gitleaks/pull/1731#discussion_r2127816521?

Sure, that sounds reasonable 👍🏻

rgmz:
I just realized the current implementation won't work for extended configs. Gotta run out now, will try to fix it later.

bplaxco:
"Crisp and clean, no caffeine" 👌👌

PR #1743, opened 2025-02-04
Add database password rule

ochcaroline:
### Description:
When doing checks in config files or workflows - there were multiple situations where the database passwords outside of connection strings were omitted during checking.
Examples would be: `DB_PASSWORD`, `DB_PASSWD`, `pwd` etc.

This adds handling of them, in various configurations and variations. There are also added unit tests for confirmation

### Checklist:


ochcaroline:
Weirdly enough - it fails with generating something that works on my machine :|
 …/gitleaks ¬   go generate ./...           
 …/gitleaks »  echo $?
0

Let me look if I can replicate

PR #1798, opened 2025-03-19
feat(regexp): use standard regexp by default, make go-re2 opt-in

### LINKED ISSUE #1796: Initialization performance regression due to switch to go-re2
**Describe the bug**

Since the switch to go-re2 in a91c6717a0bac43c292234ea0ead531e99ed2c55, gitleaks takes an extra 16ms to initialize. Previously, initialization was nearly instant.

This is likely due to the compilation of regular expressions with the new library.

Obviously this does not affect long-running processes that only pay the cost of compiling the regular expressions once on startup. It does affect short-lived processes that pay the compilation cost every time they are run.

This was first reported by @pbogut in https://github.com/twpayne/chezmoi/issues/4353 who used `git bisect` and further investigation to identify gitleaks as the cause of the slowdown. They identified `v8.21.2` as good and `v8.22.0` as bad.

**To Reproduce**

Compare the performance of gitleaks versions v8.21.2 and v8.22.0:

```console
$ git clone https://github.com/gitleaks/gitleaks.git
...
$ cd gitleaks
$ git checkout v8.21.2
...
$ ( go build . && time ./gitleaks --version ) 2>&1 | awk 'NR == 2 { print $3 }'
0.00s
$ git checkout v8.22.0
...
$ ( go build . && time ./gitleaks --version ) 2>&1 | awk 'NR == 2 { print $3 }'
0.16s
```

**Expected behavior**

gitleaks should have a near-zero start-up cost when imported as a library, especially if none of the regular expression used to detect secrets are actually used.

**Screenshots**

n/a

 - OS: Linux
 - Gitleaks Version: v8.22.0

**Additional context**

I would be happy to submit a pull request that uses something like [`sync.OnceValue`](https://pkg.go.dev/sync#OnceValue) to ensure that regular expressions are only compiled if they are actually needed.

cc @zricethezav @ahrav

twpayne:
### Description:

Fixes #1796.

This switches the default regular expression package back to Go's standard `regexp` package. Users can still opt-in to using `github.com/wasilabs/go-re2` by building with the `gore2regex` build tag.

See the discussion in #1796. This regexp will likely be a significant performance improvement in all cases except very long-running processes. Specifically, go-re2 has very high initialization and regular expression compilation costs that the standard library does not.

### Checklist:

* [x] Does your PR pass tests? Yes -- all existing tests pass.
* [x] Have you written new tests for your changes? No -- they are already covered by existing tests.
* [x] Have you lint your code locally prior to submission? Yes -- there are lint errors in the current code, but they are unrelated to this PR.

### DIFF
--- a/regexp/stdlib_regex.go
+++ b/regexp/stdlib_regex.go
@@ -1,4 +1,4 @@
-//go:build stdregex
+//go:build !gore2regex
 
 package regexp
 
--- a/regexp/wasilibs_regex.go
+++ b/regexp/wasilibs_regex.go
@@ -1,4 +1,4 @@
-//go:build !stdregex
+//go:build gore2regex
 
 package regexp
 


rgmz:
> This regexp will likely be a significant performance improvement in all cases **except very long-running processes**. Specifically, go-re2 has very high initialization and regular expression compilation costs that the standard library does not.

IIRC go-re2 is faster in all cases, at the cost of increased startup time and memory usage.

twpayne:
Picking a random line from go-re2's benchmarks:

```
name \ time/op                  build/bench_stdlib.txt  build/bench.txt   build/bench_cgo.txt

Match/Medium/1M-4                     29406989.0n ± ∞ ¹      42947.0n ± ∞ ¹     -99.85% (p=0.008 n=5)     154.6n ± ∞ ¹   -100.00% (p=0.008 n=5)
```

The standard library takes 29.4ns per medium match, whereas go-re2 takes only 0.043ns per medium match (this is very suspicious: I wonder if there's a problem with their benchmarks). Given that go-re2 has an initialization cost of 160ms, this means that go-re2 will be faster after 160ms/29.4ns = 5,442,176.9 matches.

rgmz:
> Given that go-re2 has an initialization cost of 160ms, this means that go-re2 will be faster after 160ms/29.4ns = 5,442,176.9 matches.

I don't know if the calculation is as straightforward, it'd be interesting to factor in *almost* matches.

I'm not against the change, I just want to be clear on the pros/cons. For most users there's a coin-flip whether the scan will be short or long(er) depending on the size, type, and contents of the input, so faster startup isn't necessarily always better.

anuraaga:
Just to clarify that benchmark, I think you're dividing because of the `1M` but that is the size of the input string, not number of matches. For a 1MB input, the stdlib takes 29ms vs 42us. The microbenchmarks are contrived examples though taken straight from golang/go and aren't that representative of real-world. `wafbench` is better since it's real regexes, albeit a specialized case, for 10K input it's 150ms vs 20ms.

Every project should decide about the tradeoff themselves ideally based on real-world data, as the first paragraphs of the README state - given how large I see lockfiles get in many projects, maybe it is here though.

zricethezav:
simple as 🙇🏻 

Thanks @twpayne!

PR #1810, opened 2025-04-02
Fix currentLine out of bounds error

bplaxco:
### Description:

`segment.adjustMatchIndex(matchIndex)` was above `segment.lineEndIndex(currentRaw, matchIndex[1]-matchIndex[0])` there were cases where it would cause an out of bounds error.

This change resolves the issue by just assuming the whole segment is a opaque chunk of text and starts at the boundaries of the segment instead so that the order doesn't matter as much.

The other option would have been to reorder the assignments and add tests to account for that case, but I figured this option is less likely to have gotchas when working with different encoding formats at the potential expense of getting an extra line or two if an encoding format in the future ever allows multiple lines of encoded content to be considered a single segment.

### Checklist:

* [ ] Have you written new tests for your changes? (**N/A**: covered by existing tests)

### DIFF
--- a/detect/decoder.go
+++ b/detect/decoder.go
@@ -280,24 +280,27 @@ func segmentWithDecodedOverlap(encodedSegments []EncodedSegment, start, end int)
 	return nil
 }
 
-func (s EncodedSegment) lineStartIndex(currentRaw string) int {
+func (s EncodedSegment) currentLine(currentRaw string) string {
+	start := 0
+	end := len(currentRaw)
+
+	// Find the start of the range
 	for i := s.decodedStart; i > -1; i-- {
 		c := currentRaw[i]
 		if c == '\n' {
-			return i
+			start = i
+			break
 		}
 	}
 
-	return 0
-}
-
-func (s EncodedSegment) lineEndIndex(currentRaw string, matchLen int) int {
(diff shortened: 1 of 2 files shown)


bplaxco:
@zricethezav, @thewizzy noticed this issue in our CI run that included some decoding tests:

https://github.com/leaktk/scanner/actions/runs/14210109004/job/39815606323

It thought the matchlen was 128 when len(currentRaw) was only 95.

I think this should fix the issue and be a little more robust with future decoding cases. @thewizzy, can you point the failing branch at the fork and confirm the fix as well? I tested it locally when making the change and it resolved it.

thewizzy:
> @zricethezav, @thewizzy noticed this issue in our CI run that included some decoding tests:
> 
> 
> 
> https://github.com/leaktk/scanner/actions/runs/14210109004/job/39815606323
> 
> 
> 
> It thought the matchlen was 128 when len(currentRaw) was only 95.
> 
> 
> 
> I think this should fix the issue and be a little more robust with future decoding cases. @thewizzy, can you point the failing branch at the fork and confirm the fix as well? I tested it locally when making the change and it resolved it.

I pointed it to the fix and it resolved the issue in our tests. 

https://github.com/leaktk/scanner/pull/107

zricethezav:
LGTM, thanks guys 👍🏻

PR #1817, opened 2025-04-08
Validate & sort rule when generating config

rgmz:
### Description:

This makes the following changes:
- Call Rule.Validate and Allowlist.Validate when generating the config
- Sort the allowlist stopwords so the output is deterministic (this seems to require `*Allowlist`)

### Checklist:


zricethezav:
lgtm, thanks!

PR #1841, opened 2025-04-28
WIP simple archive handling

zricethezav:
### Description:
Gitleaks previously skipped over archived files. This PR adds code that automatically expands archived data and scans it both in `dir` and `git` scanning modes.

### Checklist:


detect/archive.go
```diff
@@ -0,0 +1,96 @@
...
+		return false
+	}
+	_, ok := format.(archives.Extractor)
+	return ok
+}
+
+// ExtractArchive extracts all files from archivePath into a temp dir.
+// Returns the list of ScanTargets (with real file paths) and the temp dir for cleanup.
+func ExtractArchive(archivePath string) ([]sources.ScanTarget, string, error) {
+	tmpDir, err := os.MkdirTemp("", "gitleaks-archive-")
```
rgmz, line 31: I'm not a fan of how TH fills up the top-level tmp directory. Can we create a single temp dir per scan and nest artifacts under it? e.g.,
```
tmp/
├─ gitleaks-scan-123456/
│  ├─ archive-1234/
│  ├─ archive-4567/
│  ├─ blob.xyz
```

detect/archive.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	f, err := os.Open(archivePath)
+	if err != nil {
+		os.RemoveAll(tmpDir)
+		return nil, "", err
+	}
+	defer f.Close()
+
+	ctx := context.Background()
```
rgmz, line 43: If I push a PR adding top-level context propagation, would you merge it?

detect/directory.go
```diff
@@ -17,110 +17,151 @@ import (
 
 const maxPeekSize = 25 * 1_000 // 10kb
 
+// DetectFiles schedules each ScanTarget—file or archive—for concurrent scanning.
 func (d *Detector) DetectFiles(paths <-chan sources.ScanTarget) ([]report.Finding, error) {
 	for pa := range paths {
+		pa := pa // capture
```
rgmz, line 23: No longer necessary: https://go.dev/blog/loopvar-preview.
```suggestion
```

sources/git.go
```diff
@@ -115,12 +129,42 @@ func NewGitDiffCmd(source string, staged bool) (*GitCmd, error) {
...
+// and returns its path.
+func (g *GitCmd) CheckoutBlob(commit, filepathInRepo string) (string, error) {
+	// Create a temp file with the same extension as the blob, if possible
+	ext := filepath.Ext(filepathInRepo)
+	tmpFile, err := os.CreateTemp("", "gitleaks-blob-*"+ext)
+	if err != nil {
+		return "", fmt.Errorf("creating temp file for blob: %w", err)
+	}
+	tmpFilePath := tmpFile.Name()
+	tmpFile.Close()
```
rgmz, line 149: I'm guessing this is meant to be `defer`?

rgmz (commented):
Only a code review, I will test this branch later.

One concern I have with this approach is that it could make binary support (e.g., reading sqlite files) harder. I would rather have a "scan binary" flow that _only_ handles archives, than a hard-coded "scan archives" flow that needs to be dug up if/when we want to scan other types of binary files.

detect/archive.go
zricethezav: assuming it didn't break anything, sure

detect/detect.go
```diff
@@ -36,8 +36,17 @@ const (
 var (
 	newLineRegexp = regexp.MustCompile("\n")
 	isWindows     = runtime.GOOS == "windows"
+	tmpDir        string
 )
 
+func init() {
+	var err error
+	tmpDir, err = os.MkdirTemp("", "gitleaks-*")
```
zricethezav, line 44: @rgmz lemme know what you think of this approach. I considered creating a "tmpdirmanager" struct but that seemed like overkill.

PR #1849, opened 2025-05-02
Run commands with `context.Context`

rgmz:
### Checklist:


detect/detect.go
```diff
@@ -252,6 +252,13 @@ func (d *Detector) Detect(fragment Fragment) []report.Finding {
 	decoder := NewDecoder()
 
 	for {
+		select {
+		case <-ctx.Done():
```
rgmz, line 256: Is using the `chan` better than checking `ctx.Err() != nil`? Switch with an empty default feels weird.
https://www.reddit.com/r/golang/comments/1e1wdvy/comment/lcxfz14/

detect/git.go
```diff
@@ -28,6 +29,9 @@ func (d *Detector) DetectGit(cmd *sources.GitCmd, remote *RemoteInfo) ([]report.
 	// loop to range over both DiffFiles (stdout) and ErrCh (stderr)
 	for diffFilesCh != nil || errCh != nil {
 		select {
+		case <-ctx.Done():
+			// Return context error if context is cancelled
+			return d.findings, ctx.Err()
```
rgmz, line 34: This is awkward. IMO we'd either need to defer `sema.wait()` or call `break` here.

main.go
```diff
@@ -11,14 +13,25 @@ import (
 func main() {
 	// this block sets up a go routine to listen for an interrupt signal
 	// which will immediately exit gitleaks
-	stopChan := make(chan os.Signal, 1)
-	signal.Notify(stopChan, os.Interrupt)
-	go listenForInterrupt(stopChan)
+	ctx, cancel := context.WithCancel(context.Background())
```
rgmz, line 16: Copied from [poutine](https://github.com/boostsecurityio/poutine/blob/cee72a96707d8a389d2c40eee1fa47ea2e63af95/cmd/root.go#L71). I like the "first Ctrl+C gracefully stops, second exits" functionality but it isn't necessary.

detect/detect.go
```diff
@@ -199,19 +199,19 @@ func (d *Detector) AddGitleaksIgnore(gitleaksIgnorePath string) error {
 }
 
 // DetectBytes scans the given bytes and returns a list of findings
-func (d *Detector) DetectBytes(content []byte) []report.Finding {
```
zricethezav, line 202: @rgmz probably should keep these functions since there are more than a few companies using gitleaks as a library

rgmz: That was on my mind. What's the better option, new alternative `DetectXContext` methods?

zricethezav: Unfortunately I think that's the best option. That, or just wait until v9

rgmz: I've updated the code to preserve the original functions. They're now wrappers for the `*Context` methods.

bplaxco:
> I think this was covered by #1872, right @bplaxco?

Partially. It gets passed down in more places bit it isn't passed into things like `NewGit*Cmd` and some other places.

PR #1872, opened 2025-05-22
Archive support

bplaxco:
### Description:

Related to https://github.com/gitleaks/gitleaks/pull/1841

Talked with @zricethezav  to see if I could take a go at doing it in-memory and with a change to how sources are handled.

The idea is that sources could have a `Fragments(FragmentFunc) error` function where `type FragmentFunc func(Fragment, error) error`. The behavior should be like `WalkDir` where the callback determines how an error should be handled and if the callback returns an error, fragment traversal stops and the function returns that error.

### Tasks:

* [x] Convert the file based detect functions to Source.Fragment calls
* [x] Update the cmds for files to use `Source.Fragments`
* [x] Add basic archive support to `File`
* [x] Add new tests for the File/Files changes
* [x] Fix issues with certain types of nested archives
* [x] Add `Git` source and update the `DetectGit` related functions ad cmds
* [x] Add archive support in `Git`
* [x] Lint the code
* [x] Determine how to best represent "inner" paths (e.g. /external.zip!inner/path.tar.gz!even/more/inner)
* [x] Add add remaining tests for Git
* [x] Adjust finding link paths to stop at the inner path separator (and write tests)
* [x] Add `--max-archive-depth= (default 0)` flag && update docs (thank you @alayne222!)
* [x] Review readers to look for places where they aren't closed
* [x] Confirm backwards compatibility (i.e. line numbers, secrets counts, etc)[1]
* [x] Log file path [here](https://github.com/gitleaks/gitleaks/pull/1872/files#r2114147256)
* [x] Performance test & tune
* [x] Profit :sunglasses: 

[1]: I want to make sure I completely resolved the issue I was seeing before where the secret counts weren't lining up properly. Some of it was a problem related to me using a buffered reader in one spot and then using the reader again. But then there were some weird cases where I wasn't seeing the same number of secrets in every place. 

### Checklist:


sources/file.go
```diff
@@ -0,0 +1,112 @@
...
+
+	for {
+		n, err := reader.Read(buf)
+		if n == 0 {
+			if err != nil && err != io.EOF {
+				return yield(Fragment{}, fmt.Errorf("could not read file: %w", err))
+			}
+
+			return nil
+		}
```
rgmz, line 49: IMO it's more straightforward to do `if n > 0`, rather than checking the error in two different places.

sources/files.go
```diff
@@ -0,0 +1,192 @@
...
+				logger.Warn().Msgf(
+					"skipping file: too large: max_size=%dMiB, size=%dMiB",
+					s.MaxFileSize/1000000, info.Size()/1000000,
+				)
+				return nil
+			}
+		}
+
+		// set the inital scan target values
+		if d.Type()&fs.ModeSymlink != 0 {
```
rgmz, line 111: What black magic is this?

sources/files.go
```diff
@@ -0,0 +1,192 @@
...
+			file := File{
+				Content: f,
+				Path:    scanTarget.Path,
+				Symlink: scanTarget.Symlink,
+			}
+
+			return file.Fragments(yield)
+		})
+
+		wg.Wait()
```
rgmz, line 189: Won't this run each target serially, compared to `DetectFiles`?

cmd/directory.go
```diff
@@ -52,27 +47,24 @@ func runDirectory(cmd *cobra.Command, args []string) {
...
-	if err != nil {
-		logging.Fatal().Err(err)
-	}
-
-	findings, err = detector.DetectFiles(paths)
+	findings, err := detector.DetectSource(files)
 	if err != nil {
-		// don't exit on error, just log it
-		logging.Error().Err(err).Msg("failed scan directory")
+		logging.Fatal().Err(err).Msg("failed scan directory")
```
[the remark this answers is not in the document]
bplaxco: I think this was unintentional. I was in the middle of consolidating code and it came from here:

https://github.com/gitleaks/gitleaks/pull/1872/files/d789d47020a350e4436378e8eb5c0f1b74dfd257

Good catch!

detect/detect.go
```diff
@@ -239,12 +235,12 @@ func (d *Detector) Detect(fragment Fragment) []report.Finding {
 	}
 	// check if commit or filepath is allowed.
 	if isAllowed, event := checkCommitOrPathAllowed(logger, fragment, d.Config.Allowlists); isAllowed {
-		event.Msg("skipping file: global allowlist")
+		event.Msg("skipping file: global allowlist:")
```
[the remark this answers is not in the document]
bplaxco: The logger adds key=value pairs at the end and this was just to make the output look something like:

```
skipping file: global allowlist: condition=OR allowed-path=foo/bar/baz
```

The nice thing about this format is it nests well:

```go
// assume someIoError = "permission denied:"
err : = fmt.Errorf("some thing happened: %w some=key", someIoError)
return fmt.Errorf("could not complete scan: %w source=/some/source")
```

You get:

```
could not complete scan: some thing happened: permission denied: some=key source=/some/source
```

Just muscle memory for logs and I was looking out of the output when tweaking things and added it out of habit. but that's probably out of scope for the PR. Think it's worth skipping?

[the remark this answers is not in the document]
bplaxco: Yeah, I'll probably revert actually.

sources/files.go
bplaxco: Ah, I missed that `Type()` already `returns type bits in m (m & ModeType).` 

I was just doing double duty by mistake. (Will fix)

sources/file.go
bplaxco: I kept changing it around to do that but also not pass EOF to yield && still passing any amount of fragment and each way felt awkward to me and ended up being about the same amount of checks. 

I could also just be missing a simple way to do it though.

sources/files.go
bplaxco: Yerp! Put it in the wrong spot.

sources/files.go
zricethezav: it's interesting that goimports makes `d.Type() & fs.ModeSymlink` -> `d.Type()&fs.ModeSymlink`. TIL. Is it just me or is `d.Type()&fs.ModeSymlink` more confusing to read?

detect/reader.go
```diff
@@ -1,52 +1,36 @@
...
-	"bytes"
-	"errors"
 	"io"
 
 	"github.com/zricethezav/gitleaks/v8/report"
+	"github.com/zricethezav/gitleaks/v8/sources"
 )
 
 // DetectReader accepts an io.Reader and a buffer size for the reader in KB
 func (d *Detector) DetectReader(r io.Reader, bufSize int) ([]report.Finding, error) {
```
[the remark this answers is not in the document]
bplaxco: I was thinking that for a reader, that'd effectively be just an "unamed file"

So something like:

```go
stdin := sources.File{Content: os.Stdin}
findings, err := d.DetectSource(stdin)
// ... and so on
```

For cases where you need to stream the results:

```go
stream := &sources.File{Content: neverEndingReader}
return stream.Fragments(func (fragment sources.Fragment, err error) error) {
    // do streamy things here
})
```

Thoughts?

(Another option would be to make a Reader source, but then have File just use that internally to handle its reader bits.)

cmd/detect.go
```diff
@@ -71,44 +71,48 @@ func runDetect(cmd *cobra.Command, args []string) {
...
 		)
-		if err != nil {
-			logging.Fatal().Err(err).Send()
-		}
 
-		if findings, err = detector.DetectFiles(paths); err != nil {
-			// don't exit on error, just log it
-			logging.Error().Err(err).Msg("failed scan directory")
+		if err != nil {
+			logging.Fatal().Err(err)
```
rgmz, line 89: This is a no-op, right? The logger needs a terminal operator like `Msg()`, `Msgf()`, or `.Send()`.

cmd/stdin.go
```diff
@@ -35,10 +36,13 @@ func runStdIn(cmd *cobra.Command, _ []string) {
...
-	findings, err := detector.DetectReader(os.Stdin, 10)
+	findings, err := detector.DetectSource(
+		&sources.File{
+			Content: os.Stdin,
+		},
+	)
+
 	if err != nil {
-		// log fatal to exit, no need to continue since a report
-		// will not be generated when scanning from a pipe...for now
```
rgmz, line 41: Probably worth keeping this comment?

detect/detect.go
```diff
@@ -211,6 +189,24 @@ func (d *Detector) DetectString(content string) []report.Finding {
 	})
 }
 
+// DetectSource scans a source's fragments for findings
+func (d *Detector) DetectSource(source sources.Source) ([]report.Finding, error) {
+	err := source.Fragments(func(fragment sources.Fragment, err error) error {
+		if err != nil {
+			// don't exit on error, just log it
+			logging.Error().Err(err).Msgf("scan directory issue: %s", err)
```
rgmz, line 197: This is redundant:
```suggestion
			logging.Err(err).Msgf("scan directory issue: %s", err)
```

scripts/profile.sh
rgmz, line 1: It would be nice if this failed gracefully when you don't have `yq` or `hyperfine` installed.

detect/files.go
```diff
@@ -0,0 +1,83 @@
...
+			// Convert this to a file source
+			file := &sources.File{
+				Content: f,
+				Path:    scanTarget.Path,
+				Symlink: scanTarget.Symlink,
+			}
+
+			return file.Fragments(func(fragment sources.Fragment, err error) error {
+				if err != nil {
+					logging.Error().Msg(err.Error())
```
rgmz, line 69:
```suggestion
					logging.Err(err).Send()
```

sources/file.go
```diff
@@ -0,0 +1,219 @@
...
+	// Buffer is used for reading the content in chunks
+	Buffer []byte
+	// outerPaths is the list of container paths (e.g. archives) that lead to
+	// this file
+	outerPaths []string
+}
+
+// Fragments yields fragments for the this source
+func (s *File) Fragments(yield FragmentsFunc) error {
+	ctx := context.Background()
```
rgmz, line 44: IMO we should try to support contexts for new APIs.

See: https://github.com/gitleaks/gitleaks/pull/1841#discussion_r2071559066 and #1849.

sources/file.go
```diff
@@ -0,0 +1,219 @@
...
+
+	return s.fileFragments(s.Content, yield)
+}
+
+// extractorFragments recursively crawls archives and yields fragments
+func (s *File) extractorFragments(ctx context.Context, extractor archives.Extractor, reader io.Reader, yield FragmentsFunc) error {
+	if _, isSeekReaderAt := reader.(seekReaderAt); !isSeekReaderAt {
+		switch extractor.(type) {
+		case archives.SevenZip, archives.Zip:
+			tmpfile, err := os.CreateTemp("", "gitleaks-archive-")
```
rgmz, line 65: Could we create a single scan directory and nest things underneath?

https://github.com/gitleaks/gitleaks/pull/1841#discussion_r2071557965

rgmz (commented):
Miscellaneous questions and style suggestions.

sources/file.go
rgmz: Hmm, I will try to play around with this and see if there's a better way.

cmd/stdin.go
bplaxco: Yep!

detect/detect.go
bplaxco: Fixed!

scripts/profile.sh
bplaxco: Yeah, it currently just leaves bits of the profile data out if those aren't installed :ok_hand:

sources/file.go
bplaxco: +1 

@zricethezav want me to go through and change `Fragments(FragmentsFunc)` -> `Fragments(context.Context, FragmentsFunc)` ?

sources/file.go
bplaxco: @rgmz should be doable, but what would be the benefit? Without the parent directory it currently makes cleanup cleaner. 

(Not saying no or pushing back, just trying to get a feel for why ^_^)

detect/files.go
bplaxco: Fixed!

zricethezav:
> archive scanning is disabled by default

I think that makes sense. Disabled by default w/ the caveat that this is an experimental feature that probably _will_ be enabled by default in the future. In the scenario where it's enabled by default, we could use the same `--max-archive-depth` flag to disable the feature (`--max-archive-depth=0`).

> inner paths are separated by sources.InnerPathSeparator

That makes sense to me. Is coming up with `!` original or is there precedent for that in tooling you've seen?

> if you enable archive scanning, you'll want to make sure your tooling can handle inner paths separated by sources.InnerPathSeparator.

Probably something worth mentioning in the README. In fact, Archive handling should have an entire section in the README like `Decode`.

bplaxco:
> That makes sense to me. Is coming up with ! original or is there precedent for that in tooling you've seen?

@zricethezav, @rgmz suggested it a while back and I initially pushed back in favor of `:` but then:

- There were issues with the fingerprints
- I expected there will probably be other places that have similar quirks since `:` already means something.

> is there precedent for that in tooling you've seen

Yeah, things some things already use `!` to indicate a path within an archive. Java related stuff is the first that comes to mind. 

To help guage how intuative it is I asked Gemini 2.5 Flash:

**Prompt**:

> I want to make sure a notation is clear. What does this filepath mean to you?
> 
> ```
> some.tar.gz!some/path/to/file.txt
> ```

**Response**:

> This notation, `some.tar.gz!some/path/to/file.txt`, strongly suggests a way to reference a **file located *inside* a compressed archive**.
> 
> Here's how I interpret it:
> 
> * **`some.tar.gz`**: This is the **archive file** itself. The `.tar.gz` extension indicates it's a gzipped tarball, a common format for bundling and compressing multiple files and directories.
> * **`!`**: This acts as a **delimiter** or separator. It signifies that what follows is *within* the preceding archive.
> * **`some/path/to/file.txt`**: This is the **path to the specific file *within* the `some.tar.gz` archive**. It indicates the file `file.txt` is located in a directory structure `some/path/to/` inside the compressed archive.
> 
> This notation is not a standard filesystem path that your operating system would natively understand (like `/home/user/documents/file.txt` or `C:\Users\user\Documents\file.txt`). Instead, it's typically used by:
> 
> * **Software libraries or tools:** Many programming languages and utilities have functions to directly access content within archives using similar conventions. For example, some Java libraries use a similar `jar:file:/path/to/archive.jar!/path/to/entry` syntax.
> * **Specialized file managers or applications:** Some programs that are designed to browse or extract archives might display or allow you to input paths in this manner.
> * **Documentation or informal communication:** It's a clear and concise way to convey "this file is inside that archive."
> 
> In essence, it means "the file `file.txt` which is located at `some/path/to/` *inside* the archive `some.tar.gz`."



> Probably something worth mentioning in the README. In fact, Archive handling should have an entire section in the README like Decode.

:100: Was defiantly going to do the doco. Added a note in my TODOs above to make sure it's captured.

rgmz:
The syntax originates with Java AFAIK. `!` is the claspath separator, used to indicate a jar/zip boundary.

bplaxco:
@rgmz oh btw I added ctx to `Source.Fragments(ctx, ...)` and `Detector.DetectSource(ctx, ...)` ^_^

bplaxco:
@zricethezav heh, the compatibility testing is paying off (https://github.com/gitleaks/gitleaks/pull/1872/commits/260f83181d494f6ecf06bcb68396b175d60dd26e) there was a 2 finding difference scanning the fake-leaks repo before that tweak :sweat_smile: 

Still testing but now the diffs between the sorted[1] results files are currently coming back clean :ok_hand: 


[1] Sorting and reviewing via this since the order of the findings can differ slightly due to async:

```sh
diff  \
        <(cat master-report.json | jq 'map(. | to_entries | sort_by(.key) | from_entries) | sort_by(.Fingerprint)') \
        <(cat archives-report.json | jq 'map(. | to_entries | sort_by(.key) | from_entries) | sort_by(.Fingerprint)') \
        | bat
```

bplaxco:
Okay, I'm pretty happy with backwards compatibility when `--max-archive-depth` isn't set. I'm not seeing any outputs in the diffs. Poking at the performance.

bplaxco:
**Stats**

The `archives` and `master` commands are built from the branches by the same names. And the gitleaks.toml is the one from the archives branch. It was specified so that some repos with .gitleaks.toml's wouldn't change my test results.

**Scanning the kubernetes repo:**
```sh
./archives git \
  --config gitleaks.toml --exit-code=0 \
  --max-archive-depth=8 --max-decode-depth=8 \
  repos/git/kubernetes 
```

```
--- Statistics for Archives Run ---
  Number of samples: 28
  Mean VmRSS: 77109.14 kB
  Median VmRSS: 74426.00 kB
  Standard Deviation: 12802.88 kB
  Minimum VmRSS: 28624.00 kB
  Maximum VmRSS: 101880.00 kB
  Range (Max - Min): 73256.00 kB
```

```sh
./master git \
  --config gitleaks.toml --exit-code=0 \
  --max-decode-depth=8 \
  repos/git/kubernetes 
```

```
--- Statistics for Master Run ---
  Number of samples: 18
  Mean VmRSS: 69884.22 kB
  Median VmRSS: 67014.00 kB
  Standard Deviation: 22562.21 kB
  Minimum VmRSS: 24572.00 kB
  Maximum VmRSS: 116020.00 kB
  Range (Max - Min): 91448.00 kB
  ````
  
**Scanning the gitlab repo:**

```sh
./archives git \
  --config gitleaks.toml --exit-code=0 \
  --max-archive-depth=8 --max-decode-depth=8 \
  repos/git/gitlab 
```
```
  --- Statistics for Archives Run ---
  Number of samples: 77
  Mean VmRSS: 535470.08 kB
  Median VmRSS: 483032.00 kB
  Standard Deviation: 132735.50 kB
  Minimum VmRSS: 24268.00 kB
  Maximum VmRSS: 836596.00 kB
  Range (Max - Min): 812328.00 kB
```

```sh
./master git \
  --config gitleaks.toml --exit-code=0 \
  --max-decode-depth=8 \
  repos/git/gitlab 
```
```
--- Statistics for Master Run ---
  Number of samples: 73
  Mean VmRSS: 540648.44 kB
  Median VmRSS: 520232.00 kB
  Standard Deviation: 138287.48 kB
  Minimum VmRSS: 21216.00 kB
  Maximum VmRSS: 842784.00 kB
  Range (Max - Min): 821568.00 kB

--- Comparative Analysis ---
  Average VmRSS difference: -5178.36 kB
  Max VmRSS difference: -6188.00 kB
  Standard Deviation difference: -5551.99 kB
  ```

**Comment by zricethezav** [comment]
@bplaxco so it looks like the `archive` branch profiled against the k8 repo uses significantly less memory (20MB) than `master`?  Against GitLab it looks about the same. Any hypothesis why we see the `archive` perform well against k8? Is it as simple as `yield` is more memory efficient that sending on `chan`s?

File: cmd/detect.go [hunk 13]
```diff
@@ -51,15 +51,15 @@ var detectCmd = &cobra.Command{
 func runDetect(cmd *cobra.Command, args []string) {
 	// start timer
 	start := time.Now()
-	source := mustGetStringFlag(cmd, "source")
+	sourcePath := mustGetStringFlag(cmd, "source")
```
zricethezav, line 54: 🙏🏻 every time I opened this file I would change this to `sourcePath` then forget about it. Thank you lol

bplaxco:
> @bplaxco so it looks like the `archive` branch profiled against the k8 repo uses significantly less memory (20MB) than `master`?  Against GitLab it looks about the same. Any hypothesis why we see the `archive` perform well against k8? Is it as simple as `yield` is more memory efficient that sending on `chan`s?

@zricethezav I'm not sure, would you be opposed to me tweaking the diagnostic feature to use net/http/pprof so you can sample the pprof data as needed during the run before it writes out the final dump at the end?

I might be able to take some of those and the pprof diff flags to see where the savings are happening.

detect/detect.go
```diff
@@ -102,35 +106,15 @@ type Detector struct {
...
-	WindowsFilePath string `json:"-"` // TODO: remove this in v9.
-
-	// CommitSHA is the SHA of the commit if applicable
-	CommitSHA string
-
-	// newlineIndices is a list of indices of newlines in the raw content.
-	// This is used to calculate the line location of a finding
-	newlineIndices [][]int
-}
+type Fragment sources.Fragment
```
zricethezav, line 109: @bplaxco could a comment be added above this to deprecate v9? This confused during my first pass

cmd/root.go
```diff
@@ -240,6 +241,10 @@ func Detector(cmd *cobra.Command, cfg config.Config, source string) *detect.Dete
 		logging.Fatal().Err(err).Msg("")
 	}
 
+	if detector.MaxArchiveDepth, err = cmd.Flags().GetInt("max-archive-depth"); err != nil {
+		logging.Fatal().Err(err).Msg("")
```
rgmz, line 245: IMO the intent of [`.Send()`](https://pkg.go.dev/github.com/rs/zerolog#Event.Send) is clearer than `.Msg("")`, which could easily be a mistake or a forgotten "TODO".
```suggestion
		logging.Fatal().Err(err).Send()
```

detect/detect.go
```diff
@@ -211,6 +195,68 @@ func (d *Detector) DetectString(content string) []report.Finding {
...
+			d.addCommit(fragment.CommitSHA)
+			logger := logContext.Logger()
+			logger.Warn().Msg("commit SHAs should be >= 7 characters long")
+		}
+
+		logger := logContext.Logger()
+
+		if err != nil {
+			// Log the error and move on to the next fragment
+			logger.Error().Err(err).Send()
```
rgmz, line 221:
```suggestion
			logger.Err(err).Send()
```

detect/detect.go
```diff
@@ -322,20 +368,28 @@ func (d *Detector) detectRule(fragment Fragment, currentRaw string, r config.Rul
...
+					finding.Message = fragment.CommitInfo.Message
+				}
 				return append(findings, finding)
 			}
 		} else {
 			// if path is set _and_ a regex is set, then we need to check both
 			// so if the path does not match, then we should return early and not
 			// consider the regex
-			if !(r.Path.MatchString(fragment.FilePath) || (fragment.WindowsFilePath != "" && r.Path.MatchString(fragment.WindowsFilePath))) {
+			if !r.Path.MatchString(fragment.FilePath) && (fragment.WindowsFilePath == "" || !r.Path.MatchString(fragment.WindowsFilePath)) {
```
rgmz, line 392: Was this a De Morgan's Law suggestion? I think this is significantly harder to read than the prior version.

detect/files.go
```diff
@@ -0,0 +1,92 @@
...
+				logger.Warn().Err(err).Msg("skipping file")
+				return nil
+			}
+			defer func() {
+				_ = f.Close()
+			}()
+
+			info, err := f.Stat()
+			if err != nil {
+				logger.Error().Err(err).Msg("skipping file: could not get info")
```
rgmz, line 44:
```suggestion
				logger.Err(err).Msg("skipping file: could not get info")
```

sources/common.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	for _, a := range cfg.Allowlists {
+		if a.PathAllowed(path) ||
+			// TODO: Remove this in v9.
+			// This is an awkward hack to mitigate https://github.com/gitleaks/gitleaks/issues/1641.
+			(isWindows && a.PathAllowed(filepath.ToSlash(path))) {
+			return true
+		}
+	}
```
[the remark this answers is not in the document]
rgmz: Yes, this was implemented prior to multiple global allowlists. Could be done in a future PR if it's easier.

sources/file.go
```diff
@@ -0,0 +1,247 @@
...
+	// MaxArchiveDepth limits how deep the sources will explore nested archives
+	MaxArchiveDepth int
+	// archiveDepth is the current archive nesting depth
+	archiveDepth int
+}
+
+// Fragments yields fragments for the this source
+func (s *File) Fragments(ctx context.Context, yield FragmentsFunc) error {
+	format, _, err := archives.Identify(ctx, s.Path, nil)
+	if err == nil && format != nil {
```
rgmz, line 53: If `err != nil` we swallow it. Is that intended? If so, probably worth a comment.

sources/file.go
```diff
@@ -0,0 +1,247 @@
...
+		}
+
+		innerReader, err := d.Open()
+		if err != nil {
+			logging.Error().Str("path", s.FullPath()).Err(err).Msg("could not open archive inner file")
+			return nil
+		}
+		path := filepath.Clean(d.NameInArchive)
+
+		if s.Config != nil && shouldSkipPath(s.Config, path) {
```
rgmz, line 111: `innerReader.Close()` should be a deferred statement after the err check. Presently, I believe this will cause a leak if this is true.

sources/file.go
```diff
@@ -0,0 +1,247 @@
...
+					fmt.Errorf("could not read file: could not determine type: %w", err),
+				)
+			} else if mimetype.MIME.Type == "application" {
+				logging.Debug().Str(
+					"mime_type", mimetype.MIME.Value,
+				).Str(
+					"path", s.FullPath(),
+				).Msgf(
+					"skipping binary file",
+				)
```
rgmz, line 189: This is kinda whack, why not something like?
```suggestion
				logging.Debug().
				    Str("mime_type", mimetype.MIME.Value).
				    Str("path", s.FullPath()).
				    Msgf("skipping binary file")
```

detect/git.go
```diff
@@ -1,192 +1,32 @@
...
-	"github.com/gitleaks/go-gitdiff/gitdiff"
+	"context"
 
 	"github.com/zricethezav/gitleaks/v8/cmd/scm"
-	"github.com/zricethezav/gitleaks/v8/logging"
 	"github.com/zricethezav/gitleaks/v8/report"
 	"github.com/zricethezav/gitleaks/v8/sources"
 )
 
+type RemoteInfo sources.RemoteInfo
```
rgmz, line 11: It's not immediately obvious why this is done. Worth a comment explaining it's there for backwards compatibility.

sources/git.go
```diff
@@ -21,6 +30,27 @@ type GitCmd struct {
...
+// blobReader provides a ReadCloser interface git cat-file blob to fetch
+// a blob from a repo
+type blobReader struct {
+	io.ReadCloser
+	cmd *exec.Cmd
+}
+
+// Close closes the underlying reader and then waits for the command to complete,
+// releasing its resources.
+func (br *blobReader) Close() error {
```
rgmz, line 45: This might need `io.Copy(io.Discard, br)` or it could lead to the process never exiting (https://github.com/trufflesecurity/trufflehog/pull/3375#issuecomment-2396822320). Worth double-checking.

rgmz (commented):
Some additional comments and observations, while I wait for this to chew through a >10GB repository...

sources/git.go
bplaxco: Good call :ok_hand:

sources/file.go
bplaxco: Habit from Python (I prefer your version). :+1:

sources/file.go
bplaxco: Yeah it was intended. Adding a comment :ok_hand:

cmd/root.go
bplaxco: Agreed. I was just matching the style of the other spots in the file, but I'll refactor those as well.

detect/detect.go
bplaxco: Yeah. It was staticcheck QF1001. 

To me the newer version is slightly easier to keep in my head[1], but I'll go ahead and revert and exclude `QF1001` in the golangci-lint to keep things how they were. :ok_hand: 

[1] The `!(...)` means I just have to keep a little more in my head, but it also reads more like normal English I think.

bplaxco:
> > @zricethezav I'm not sure, would you be opposed to me tweaking the diagnostic feature to use net/http/pprof so you can sample the pprof data as needed during the run before it writes out the final dump at the end?
> 
> If you're up for it!

Ended up not being too bad ^_^ https://github.com/gitleaks/gitleaks/pull/1872/commits/2886f7745fd65a4ad0f7173fb93069dcefcfed7a

zricethezav:
Amazing work @bplaxco! (per usual). @rgmz thank you for the help in the review 🙇🏻. I think this PR is in fantastic shape. All the testing, profiling, and manually running gives me confidence this likely won't break anything. If it does, we'll fix'er up real quick.

Not only do we get archive scanning out of this but it's also a huge improvement to the source package. _Sources generate fragments_. Simple as.

Gonna hit the merge button

PR #1883, opened 2025-06-04
Added Heroku API Key New Version

kashifkhan0771:
### Description:
Heroku API Keys Pattern is updated.

### Checklist:

### DIFF
--- a/cmd/generate/config/rules/heroku.go
+++ b/cmd/generate/config/rules/heroku.go
@@ -30,7 +30,8 @@ func HerokuV2() *config.Rule {
 	r := config.Rule{
 		Description: "Detected a Heroku API Key, potentially compromising cloud application deployments and operational security.",
 		RuleID:      "heroku-api-key-v2",
-		Regex:       utils.GenerateUniqueTokenRegex(`\b(HRKU-AA[0-9a-zA-Z_-]{58})\b`, true),
+		Regex:       utils.GenerateUniqueTokenRegex(`(HRKU-AA[0-9a-zA-Z_-]{58})`, false),
+		Entropy:     4,
 		Keywords:    []string{"HRKU-AA"},
 	}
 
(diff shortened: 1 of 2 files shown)


cmd/generate/config/rules/heroku.go
```diff
@@ -24,3 +24,21 @@ func Heroku() *config.Rule {
...
 	return utils.Validate(r, tps, nil)
 }
+
+func HerokuV2() *config.Rule {
+	// define rule
+	r := config.Rule{
+		Description: "Detected a Heroku API Key, potentially compromising cloud application deployments and operational security.",
+		RuleID:      "heroku-api-key-v2",
+		Regex:       utils.GenerateUniqueTokenRegex(`\b(HRKU-AA[0-9a-zA-Z_-]{58})\b`, true),
+		Keywords:    []string{"HRKU-AA"},
```
rgmz, line 34: A few observations:
1. Using `\b` isn't required in conjunction with [`GenerateUniqueTokenRegex`](https://github.com/gitleaks/gitleaks/blob/c5683ca6a3158d340087b88ac5407530f6c8d53c/cmd/generate/config/utils/generate.go#L69). Also, `\b` is never valid beside a character like `-`, on `[a-zA-Z0-9_]`.
2. `GenerateUniqueTokenRegex` also take scare of capture groups, so they don't have to be manually defined.
3. The second parameter should be `false`, otherwise it marks the entire regex as case-insensitive which makes it possible to match invalid strings (`hrku-aa`).
4. A rule should almost always have a minimum entropy threshold.

```suggestion
		Regex:       utils.GenerateUniqueTokenRegex(`HRKU-AA[0-9a-zA-Z_-]{58}`, false),
		Entropy: 4,
		Keywords:    []string{"HRKU-AA"},
```

rgmz (approved):

zricethezav:
nice, thanks @kashifkhan0771 and @rgmz

PR #1905, opened 2025-07-01
Composite rules

zricethezav:
### Description:

Screenshot 2025-07-01 at 3 45 15 PM

### Example composite rule

```toml
[[rules]]
id = "aws-access-key"
description = "AWS Access Key ID - requires secret key and region"
regex = '''AKIA[0-9A-Z]{16}'''
keywords = ["AKIA"]
[[rules.requires]]
id = "aws-secret-key"
within_lines = 20
[[rules.requires]]
id = "aws-region"
within_lines = 20

[[rules]]
id = "aws-secret-key"
description = "AWS Secret Access Key"
regex = '''[A-Za-z0-9/+=]{40}'''
keywords = ["secret", "key"]
skipReport = true

[[rules]]
id = "aws-region"
description = "AWS Region"
regex = '''(us|eu|ap|ca|sa)-(north|south|east|west|central)-[0-9]'''
keywords = ["region", "us-", "eu-", "ap-"]
skipReport = true

```

Fun diagrams for the visual learners (also seen in the readme)
```
    Fragment-level matching:               
    Any auxiliary finding within           
    the fragment boundary                  
    matches.                               
          ┌────────┐                       
   ┌──────┤fragment├─────┐                 
   │      └──────┬─┤     │ ┌───────┐       
   │             │a│◀────┼─│✓ MATCH│       
   │          ┌─┐└─┘     │ └───────┘       
   │┌─┐       │p│        │                 
   ││a│    ┌─┐└─┘        │ ┌───────┐       
   │└─┘    │a│◀──────────┼─│✓ MATCH│       
   └─▲─────┴─┴───────────┘ └───────┘       
     │    ┌───────┐                        
     └────│✓ MATCH│                        
          └───────┘                        
                                           
                                           
   example:                                
   `within_columns = 3`                    
          ┌────────┐                       
   ┌────┬─┤fragment├─┬───┐                 
   │      └──────┬─┤     │ ┌───────────┐   
   │    │        │a│◀┼───┼─│+1C ✓ MATCH│   
   │          ┌─┐└─┘     │ └───────────┘   
   │┌─┐ │     │p│    │   │                 
┌──▶│a│  ┌─┐  └─┘        │ ┌───────────┐   
│  │└─┘ ││a│◀────────┼───┼─│-2C ✓ MATCH│   
│  │       ┘             │ └───────────┘   
│  └── -3C ───0C─── +3C ─┘                 
│  ┌─────────┐                             
│  │ -4C ✗ NO│                             
└──│  MATCH  │                             
   └─────────┘                             
                                           
                                           
   example:                                
   `within_lines = 4`                      
         ┌────────┐                        
   ┌─────┤fragment├─────┐                  
  +4L─ ─ ┴────────┘─ ─ ─│                  
   │                    │                  
   │              ┌─┐   │ ┌────────────┐   
   │         ┌─┐  │a│◀──┼─│+1L ✓ MATCH │   
   0L  ┌─┐   │p│  └─┘   │ ├────────────┤   
   │   │a│◀──┴─┴────────┼─│-1L ✓ MATCH │   
   │   └─┘              │ └────────────┘   
   │                    │ ┌─────────┐      
  -4L─ ─ ─ ─ ─ ─ ─ ─┌─┐─│ │-5L ✗ NO │      
   │                │a│◀┼─│  MATCH  │      
   └────────────────┴─┴─┘ └─────────┘      
                                           
                                           
   example:                                
   `within_lines = 4`                      
   `within_columns = 3`                    
         ┌────────┐                        
   ┌─────┤fragment├─────┐                  
  +4L   ┌└────────┴ ┐   │                  
   │            ┌─┐     │ ┌───────────────┐
   │    │       │a│◀┼───┼─│+2L/+1C ✓ MATCH│
   │         ┌─┐└─┘     │ └───────────────┘
   0L   │    │p│    │   │                  
   │         └─┘        │                  
   │    │           │   │ ┌────────────┐   
  -4L    ─ ─ ─ ─ ─ ─┌─┐ │ │-5L/+3C ✗ NO│   
   │                │a│◀┼─│   MATCH    │   
   └───-3C────0L───+3C┴─┘ └────────────┘   
```

### Question for the community:
1. Thoughts on the terms `required`, `skipReport`, `Auxiliary`?
2. What rules should be updated to be composite rules?
3. Do we need `skipReport`?

### Checklist:

### DIFF
--- a/detect/detect.go
+++ b/detect/detect.go
@@ -588,18 +588,18 @@ func (d *Detector) processRequiredRules(fragment Fragment, currentRaw string, r 
 
 	// Now process each primary finding against the pre-collected required findings
 	for _, primaryFinding := range primaryFindings {
-		var auxiliaryFindings []*report.AuxiliaryFinding
+		var requiredFindings []*report.RequiredFinding
 
 		for _, requiredRule := range r.RequiredRules {
-			requiredFindings, exists := allRequiredFindings[requiredRule.RuleID]
+			foundRequiredFindings, exists := allRequiredFindings[requiredRule.RuleID]
 			if !exists {
 				continue // Rule wasn't found earlier, skip
 			}
 
 			// Filter findings that are within proximity of the primary finding
-			for _, requiredFinding := range requiredFindings {
+			for _, requiredFinding := range foundRequiredFindings {
 				if d.withinProximity(primaryFinding, requiredFinding, requiredRule) {
-					aux := &report.AuxiliaryFinding{
+					req := &report.RequiredFinding{
 						RuleID:      requiredFinding.RuleID,
 						StartLine:   requiredFinding.StartLine,
 						EndLine:     requiredFinding.EndLine,
@@ -609,22 +609,22 @@ func (d *Detector) processRequiredRules(fragment Fragment, currentRaw string, r 
 						Match:       requiredFinding.Match,
 						Secret:      requiredFinding.Secret,
 					}
-					auxiliaryFindings = append(auxiliaryFindings, aux)
+					requiredFindings = append(requiredFindings, req)
 				}
 			}
 		}
 
 		// Check if we have at least one auxiliary finding for each required rule
-		if len(auxiliaryFindings) > 0 && d.hasAllRequiredRules(auxiliaryFindings, r.RequiredRules) {
+		if len(requiredFindings) > 0 && d.hasAllRequiredRules(requiredFindings, r.RequiredRules) {
 			// Create a finding with auxiliary findings
 			newFinding := primaryFinding // Copy the primary finding
-			newFinding.AddAuxiliaryFindings(auxiliaryFindings)
+			newFinding.AddRequiredFindings(requiredFindings)
 			finalFindings = append(finalFindings, newFinding)
 
 			logger.Debug().
 				Str("primary-rule", r.RuleID).
 				Int("primary-line", primaryFinding.StartLine).
-				Int("auxiliary-count", len(auxiliaryFindings)).
+				Int("auxiliary-count", len(requiredFindings)).
 				Msg("multi-part rule satisfied")
 		}
 	}
@@ -633,7 +633,7 @@ func (d *Detector) processRequiredRules(fragment Fragment, currentRaw string, r 
 }
 
 // hasAllRequiredRules checks if we have at least one auxiliary finding for each required rule
-func (d *Detector) hasAllRequiredRules(auxiliaryFindings []*report.AuxiliaryFinding, requiredRules []*config.Required) bool {
+func (d *Detector) hasAllRequiredRules(auxiliaryFindings []*report.RequiredFinding, requiredRules []*config.Required) bool {
 	foundRules := make(map[string]bool)
 	// AuxiliaryFinding
 	for _, aux := range auxiliaryFindings {
--- a/report/finding.go
+++ b/report/finding.go
@@ -51,10 +51,10 @@ type Finding struct {
 	Fragment *sources.Fragment `json:",omitempty"`
 
 	// TODO keeping private for now to during experimental phase
-	auxiliaryFindings []*AuxiliaryFinding
+	requiredFindings []*RequiredFinding
 }
 
-type AuxiliaryFinding struct {
+type RequiredFinding struct {
 	// contains a subset of the Finding fields
 	// only used for reporting
 	RuleID      string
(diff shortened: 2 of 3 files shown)


bplaxco:
> Thoughts on the terms required, skipReport, Auxiliary?

- `skipReport` - sounds good to me! :+1: 
- `Auxiliary` - I'd prefer `RelatedFindings` but I'm not totally against Auxiliary. ^\_^ Rationale: 
  - If `skipReport = False` then they're not really additional findings, they just happen to be related.
  - If there are more ways to relate the findings in the future, then we already have a place to put them and we can just note the relationship in the report somewhere. 
 - `requires` - I can also dig it :ok_hand: the only other thing that would come to mind would be something like this, but I think it's probably scope creep so I'm not actually for it:
   ```toml
   [[rules.related]]
      id = "aws-secret-key"
      withinLines = 20
      required = true
   ```

> Do we need skipReport?

I think it's great and will be really helpful for one of my use cases ^\_^. In v9 I would like to see `Secret` -> `Target` (or something similar) on finding since some of the findings from these compost rules might pull in things that aren't actually secrets they're just related information needed to verify the secret.

config/config.go
```diff
@@ -153,6 +164,23 @@ func (vc *ViperConfig) Translate() (Config, error) {
...
+			}
+			requiredRule := Required{
+				RuleID:        r.ID,
+				WithinLines:   r.WithinLines,
+				WithinColumns: r.WithinColumns,
+				// Distance: r.Distance,
+			}
+			// if _, ok := rulesMap[r.RuleID]; !ok {
+			// 	return Config{}, fmt.Errorf("%s: [[rules.required]] rule ID '%s' does not exist", cr.RuleID, r.RuleID)
+			// }
```
bplaxco, line 180: I kinda like this check. Any reason for commenting it out?

detect/detect.go
```diff
@@ -360,6 +360,10 @@ func (d *Detector) detectRule(fragment Fragment, newlineIndices [][]int, current
 		}()
 	)
 
+	if r.SkipReport == true && !fragment.InheritedFromFinding {
```
bplaxco, line 363:
```suggestion
	if r.SkipReport && !fragment.InheritedFromFinding {
```

report/finding.go
```diff
@@ -68,3 +98,29 @@ func maskSecret(secret string, percent uint) string {
 
 	return secret[:lth] + "..."
 }
+
+func (f *Finding) PrintAuxiliaryFindings() {
```
bplaxco, line 102: Is this here instead of util because `auxiliaryFindings` is private? Would it make sense for the all the `Print*` functions to go in report since it seems like they're more focused on reporting info back to the user instead of the detection itself?

detect/detect.go
```diff
@@ -534,7 +538,145 @@ func (d *Detector) detectRule(fragment Fragment, newlineIndices [][]int, current
...
+			logger.Error().Str("rule-id", requiredRule.RuleID).Msg("required rule not found in config")
+			continue
+		}
+
+		// Mark fragment as inherited to prevent infinite recursion
+		inheritedFragment := fragment
+		inheritedFragment.InheritedFromFinding = true
+
+		// Call detectRule once for each required rule
+		requiredFindings := d.detectRule(inheritedFragment, newlineIndices, currentRaw, rule, encodedSegments)
```
bplaxco, line 573: If two rules have the same required rule, would `d.detectRule` get called twice for that rule?

Would something like this save extra scans and passes to see if the reqs are met? (doing it in python/pseudo code):

```python
def detect_with_reqs(rule:Rule, fragment:Fragement, findings_by_rule:dict[str,list[Finding]], to_report:list[Finding]) -> list[Finding]:
    # Already processed just return the findings
    if rule.id in findings_by_rule:
        return findings_by_rule[rule.id]

    # Run detections for this rule
    findings = detect(rule, fragment, ...)

    # If there's nothing left to do, add the findings to the report set and
    # return the findings
    if not findings or not rule.requires:
        if not rule.skip_report:
            to_report.extend(findings)
        return findings

    # Handle the requirements
    for req in rule.requires:
        req_rule = rules.config[req.id]

        # This handles the req tree recursively
        req_findings = detect_with_reqs(req_rule, fragment, findings_by_rule, to_report)

        # No need to go further if a requirement didn't have findings
        if not req_findings:
            return []

        # Now check if the reqs' findings meet the criteria
        for finding in findings:
            # Filter the req_findings to the one that meet the proximity requirement
            req_findings = within_proximity(req, finding, req_findings)

            # No need to go further given a requirement wasn't met
            if not req_findings:
                return []

            # Requirement met! add the findings
            finding.add_aux_findings(req_findings)

    return findings


def detect(fragment):
    to_report = []
    findings_by_rule = {}

    for rule in config.rules.values():
        # These cases will be handled by detect_rule
        if rule.skipReport:
            continue

        _ = detect_with_reqs(rule, fragment, findings_by_rule, to_report)

    return to_report
```

config/config.go
```diff
@@ -58,6 +62,12 @@ type ViperConfig struct {
 	EnableExperimentalAllowlistOptimizations bool
 }
 
+type viperRequired struct {
+	ID            string
+	WithinLines   *int `mapstructure:"within_lines"`
+	WithinColumns *int `mapstructure:"within_columns"`
```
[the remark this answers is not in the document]
zricethezav: ah shoot, you're right. Looks like we have `regexTarget` not `regex_target`. Consistency wins this one, camel case it is

config/config.go
zricethezav: It needs to happen _after_ all the rules have been seen otherwise it'll flag an err

detect/detect.go
zricethezav: This should signal to folks that this code is 100% handcrafted by human hands and not generated by an LLM

zricethezav:
whelp in she goes

PR #1750, opened 2025-02-11
Feat/Passing multiple arguments to dir

recreator66:
### Description:

Closes #1727 

### Checklist:

* [x] Does your PR pass tests?
* [ ] Have you written new tests for your changes?
* [x] Have you lint your code locally prior to submission?

Issue #1727, opened 2025-01-25
Passing multiple arguments to `dir` causes a fill scan of the current directory


det:
**Describe the bug**
Passing multiple arguments to `dir` causes a fill scan of the current directory

**To Reproduce**
`gitleaks dir path/to/file` results in a scan of `path/to/file`
`gitleaks dir path/to/file other/path/to/file2` results in a scan of `.`

**Expected behavior**
Either a full scan of `path/to/file` and `other/path/to/file2` or an error indicating multiple arguments is not allowed.
Ideally, I would like a way to scan multiple files in a single invocation of gitleaks. Thank you!

cc @zricethezav

PR #1918, opened 2025-07-18
Fix pre-commit hook by use git-agnostic file-based checking 

Holzhaus:
### Description:

This fixes gitleaks invocations via pre-commit when there are no changed to-be-committed (i.e., via `pre-commit run --files` or `pre-commit run --all-files`). Currently, these will pass even if there would be gitleaks findings because `gitleaks git` checks the diff only.

When called via pre-commit framework, gitleaks does not need to git-aware because the framework already detects which files need to be checked (either the diff when a commit range or changes that are about-to-be-committed are checked, or the files passed to the `--files` command line argument).

Fixes https://github.com/gitleaks/gitleaks/issues/1409.

### Checklist:

* [x] Does your PR pass tests? - unit tests are unaffected, manual testing of pre-commit hook passed
* [x] Have you written new tests for your changes? - not applicable
* [x] Have you lint your code locally prior to submission? - passed my yaml formatter

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,15 +1,18 @@
 - id: gitleaks
   name: Detect hardcoded secrets
   description: Detect hardcoded secrets using Gitleaks
-  entry: gitleaks dir --redact --verbose
+  entry: gitleaks dir
+  args: [--redact, --verbose]
   language: golang
 - id: gitleaks-docker
   name: Detect hardcoded secrets
   description: Detect hardcoded secrets using Gitleaks
-  entry: zricethezav/gitleaks dir --redact --verbose
+  entry: zricethezav/gitleaks dir
+  args: [--redact, --verbose]
   language: docker_image
 - id: gitleaks-system
   name: Detect hardcoded secrets
   description: Detect hardcoded secrets using Gitleaks
-  entry: gitleaks dir --redact --verbose
+  entry: gitleaks dir
+  args: [--redact, --verbose]
   language: system


UnknownPlatypus:
> Is there an ETA for https://github.com/gitleaks/gitleaks/issues/1727?

No Idea, I'm waiting for it to solve the pre-commit use case but I'm not a maintainer nor have enough go knowledge to help with the review. 

> Because it is somewhat unexpected that pre-commit run --all-files tell you that you're fine when you actually have a lot of hardcoded secrets that are just not detected due to https://github.com/gitleaks/gitleaks/issues/1409. From my perspective, I'd rather live with the performance regression than with the risk of undetected secrets.

Unfortunately, this would cause very huge latency for the most common use case of this hook ie running before a commit on changed files. For exemple on my repo, it would make the `pre-commit` hook completely unusable.

```bash
$ hyperfine -i "gitleaks dir --redact --verbose" "gitleaks git --pre-commit --redact --staged --verbose"

Benchmark 1: gitleaks dir --redact --verbose
  Time (mean ± σ):     30.943 s ±  0.527 s    [User: 179.861 s, System: 9.130 s]
  Range (min … max):   30.232 s … 31.886 s    10 runs
 
  Warning: Ignoring non-zero exit code.
 
Benchmark 2: gitleaks git --pre-commit --redact --staged --verbose
  Time (mean ± σ):     342.3 ms ±   3.7 ms    [User: 330.6 ms, System: 25.2 ms]
  Range (min … max):   337.6 ms … 350.6 ms    10 runs
 
Summary
  'gitleaks git --pre-commit --redact --staged --verbose' ran
   90.40 ± 1.82 times faster than 'gitleaks dir --redact --verbose'

```

I completely agree with you this is very unexpected that `pre-commit run --all-files` tell you that you're fine but slowing the hook too much will make people disable it which is not really better imo. Maybe you could try to ping a maintainer in #1750, they might have simply missed the PR.

Issue #1888, opened 2025-06-05
gitleaks dir [path] argument does not allow multiple values or globs


ssbarnea:
**Describe the bug**

It seems that `gitleaks dir` argument 'path' does not allow multiple values or use of globs.

When passing more than one argument, `gitleaks dir` silently tries to run on `.`, which makes this count more as a bug than a feature request.

**To Reproduce**

```
gitleaks dir foo bar

# more advanced usage with patterns
gitleaks dir 'foo/**/*.log' bar
```

**Expected behavior**

- Be able to allow multiple path arguments
- Be able to expand globs, or at least to fail if globs are passed

**Additional context**

The tool itself should be able to expand globs and this is because otherwise its usage cross platform would be highly impacted as not all shells are able to expand the same way. Many other similar tools do have their own glob processing implemented in order to make them more portable.

cc @zricethezav

PR #1238, opened 2023-07-24
Add include files

rajatmehra2307:
### Description:
This PR adds the ability to scan specific files inside the source directory. #1234 

### Checklist:

* [x] Does your PR pass tests?
* [x] Have you written new tests for your changes?
* [x] Have you lint your code locally prior to submission?

Issue #1234, opened 2023-07-18
Adding the ability to scan selected files from the source folder


rajatmehra2307:
**Is your feature request related to a problem? Please describe.**
When running  GitLeaks  with --no-git option and specifying the source folder, it scans every file in the folder. There should be a way of specifying a list of files inside the source folder to include in the scan.

**Describe the solution you'd like**
Adding a new a flag --include-files where the user can specify a comma separated list of file names and GitLeaks should only run the scan for those files. By default if nothing is specified, it should run for every file inside the source folder. 

**Describe alternatives you've considered**
The only other option right now to achieve this is to run GitLeaks for every file within the desired set of files using the --source flag and specifying the filename individually. But this takes a lot of time.

cc @zricethezav

zricethezav:
Thanks for opening this issue but I do not want to add another option to configure something that can be done with `allowlists`. You can use a custom gitleaks configuration file with a global allowlist specifying files you do not want scanned.

mkillianey:
As a user in a monorepo, I have a similar situation, and I understand why you don't want to add a new flag -- but could the semantics of the existing `--source` flag be expanded to allow it to be specified multiple times?

zricethezav:
@mkillianey that's a good solution! I'd happily review a PR for that change

PR #1936, opened 2025-08-27
Allow multiple sources for `dir` command

maxvt:
### Description:
My use for this is for feeding gitleaks an explicit list of changed files for continuous scanning of pull requests to our repository.

This has been requested in #1888, #1727, #1234, and possibly others.

This has previously been implemented in #1238 but for a much older version and it seemed easier to re-do this rather than to rebase

There are some "source"-related configuration processing tasks done in `initConfig()` and `Detector()`; for simplicity, I chose to require explicit configuration of those items (rather than guessing per source, having to distinguish files vs dirs, handling different configuration and ignore files etc.)

I also didn't really understand the logic about figuring out the relative baseline path with respect to the source, and how should that work when multiple sources may be present. It seems easier to simply disable the baseline in a "multiple source" scenario.

### Checklist:

* [-] Have you written new tests for your changes? - no, there are no tests for CLI commands, and since all of these are private, I'm not sure how those tests could be added without a substantial refactoring.

### DIFF
--- a/cmd/directory.go
+++ b/cmd/directory.go
@@ -22,17 +22,27 @@ var directoryCmd = &cobra.Command{
 	Run:     runDirectory,
 }
 
-func runDirectory(cmd *cobra.Command, args []string) {
-	// grab source
-	source := "."
-	if len(args) == 1 {
-		source = args[0]
-		if source == "" {
-			source = "."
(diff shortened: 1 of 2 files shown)


Holzhaus:
Might be a duplicate of #1750.

Issue #1932, opened 2025-08-18
Creating Composite Rule Errors


henrysachs:
**Describe the bug**

I tried to define a new composite rule using code but the gitleaks generator always throws the following error: 

```bash
4:02PM ERR required rule not found in config path= rule-id=anthropic-api-key
```

I can add any rule in ruleId and get the same result so i tried it with a rule that is initialized before the aws one. 

**To Reproduce**

```go
	// define rule
	r := config.Rule{
		RuleID:      "aws-access-token",
		Description: "Identified a pattern that may indicate AWS credentials, risking unauthorized cloud resource access and data breaches on AWS platforms.",
		Regex:       regexp.MustCompile(`\b((?:A3T[A-Z0-9]|AKIA|ASIA|ABIA|ACCA)[A-Z2-7]{16})\b`),
		Entropy:     3,
		RequiredRules: []*config.Required{
			{RuleID: "anthropic-api-key", WithinLines: &maxLines},
		},
		Keywords: []string{
			// https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_identifiers.html#identifiers-unique-ids
			"A3T",  // todo: might not be a valid AWS token
			"AKIA", // Access key
			"ASIA", // Temporary (AWS STS) access key
			"ABIA", // AWS STS service bearer token
			"ACCA", // Context-specific credential
		},
		Allowlists: []*config.Allowlist{
			{
				Regexes: []*regexp.Regexp{
					regexp.MustCompile(`.+EXAMPLE$`),
				},
				StopWords: []string{
					"AKIAIOSFODNN7EXAMPLE",
					"AKIA222222222EXAMPLE",
				},
			},
		},
	}
```

**Expected behavior**
A Composite Rule gets created

**Screenshots**
-

**Additional context**

cc @zricethezav

zricethezav:
@henrysachs thanks for creating an issue for this. Basically we need to update the generator to support this I just havent found time too. I'll try to get to it soon, but if someone wants to take a stab at it before, feel free!

alephnull-sh:
Took a smaller pass here: #2098.

I missed #1940 earlier. This keeps the validation helper backward-compatible, adds the missing `[[rules.required]]` output, and includes tests for both pieces.

PR #1940, opened 2025-09-05
create composite rule error

moamlrh:
### Description:
I know there are a lot of files changed, but the core change is small, most of it is due to updating the `utils.Validate()` method signature

This is a fix for #1932 issue, where creating a composite rule using `RequiredRules` fails.  
when `detectRule` calls `processRequiredRules`, it cant locate the required rules in the config, causing the validation to fail and leading to incorrect or missing findings (_wrong FPs/TPs_).

**My Fix:**
After initializing all rule objects, `utils.Validate()` is called on each rule. with passing required rules (the full rule objects), and on creating the **detector** (`createSingleRuleDetector()`), required rules are used to populate the config rules.

**note:** This is a simple working fix. if there’s a better way to pass the full rules data (or another implementation), feel free to suggest changes or just close this PR.

### Checklist:
* [x] Does your PR pass tests? ✅
* [ ] Have you written new tests for your changes? ❌
* [x] Have you lint your code locally prior to submission? ✅

### DIFF
--- a/cmd/generate/config/main.go
+++ b/cmd/generate/config/main.go
@@ -257,7 +257,7 @@ func main() {
 	// ensure rules have unique ids
 	ruleLookUp := make(map[string]config.Rule, len(configRules))
 	for _, rule := range configRules {
-		requiredRules := make([]*config.Rule, len(rule.RequiredRules))
+		requiredRules := make([]*config.Rule, 0, len(rule.RequiredRules))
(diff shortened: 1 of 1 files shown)


henrysachs:
I will really appreciate if this would be working but can understand that the code has a lot of changes. but as the author stated its really just because of the newly introduced fields. @zricethezav any thoughts?

PR #1992, opened 2025-11-19
Add URL & JSON sources

bplaxco:
Add support for scanning URLs directly and JSON data.

### Description:

- If the URL returns an application/json type then treat the response as JSON data to scan.
- Support scanning URLs included in the JSON data where:
  - The full value of the field is the URL
  - AND the path to the URL is in the  `FetchURLPatterns` list.

**FetchURLPatterns**

This is a list of glob and recursive glob (i.e. `**`) patterns that can be passed in when fetching content 

**Paths and Path Separators**

The path returned for items in a JSON scan look like file paths:

For example, given:

```json
{
   "some": [{}, {"target": "secret"}]
}
```

The path to "secret" would be:

```
some/1/target
```

If a URL returns JSON or a URL is fetched from JSON the `!` inner path separator is used to indicate it crossed the bound from one source to another.

**Extra**

I also added an `import` target to the Makefile to run goimports to auto import and tidy in the project as needed to make it easier for folks with really minimal editor setups ( :raised_hand_with_fingers_splayed: ) ^\_^
    
### To-do

* [x] Add the sources
* [x] Add `json` and `url` commands
* [ ] Test/discuss changes (**WIP**)
* [ ] Update docs
 
### Checklist:

### DIFF
--- a/cmd/json.go
+++ b/cmd/json.go
@@ -12,18 +12,17 @@ import (
 )
 
 const fetchURLGlobDescription = `
-scan URL values in the JSON in matching paths:
-- the full value must be a URL (i.e. URLs located just somewhere in a string will be skipped)
-- currently only HTTP(S) URLs and GET requests are supported
-- paths are formatted like file paths where keys and array indices are path elements
-- finding paths will be reported as the path to the URL in the JSON
-- * acts as a wildcard within single path elements (i.e. '[^/]*' in regex)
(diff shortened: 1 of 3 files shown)


bplaxco:
@zricethezav && @rgmz before I give this a solid testing and write docs, let me know your thoughts and if you'd like to see anything head in a different direction. :ok_hand:

PR #2012, opened 2025-12-30
fix: Configuration with empty [[rules]] array fails to load in v8.29.0

### LINKED ISSUE #1985: Configuration with empty [[rules]] array fails to load in v8.29.0
**Describe the bug**
Starting with v8.29.0, gitleaks fails to load configuration files that contain an empty `[[rules]]` section. This worked correctly in v8.28.0.

**To Reproduce**

1. Create a .gitleaks.toml configuration file with the following content:
```toml
title = "Custom Gitleaks configuration"
[extend]
useDefault = true
[[rules]]
```

2. Run gitleaks v8.28.0:
```shell
$ docker run -v $(pwd):/path ghcr.io/gitleaks/gitleaks:v8.28.0 dir /path --config /path/.gitleaks.toml --no-banner
7:06AM INF scanned ~2375432 bytes (2.38 MB) in 460ms
7:06AM WRN leaks found: 117
```

3. Run gitleaks v8.29.0 with the same configuration:
```shell
$ docker run -v $(pwd):/path ghcr.io/gitleaks/gitleaks:v8.29.0 dir /path --config /path/.gitleaks.toml --no-banner
7:06AM FTL Failed to load config error="rule |id| is missing or empty"
```

**Expected behavior**
An empty [[rules]] section should be treated as valid, allowing the scan to proceed using only the default rules. The behavior should be consistent with v8.28.0.

 - OS: macOS 26.1
 - Gitleaks Version: v8.29.0 (broken), v8.28.0 (working)

cc @zricethezav

majiayu000:
## Summary
This PR fixes #1985

## Changes
- Skip completely empty rule entries during config parsing (e.g., from an empty `[[rules]]` in TOML)
- Add test case to verify empty `[[rules]]` entries are handled correctly without causing errors
- The fix checks if all rule fields are at their zero/default values before processing, allowing users to have empty `[[rules]]` blocks in their configuration files

## Test plan
- [x] Added `TestTranslateEmptyRules` test case
- [x] All existing tests pass
- [x] Verified the test config file `testdata/config/valid/empty_rules.toml` loads successfully

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -116,6 +116,14 @@ func (vc *ViperConfig) Translate() (Config, error) {
 
 	// Validate individual rules.
 	for _, vr := range vc.Rules {
+		// Skip completely empty rule entries (e.g., from an empty [[rules]] in TOML).
+		if vr.ID == "" && vr.Description == "" && vr.Path == "" && vr.Regex == "" &&
+			vr.SecretGroup == 0 && vr.Entropy == 0 && len(vr.Keywords) == 0 &&
+			len(vr.Tags) == 0 && vr.AllowList == nil && len(vr.Allowlists) == 0 &&
+			len(vr.Required) == 0 && !vr.SkipReport {
+			continue
+		}
+
 		var (
 			pathPat  *regexp.Regexp
 			regexPat *regexp.Regexp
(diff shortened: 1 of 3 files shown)


rgmz:
The original described issue is invalid, so no fix is required.

PR #2080, opened 2026-04-10
fix(sources): propagate git stderr errors instead of swallowing them

### LINKED ISSUE #1981: git errors are being incorrectly ignored
**Describe the bug**

When the `git` command called indirectly by `gitleaks git` or `gitleaks detect` fails, its exit code is ignored and `gitleaks` exists with `0` as if no issues happened.

This is easy to reproduce by running on a directory that is not a git repo or passing invalid git options to `--log-opts`.

**To Reproduce**

Try `gitleaks git --log-level=debug` on an empty dir or `gitleaks git --log-opts="--invalid" --verbose`.

**Expected behavior**

The exit code should NOT be 0.

**Screenshots**

```sh
► gitleaks git --log-opts="--invalid" --verbose

    ○
    │╲
    │ ○
    ○ ░
    ░    gitleaks

9:21AM ERR [git] fatal: not a git repository (or any of the parent directories): .git
9:21AM ERR error="stderr is not empty"
9:21AM INF 0 commits scanned.
9:21AM INF scanned ~0 bytes (0) in 184ms
9:21AM INF no leaks found

► echo $?
0

► gitleaks git --log-level=debug

    ○
    │╲
    │ ○
    ○ ░
    ░    gitleaks

9:24AM DBG using github.com/wasilibs/go-re2 regex engine
9:24AM DBG no gitleaks config found in path .gitleaks.toml, using default gitleaks config
9:24AM DBG executing: /usr/bin/git -C . log -p -U0 --full-history --all --diff-filter=tuxdb
9:24AM ERR [git] fatal: not a git repository (or any of the parent directories): .git
9:24AM DBG skipping finding links: repository has no configured remote.
9:24AM ERR error="stderr is not empty"
9:24AM INF 0 commits scanned.
9:24AM DBG Note: this number might be smaller than expected due to commits with no additions
9:24AM INF scanned ~0 bytes (0) in 155ms
9:24AM INF no leaks found

► echo $?
0

► /usr/bin/git -C . log -p -U0 --full-history --all --diff-filter=tuxdb
fatal: not a git repository (or any of the parent directories): .git

► echo $?
128
```

 - OS: `Ubuntu 24.04.3 LTS x86_64`
 - Gitleaks Version: `8.28.0`

**Additional context**

- This was initially reported on https://github.com/gitleaks/gitleaks/pull/1461
- It may have been a regression introduced at https://github.com/gitleaks/gitleaks/pull/1872

cc @zricethezav @rgmz @bplaxco

Smeet23:
## Problem

When `gitleaks git` (or `gitleaks detect`) runs in a directory that is not a git repository (or git fails for any reason), gitleaks exits with code `0` and prints `no leaks found` — as if the scan succeeded.

```sh
$ gitleaks git --log-opts="--invalid" --verbose
...
ERR [git] fatal: not a git repository
ERR error="stderr is not empty"
INF no leaks found
$ echo $?
0
```

## Root cause

`sources.Git.Fragments()` sends git stderr errors through `yield(Fragment{}, err)`. The fragment callback in `detect.DetectSource` is designed to handle per-fragment errors gracefully — it logs them and returns `nil`. This means the git failure error was swallowed, `DetectSource` returned `nil`, and the command exited cleanly with `0`.

## Fix

Return the git error directly from `Fragments()` instead of routing it through the yield callback. This ensures the error propagates to `DetectSource` and then to the caller, which already calls `os.Exit(1)` when `err != nil`.

```go
// Before
return yield(Fragment{}, err)

// After
return err
```

## Test plan

- [ ] `TestGitFragments_ErrorPropagation`: runs `NewGitLogCmd` on a temp dir that is not a git repo, calls `Fragments()`, and asserts a non-nil error is returned

Fixes #1981

### DIFF
--- a/sources/git.go
+++ b/sources/git.go
@@ -417,7 +417,10 @@ func (s *Git) Fragments(ctx context.Context, yield FragmentsFunc) error {
 				break
 			}
 
-			return yield(Fragment{}, err)
+			// Return the git error directly so it propagates up through DetectSource
+			// and the caller can exit with a non-zero exit code. Previously this went
+			// through yield() where the fragment callback swallowed it.
+			return err
 		}
 	}
 
--- a/sources/git_test.go
+++ b/sources/git_test.go
@@ -1,6 +1,48 @@
 package sources
 
-// TODO: commenting out this test for now because it's flaky. Alternatives to consider to get this working:
+import (
+	"context"
+	"os"
+	"testing"
+
+	"github.com/fatih/semgroup"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/zricethezav/gitleaks/v8/config"
+)
+
+// TestGitFragments_ErrorPropagation verifies that a git failure (e.g. running
+// in a directory that is not a git repository) propagates as a non-nil error
+// from Fragments() rather than being silently swallowed. Regression test for
+// https://github.com/gitleaks/gitleaks/issues/1981.
+func TestGitFragments_ErrorPropagation(t *testing.T) {
+	// Use a temp dir that is not a git repository.
+	dir, err := os.MkdirTemp("", "gitleaks-no-git-*")
+	require.NoError(t, err)
+	defer os.RemoveAll(dir)
+
+	ctx := context.Background()
+
+	gitCmd, err := NewGitLogCmdContext(ctx, dir, "")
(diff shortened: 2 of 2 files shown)


Smeet23:
Hi team — just a gentle ping on this PR. All checks are passing. Would appreciate a review when you get a chance. Happy to address any feedback. Thanks!

PR #2101, opened 2026-05-05
feat: Add secret-value support to .gitleaksignore

rossguilbert-gov:
### Description:

This PR adds support for secret-value based entries in `.gitleaksignore` alongside the existing fingerprint-based format.

Today `.gitleaksignore` only supports fingerprints. That makes ignores brittle when a file changes but the same false positive remains present. This change adds a more stable ignore mechanism by allowing users to ignore findings by exact extracted `finding.Secret` value, with optional `path` and/or `rule` scoping to reduce blast radius.

Supported entry formats:

- `secret:<value>`
- `rule:<rule-id>:secret:<value>`
- `path:<normalized-path>:secret:<value>`
- `path:<normalized-path>:rule:<rule-id>:secret:<value>`

### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -517,7 +517,20 @@ class CustomClass:
 
 #### .gitleaksignore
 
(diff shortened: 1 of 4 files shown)


thomashorrobin:
great PR, please merge

PR #2124, opened 2026-05-18
config: switch the bundled gitleaks.toml to [[allowlists]]

### LINKED ISSUE #2120: fix: update default gitleaks.toml to use [[allowlists]] instead of deprecated [allowlist]
The bundled config/gitleaks.toml at line 22 still uses the deprecated [allowlist] syntax. The maintainer already flagged this with a # TODO: change to [[allowlists]] comment in the file. The [[allowlists]] form has been current since v8.25.0.
This also affects the config generator, it should emit [[allowlists]] going forward (related to #2048).
I'd like to work on this fix if it's welcome.

cc @zricethezav

c-tonneslan:
Closes #2120.

Picks up the `# TODO: change to [[allowlists]]` next to the global block and also updates the config generator template so the next regeneration doesn't reintroduce the deprecated `[allowlist]` form. Both `[allowlist]` and `[[allowlists]]` still parse, but the latter has been the documented shape since v8.25.0.

### DIFF
--- a/cmd/generate/config/rules/config.tmpl
+++ b/cmd/generate/config/rules/config.tmpl
@@ -18,7 +18,7 @@ title = "{{.Title}}"
 # config-enabled features are guaranteed to work.
 minVersion = "v8.25.0"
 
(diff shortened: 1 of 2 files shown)


zricethezav:
Thanks for opening the PR. I may review it in the future but for now my focus has shifted to Betterleaks, which you can find here https://github.com/betterleaks/betterleaks

PR #2169, opened 2026-06-13
fix(detect): surface git stderr errors instead of swallowing them (#2129)

### LINKED ISSUE #2129: `gitleaks git` silently reports "no leaks found" with exit 0 when `git log` writes anything to stderr (git-crypt repos affected by default)
### Summary

`gitleaks git` treats **any** output on `git log`'s stderr as a fatal abort,
then prints `0 commits scanned`, `no leaks found`, and **exits 0**. This is a
silent false negative: a user running `gitleaks git` in CI or at the command
line sees a clean report when in fact nothing was scanned.

This is the same class of issue as #1450 (invalid `--log-opts`) and #1455
(broken repo), but it bites a much broader population: **any repository that
uses [git-crypt](https://github.com/AGWA/git-crypt) triggers it by default,
with no misconfiguration on the user's part**. The git-crypt `diff.textconv`
filter writes a multi-line `Warning: file not encrypted` to stderr for every
historical commit that touches a now-encrypted path. That's enough to abort
`gitleaks git`.

### Versions

- gitleaks **v8.30.1** (latest as of 2026-05-21)
- git-crypt 0.7.0 (relevant only as a stderr source)
- Reproduced on Windows 11 with Git for Windows 2.x

### Reproduction

In any git-crypt repo with at least one encrypted-path commit in history:

```bash
gitleaks git
```

Observed output (truncated):

```
ERR [git] git-crypt: Warning: file not encrypted
ERR [git] git-crypt: Run 'git-crypt status' to make sure ...
... (repeated per commit) ...
ERR  error="stderr is not empty"
INF 0 commits scanned.
INF scanned ~2622506 bytes (2.62 MB) in 2.22s
INF no leaks found
$ echo $?
0
```

Note the `0 commits scanned` and the `0` exit code. A user who only reads the
final `no leaks found` line believes the repo is clean.

### Workaround (works, but should not be necessary)

```bash
git log -p --all 2>/dev/null | gitleaks stdin
```

This bypasses gitleaks' internal `git log` invocation and discards stderr,
producing a real scan (`~2.81 MB` vs `~2.62 MB` on the broken path in our
repo; `commits scanned` is no longer surfaced because `stdin` doesn't track
that, but the diff stream is fully consumed).

### Suggested fixes (any one would close the silent-false-negative class)

1. When `git log -p` produces no commits **and** stderr is non-empty, exit
   non-zero (configurable via `--exit-code` already, but it isn't honored
   in this path).
2. Make `error="stderr is not empty"` a hard error by default, with an
   opt-in `--allow-git-stderr` flag for users who knowingly run in noisy
   contexts (git-crypt, hooks, custom textconv filters).
3. Distinguish "git log spoke to stderr while still streaming commits"
   (warning) from "git log spoke to stderr and produced zero commits"
   (fatal).

The current behavior conflates a security-relevant zero-commits-scanned
condition with a successful clean scan. Either of the first two options
would fix that without breaking the noisy-stderr-but-still-scanning case.

### Related issues

- #1450: Invalid `--log-opts` disregards git log exit code and silently does nothing at all
- #1455: Scan for broken git repo

This issue is narrower than either of those in that the trigger is a
common, non-broken configuration (git-crypt is widely used for repo-versioned
secrets), but the underlying defect is the same: stderr-driven abort + exit 0
+ "no leaks found".

### Why this matters

git-crypt is one of the standard answers to "how do I version-control secrets
in a private repo?" Running `gitleaks git` on such a repo is a routine
defense-in-depth step. A silent false negative here is exactly the
threat model gitleaks exists to defend against.

acebot712:
## Description

Fixes #2129.

When the git source runs `git log` and git writes to stderr, `listenForStdErr` (in `sources/git.go`) flips any non-allowlisted line into a fatal `"stderr is not empty"` error, yielded with an **empty fragment**. `DetectSource`'s yield callback treated that the same as a recoverable per-fragment error — `logger.Error(); return nil` — so the error was swallowed. The scan aborted but reported `no leaks found` and exited **0**.

This is a silent false negative, and it triggers **by default on any [git-crypt](https://github.com/AGWA/git-crypt) repo**: git-crypt's `diff.textconv` filter writes `Warning: file not encrypted` to stderr for historical commits touching now-encrypted paths. Running `gitleaks git` on such a repo prints a clean report while scanning nothing.

## Fix

`DetectSource` now propagates **source-level** errors (yielded with an empty fragment, which is uniquely how the git source signals a fatal abort — every other error-yield carries a populated fragment) instead of swallowing them. The existing partial-scan path in `findingSummaryAndExit` then runs: it logs `partial scan completed` and exits non-zero. Recoverable **per-fragment** errors (populated fragment, e.g. one unreadable file in a directory scan) are still logged and skipped, so the rest of the scan continues.

The whole partial-scan/exit-1 machinery already existed — it was simply never reached because the error was eaten one layer up. One targeted change unblocks it.

## Before / after (real repo with a stderr-writing textconv filter)

```
# before — silent false negative
INF 0 commits scanned.
INF no leaks found
$ echo $?  → 0

# after — abort is surfaced
INF 1 commits scanned.
ERR failed to scan Git repository error="stderr is not empty"
WRN partial scan completed
WRN no leaks found in partial scan
$ echo $?  → 1
```

## Tests

Added `TestDetectSourceErrorPropagation` in `detect/detect_test.go`:
- source-level error (empty fragment) → `DetectSource` returns an error (**fails before this fix, passes after**)
- per-fragment error (populated fragment) → still swallowed (returns nil), guarding against over-propagation that would abort directory scans
- clean scan → no error

`go test ./detect/`, `go build ./...`, `gofmt`, and `go vet` (no new findings) all pass.

## Note

This restores the *intended* behavior (the `"stderr is not empty"` error and the partial-scan path already exist). If you'd rather make noisy-but-non-fatal stderr opt-in via a flag (issue #2129 option 2), happy to follow up — I kept this PR to the minimal correctness fix.

---

I used Claude Code to help locate the root cause and write the test; I reproduced the bug end-to-end and verified the before/after behavior myself.

### DIFF
--- a/detect/detect.go
+++ b/detect/detect.go
@@ -226,8 +226,17 @@ func (d *Detector) DetectSource(ctx context.Context, source sources.Source) ([]r
 		logger := logContext.Logger()
 
 		if err != nil {
-			// Log the error and move on to the next fragment
 			logger.Error().Err(err).Send()
+			// A source signals a fatal, scan-aborting error by yielding it with
+			// an empty fragment (e.g. the git source when `git log` writes to
+			// stderr). Propagate it so the caller reports a partial scan and
+			// exits non-zero, instead of silently reporting "no leaks found"
+			// with exit 0. A populated fragment indicates a recoverable
+			// per-item error (e.g. a single unreadable file), which we log and
+			// skip so the rest of the scan continues. See #2129.
+			if fragment.FilePath == "" && fragment.CommitSHA == "" && len(fragment.Raw) == 0 {
+				return err
+			}
 			return nil
 		}
 
--- a/detect/detect_test.go
+++ b/detect/detect_test.go
@@ -3,6 +3,7 @@ package detect
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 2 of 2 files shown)


acebot712:
Cross-referencing #2080 (by @Smeet23, open since April), which targets the same root cause via #1981 — I missed it before opening this; apologies for the overlap.

The two take different approaches, so please take whichever you prefer, or close this in favor of #2080:

- **#2080** fixes it in `sources/git.go` by returning the error directly from `Fragments()` (smaller diff). Its regression test spawns a real `git` process in a temp non-repo dir — the same pattern the existing `git_test.go` tests were commented out for (flakiness).
- **This PR (#2169)** fixes it one layer up in `DetectSource`, propagating any source-level error (yielded with an empty fragment) generically. Its test uses an in-memory fake source — deterministic, no `git` process — and the description includes an end-to-end repro against a git-crypt-style stderr-writing textconv filter.

Both are correct. Happy to defer to #2080 if you'd rather take the simpler change.

acebot712:
Closing in favor of #2080, which covers this. Apologies for the duplicate noise.

PR #2172, opened 2026-06-15
Decode base64 values containing valid UTF-8 characters

### LINKED ISSUE #2000: Decoding skips values with Unicode characters
**Describe the bug**
The current decoder logic filters values that contain non-ascii characters. This works quite well, however, if the encoded value contains a unicode character (chinese symbols, `ê`, etc.) it will be silently skipped.

https://github.com/gitleaks/gitleaks/blob/6eaad039603a4de39fddd1cf5f727391efe9974e/detect/codec/ascii.go#L13-L22

**To Reproduce**
1. Base64-encode a value that uses a valid UTF8 character
2. Attempt to decode it; it will be skipped.

**Expected behavior**
Segments with valid UTF8 characters ideally shouldn't be skipped. I don't know what the performance implications are of this.

cc @zricethezav @bplaxco

arpitjain099:
The base64 decoder dropped any decoded value that was not pure printable ASCII (`isPrintableASCII` in `detect/codec/ascii.go`). So when a base64 segment decoded to text containing a valid UTF-8 character (an accented letter like `é`, a CJK character, etc.), the whole segment was silently skipped and any secret next to it was never scanned. This is #2000.

The fix swaps that check for a UTF-8-aware one:
- ASCII bytes still go through the existing `printableASCII` table, so the current tab / newline / carriage-return allowances are unchanged.
- Multi-byte sequences are accepted when they are valid UTF-8 and the rune is printable (`unicode.IsPrint`), and rejected when the bytes are not valid UTF-8. So genuine binary still gets filtered out; only legitimately printable Unicode text now survives.

`isPrintableASCII` was the only consumer of that name (both calls in `decodeBase64`), so I renamed it to `isPrintable` to match the new behavior.

Added two `TestDecode` cases that decode to accented (`café-...`) and CJK (`...日本語...`) text; both failed before the change and pass now. `go test ./detect/codec/...` is green.

Fixes #2000

### DIFF
--- a/detect/codec/ascii.go
+++ b/detect/codec/ascii.go
@@ -1,5 +1,10 @@
 package codec
 
+import (
+	"unicode"
+	"unicode/utf8"
+)
+
 var printableASCII [256]bool
 
 func init() {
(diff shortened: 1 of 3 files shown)


sanmaxdev (approved):
Clean fix. The renamed `isPrintable` correctly preserves the existing `printableASCII` table for single-byte characters while accepting valid UTF-8 multi-byte runes via `utf8.DecodeRune` + `unicode.IsPrint`. Invalid UTF-8 (`RuneError` with size 1) is rejected, so no partial/garbled sequences slip through.

Verified:
- `go test ./detect/codec/...` (all pass, including new accented/CJK cases)
- `go vet ./detect/codec/...`
- `go build ./...`

arpitjain099:
@sanmaxdev this has carried your approval since 14 July and CI is green. Is anything still blocking it, or is it waiting on a release window? Happy to rebase if that would help.

PR #2174, opened 2026-06-17
doc: document --platform flag in helptext & readme

### LINKED ISSUE #2173: --platform enumeration insufficiently documented
CLI helptext and (or?) README/docs should accurately reflect the supported SCM platforms

**Additional context**
Helptext only lists github, gitlab

Code explicitly supports [more forges](https://github.com/gitleaks/gitleaks/blob/8ad8470035d31a209322c580153b45c18e21b980/cmd/scm/scm.go#L21)

Docs (and helptext, IMO) should make it easy to properly use this feature on private forges running a supported platform (e.g., forgejo on a custom domain) without examining the code directly.

cc @zricethezav

moutons:
Fixes: #2173

### Description:

- document supported `--platform` options
- prevent `--platform` helptext drift
-  add minimal docs for `--staged` and `--precommit` flags in the `git` command

### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,6 +179,8 @@ For example, if you wanted to run gitleaks on a range of commits you could use t
 command: `gitleaks git -v --log-opts="--all commitA..commitB" path_to_repo`. See the [git log](https://git-scm.com/docs/git-log) documentation for more information.
 If there is no target specified as a positional argument, then gitleaks will attempt to scan the current working directory as a git repo.
 
+Use `--platform` to generate links to findings on a specific SCM platform. Supported values: `github`, `gitlab`, `azuredevops`, `gitea`, `bitbucket`. Use `--staged` to scan staged (index) changes only. Use `--pre-commit` to scan unstaged working tree changes only (via `git diff`).
+
 #### Dir
 
 The `dir` (aliases include `files`, `directory`) command lets you scan directories and files. Example: `gitleaks dir -v path_to_directory_or_file`.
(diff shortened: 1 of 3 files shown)


moutons:
if y'all don't want the automated string build I can easily switch this back and write a test for it, but it seemed like overkill for such a simple change. building the string automatically prevents drift if/when platforms are added or removed, this seemed like the way to go.

ed: if you don't want the automated string build and a test isn't necessary, the first commit is just the proposed change without auto-helptext-build and no test

sanmaxdev (changes requested):
One gap I found while checking the help output: the same platform parser is still exposed on hidden `detect`, but its help still says only `(github, gitlab)` while `PlatformFromString` accepts `azuredevops`, `gitea`, and `bitbucket` there too. Can you update that flag text as well, or share if you want to leave hidden deprecated commands unchanged?

Verified:
- `go test ./cmd ./cmd/scm`
- `go run . git --help`
- `go run . detect --help`
- `git diff --check upstream/master...HEAD`

sanmaxdev (approved):
Thanks for the update. The `UserFacingPlatforms()` approach prevents the help text from drifting when new platforms are added, which is cleaner than hardcoding.

The hidden `detect` command still has the old `(github, gitlab)` text, but since it's deprecated and hidden, that's fine to leave as-is.

Verified:
- `go test ./cmd/...`
- `go build ./...`
- `go run . git --help` (shows all platforms dynamically)

PR #2175, opened 2026-06-19
fix: only apply withinColumns proximity within the same line

### LINKED ISSUE #2122: bug: withinColumns compares column numbers across different lines in composite rules, producing incorrect proximity results
## Summary

Hey! While exploring the composite rules feature introduced in v8.28.0, I noticed 
what looks like a bug in how `withinColumns` proximity is calculated in 
`detect/detect.go`.

The `withinProximity()` function checks column proximity by doing a raw subtraction 
of `StartColumn` values between the primary and required findings. The problem is 
that **column numbers reset at every new line** — so comparing columns across 
findings that live on different lines produces a meaningless result.

## Location

`detect/detect.go` — `withinProximity()` lines 675–700

## The Code

```go
if requiredRule.WithinColumns != nil {
    // BUG: column numbers restart at each newline.
    // primary on line 1 col 5, required on line 300 col 6:
    // colDiff = |5 - 50| = 45 — meaningless cross-line comparison
    colDiff := abs(primary.StartColumn - required.StartColumn)
    if colDiff > *requiredRule.WithinColumns {
        return false
    }
}
```

## A Concrete Example

Say you have a composite rule with `withinColumns: 5`:

- Primary finding → line 1, column 5
- Required finding → line 300, column 6

`colDiff = |5 - 6| = 1` → satisfies `withinColumns: 5` ✓

But these two findings are **300 lines apart**. That's not what column proximity 
is supposed to mean. The README diagram clearly shows `withinColumns` as a 
horizontal (same-line) constraint.

The reverse problem also exists — two findings on the **same line** 60 characters 
apart would fail `withinColumns: 50` even though they're exactly the case this 
feature is designed for.

## Also Found

There's a commented-out debug print left in this same function on line 676:

```go
// fmt.Println(requiredRule.WithinLines)
```

Happy to clean that up in the same PR.

## The Fix

When findings are on different lines, the column constraint should not be 
considered satisfiable — column proximity is a same-line concept:

```go
if requiredRule.WithinColumns != nil {
    // Column proximity is only meaningful on the same line.
    if primary.StartLine != required.StartLine {
        return false
    }
    colDiff := abs(primary.StartColumn - required.StartColumn)
    if colDiff > *requiredRule.WithinColumns {
        return false
    }
}
```

## What I'll Include in the PR

- The fix in `detect/detect.go`
- Removal of the leftover debug print on line 676
- Unit tests covering:
  - Same line, within columns → should match 
  - Same line, outside columns → should not match 
  - Different lines, close columns → should not match 
  - Different lines, withinLines also set → correct rectangular behavior 

I know composite rules are marked experimental so I'll make sure to note the 
behavior change clearly in the PR description. Happy to open it right away if 
this looks correct to you, and let me know if you see the intended behavior 
differently!

cc @zricethezav

ardittirana:
## What was broken

`withinProximity()` computed `withinColumns` by subtracting the `StartColumn` of the two findings without checking they were on the same line. Column numbers restart at every new line, so the comparison is only meaningful within a single line.

Closes #2122.

## Why it matters

For composite rules this produced incorrect proximity results: a primary finding on line 1 column 5 and a required finding on line 300 column 6 gave `colDiff = 1`, satisfying `withinColumns: 5` even though the findings are 300 lines apart. The README presents `withinColumns` as a horizontal, same-line constraint.

## What changed

`withinColumns` now returns false when the two findings are on different lines, then compares columns only for same-line findings. A leftover commented-out debug print in the same function was removed.

## Tests

Added `TestWithinProximityColumns` covering same-line within range, same-line out of range, and the cross-line case. The cross-line test fails on current code (returns true) and passes with this change. `go build ./...`, `go vet ./detect/`, and the detect suite pass.

### DIFF
--- a/detect/detect.go
+++ b/detect/detect.go
@@ -673,7 +673,6 @@ func (d *Detector) hasAllRequiredRules(auxiliaryFindings []*report.RequiredFindi
 }
 
 func (d *Detector) withinProximity(primary, required report.Finding, requiredRule *config.Required) bool {
-	// fmt.Println(requiredRule.WithinLines)
 	// If neither within_lines nor within_columns is set, findings just need to be in the same fragment
 	if requiredRule.WithinLines == nil && requiredRule.WithinColumns == nil {
 		return true
@@ -687,9 +686,13 @@ func (d *Detector) withinProximity(primary, required report.Finding, requiredRul
 		}
 	}
 
-	// Check column proximity (horizontal distance)
+	// Check column proximity (horizontal distance). Column numbers restart at
+	// each new line, so a column comparison is only meaningful when both
+	// findings are on the same line.
 	if requiredRule.WithinColumns != nil {
-		// Use the start column of each finding for proximity calculation
+		if primary.StartLine != required.StartLine {
+			return false
+		}
 		colDiff := abs(primary.StartColumn - required.StartColumn)
 		if colDiff > *requiredRule.WithinColumns {
 			return false
--- a/detect/within_proximity_test.go
+++ b/detect/within_proximity_test.go
@@ -1,0 +1,52 @@
+package detect
+
+import (
+	"testing"
+
+	"github.com/zricethezav/gitleaks/v8/config"
+	"github.com/zricethezav/gitleaks/v8/report"
+)
+
+func TestWithinProximityColumns(t *testing.T) {
+	col := func(n int) *int { return &n }
+	d := &Detector{}
+
+	tests := []struct {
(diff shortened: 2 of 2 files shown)


sanmaxdev (commented):
I checked the boundary here. The guard preserves same-line column proximity and rejects cross-line pairs before comparing columns, which matches how `StartColumn` resets per line.

Passed:
- `go test ./detect`
- `git diff --check upstream/master...HEAD`

Issue #2158, opened 2026-06-04
Add detection for Anthropic OAuth tokens (sk-ant-oat01-, sk-ant-ort01-)


mazen-salah:
### Description

gitleaks already detects Anthropic API keys (`sk-ant-api03-`) and admin keys (`sk-ant-admin01-`), but not the OAuth tokens issued by `claude setup-token`:

- `sk-ant-oat01-` — OAuth access token (tied to a Claude.ai Pro/Max subscription)
- `sk-ant-ort01-` — OAuth refresh token (can mint new access tokens)

These are long-lived credentials stored in `~/.claude/.credentials.json` and exported as `CLAUDE_CODE_OAUTH_TOKEN`, so they can be committed by accident the same way API keys are. The refresh token is especially sensitive.

I'd like to add two rules for these, following the existing `anthropic` rule style. Happy to open a PR.

Issue #2104, opened 2026-05-09
Request: detection rules for Claude Code OAuth tokens and MCP server credentials


ipunithgowda:
## Summary
Two AI-tooling secret patterns that have become widespread in 2025-2026 and don't currently have gitleaks coverage:

1. **Claude Code OAuth tokens** issued by `claude setup-token`. Format `sk-ant-oat01-<95 chars from [A-Za-z0-9_-]>`, total length 108. Distinct from the existing `anthropic-api-key` (`sk-ant-api03-`) and `anthropic-admin-api-key` (`sk-ant-admin01-`) rules. Stored at `~/.claude/.credentials.json` and commonly exported as `CLAUDE_CODE_OAUTH_TOKEN` in CI / GitHub Actions.

2. **MCP server credentials** embedded in Model Context Protocol config files (`claude_desktop_config.json`, `.mcp.json`, `mcp.json`). These `env` blocks routinely hold long-lived Linear / GitHub / Slack / custom MCP server tokens in plain text and frequently get committed by mistake.

## Why
Both patterns have public references and zero current coverage. Verified via `grep -ri claude-code cmd/generate/config/rules/` (zero hits) and `grep -ri mcp cmd/generate/config/rules/` (zero hits).

## Proposal
Two rules:
- `claude-code-oauth-token` modeled on `anthropic-api-key` structure (regex + keywords + helper-validated TPs/FPs).
- `mcp-server-credential` path-scoped to MCP config filenames, with placeholder allowlist and entropy floor.

Happy to send a PR for both in one branch. Will reference this issue in the PR body.

PR #2177, opened 2026-06-22
feat(rules): add Pinecone and LangSmith API key detection rules

DevamShah:
## Summary
Adds two new high-confidence Gitleaks rules — `pinecone-api-key` and `langsmith-api-key` — covering credentials for the vector-database and LLM-observability layers that now sit at the centre of most RAG/AI stacks. Both rules anchor on vendor-issued, structurally distinctive prefixes, so they detect real leaks while keeping false positives near zero.

## Problem / motivation
Gitleaks ships rules for the AI *model* providers (Anthropic, OpenAI, Cohere, Hugging Face, Perplexity, Bedrock) but has no coverage for the *retrieval and observability* tier those models depend on. A leaked Pinecone or LangSmith key is not a low-value finding:

- **Pinecone** keys grant read/write access to the vector indexes that store an application's embedded private data. A compromised key enables silent exfiltration of the underlying corpus and index poisoning of downstream retrieval.
- **LangSmith** keys (personal access tokens and service keys) expose tracing data — full prompts, completions, tool I/O — plus the ability to act against connected LangChain/LangGraph workloads.

Both are routinely committed to `.env` files, notebooks, CI config, and IaC, and today pass through Gitleaks undetected (at best they trip the noisy `generic-api-key` rule, which yields a generic finding with no provenance).

## Change
Following `CONTRIBUTING.md` exactly:

- `cmd/generate/config/rules/pinecone.go` — `PineconeApiKey()`
- `cmd/generate/config/rules/langsmith.go` — `LangsmithApiKey()`
- Registered both in `cmd/generate/config/main.go` (alphabetical ordering preserved).
- Regenerated `config/gitleaks.toml` via `go generate ./cmd/generate/config/`.

Both rules use `GenerateUniqueTokenRegex`, consistent with the existing prefix-anchored AI-provider rules (`anthropic.go`, `openai.go`), because the token prefixes are themselves the high-signal identifier:

| Rule | Pattern | Source of truth |
|------|---------|-----------------|
| `pinecone-api-key` | `pcsk_[A-Za-z0-9]{5,6}_[A-Za-z0-9]{63}` | Verified against Pinecone's `pcsk_<label>_<key>` format |
| `langsmith-api-key` | `lsv2_(?:pt\|sk)_[a-z0-9]{32}_[a-z0-9]{10}` | LangSmith PAT (`lsv2_pt_`) and service-key (`lsv2_sk_`) prefixes per LangSmith admin docs |

[...]

PR #2178, opened 2026-06-22
feat(rules): add Anthropic OAuth refresh token rule (sk-ant-ort01-)

DevamShah:
### Summary

Adds a default rule `anthropic-oauth-refresh-token` detecting the OAuth **refresh** tokens (`sk-ant-ort01-`) minted by Claude Code and other Anthropic OAuth flows. This complements the existing `sk-ant-api03-` / `sk-ant-admin01-` API-key rules and the OAuth **access**-token work tracked in #2104 / #2105.

### Why this is distinct from #2105

Issue #2104 and the open PR #2105 cover the OAuth **access** token (`sk-ant-oat01-`). Neither covers the **refresh** token (`sk-ant-ort01-`), which is stored alongside the access token (e.g. in the macOS Keychain `claudeAiOauth` entry: `{accessToken: sk-ant-oat01-…, refreshToken: sk-ant-ort01-…}`). A leaked refresh token is arguably higher-impact than the access token because it can mint new access tokens until revoked. This PR adds only that missing refresh-token rule, so it stacks cleanly on top of #2105 without duplicating it.

### Change

- New `AnthropicOauthRefreshToken()` rule in `cmd/generate/config/rules/anthropic.go`:
  - `RuleID: anthropic-oauth-refresh-token`
  - Regex: `sk-ant-ort01-[a-zA-Z0-9_\-]{95}` (same char class and body length as the confirmed `sk-ant-oat01-` access token, which is documented at 95 chars / 108 total). The refresh token shares the `sk-ant-` OAuth issuer family; if maintainers have a confirmed refresh-token length that differs, happy to adjust.
  - Keyword prefilter `sk-ant-ort01`, with too-short and wrong-prefix false-positive samples.
- Registered in `cmd/generate/config/main.go` and `config/gitleaks.toml` regenerated via `go generate ./...`.

### Testing

- `go generate ./...` runs `utils.Validate(tps, fps)` — passes (TPs match, FPs rejected).
- `go build ./...` clean; `go generate` is idempotent (CI `git diff --exit-code` passes).
- Live: built the binary and scanned a file containing `sk-ant-ort01-<95>` → detected as `anthropic-oauth-refresh-token`.

### DIFF
--- a/cmd/generate/config/main.go
+++ b/cmd/generate/config/main.go
@@ -40,6 +40,7 @@ func main() {
 		rules.AmazonBedrockAPIKeyShortLived(),
 		rules.AnthropicAdminApiKey(),
 		rules.AnthropicApiKey(),
+		rules.AnthropicOauthRefreshToken(),
 		rules.ArtifactoryApiKey(),
 		rules.ArtifactoryReferenceToken(),
 		rules.AsanaClientID(),
--- a/cmd/generate/config/rules/anthropic.go
+++ b/cmd/generate/config/rules/anthropic.go
@@ -67,3 +67,32 @@ func AnthropicAdminApiKey() *config.Rule {
 
 	return utils.Validate(r, tps, fps)
 }
+
+func AnthropicOauthRefreshToken() *config.Rule {
+	// define rule
+	r := config.Rule{
+		RuleID:      "anthropic-oauth-refresh-token",
+		Description: "Identified an Anthropic OAuth refresh token (e.g. Claude Code), which can be exchanged for new access tokens and provides long-lived access to Anthropic APIs on behalf of a user.",
+		Regex:       utils.GenerateUniqueTokenRegex(`sk-ant-ort01-[a-zA-Z0-9_\-]{95}`, false),
+		Keywords: []string{
+			"sk-ant-ort01",
+		},
+	}
+
+	// validate
+	tps := []string{
+		// Valid OAuth refresh token example
+		"sk-ant-ort01-ort01samplexyzABC0123456789-_ort01samplexyzABC0123456789-_ort01samplexyzABC0123456789-_ort01sam",
+		// Generate additional random test tokens
+		utils.GenerateSampleSecret("anthropic", "sk-ant-ort01-"+secrets.NewSecret(utils.AlphaNumericExtendedShort("95"))),
+	}
+
+	fps := []string{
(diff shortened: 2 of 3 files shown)


DevamShah:
I went back through this one properly before asking anyone to look at it, and I think the right outcome is to **stand it down in favour of #2159**.

**It's a duplicate, and mine is the later and narrower one.** #2159 (@mazen-salah) was opened 2026-06-04, eighteen days before this PR, and closes #2158. It adds the same rule ID, `anthropic-oauth-refresh-token`, *and* `anthropic-oauth-token` for the `sk-ant-oat01-` access token. This PR only covers `ort01`, so it's a strict subset. I opened it without spotting #2158/#2159 first — my mistake.

**More importantly, my regex is the weaker of the two, and I measured it.** This PR anchors on a fixed body length:

```
sk-ant-ort01-[a-zA-Z0-9_\-]{95}
```

#2159 uses an open-ended minimum instead, `{40,}`, on the grounds that Anthropic doesn't publish a fixed length for these tokens. I built both branches and scanned one file containing synthetic `sk-ant-ort01-` tokens with body lengths of 40, 60, 94, 95, 96 and 120, running each binary with `--enable-rule anthropic-oauth-refresh-token`:

| branch | findings |
|---|---|
| this PR, `{95}` | **1 of 6** — only the body of exactly 95 |
| #2159, `{40,}` | **6 of 6** |

That's the whole argument. A fixed `{95}` fails closed and silently: if the real token body is ever 94 or 96 characters, or Anthropic changes it, the rule matches nothing and nobody finds out, because a missed detection produces no signal. I could not verify the true body length against a documented source, which is exactly the situation where a conservative lower bound is the correct call and a hardcoded length is not. #2159 got that judgement right and I got it wrong.

So: **#2159 is the one to land, not this.** Happy to close this myself — say the word and I'll do it, or just close it directly, whichever you prefer. I'd rather not leave a duplicate rule ID sitting in the queue for a maintainer to have to untangle.

For what it's worth, while checking the above I also confirmed #2159 builds and its config regenerates clean on current `master` (`b58d3f1`), so it isn't stale either. If it would help move it along I'm glad to review it in detail.

Separately, my other rules PR #2177 is unrelated to this one (Pinecone + LangSmith), has no duplicate in the queue, and is not affected by whatever happens here.

PR #2193, opened 2026-07-16
fix(config): detect new GitHub App installation tokens (fixes #2192)

### LINKED ISSUE #2192: GitHub App format changing
I do not use this product, so this is just a heads-up that https://github.com/gitleaks/gitleaks/blob/4c232b5014f7618360bd992b4c489cb055881c6b/config/gitleaks.toml#L2133 might break: https://github.blog/changelog/2026-04-24-notice-about-upcoming-new-format-for-github-app-installation-tokens/

tianrking:
### Description:

Fixes #2192. GitHub App installation tokens now use `ghs_<APPID>_<JWT>` and are variable length, so the former fixed-length rule misses them. This preserves legacy `ghu_` and 36-character `ghs_` token coverage, while adding a structurally constrained matcher for the new token format: a numeric app ID plus a three-segment base64url JWT beginning with `eyJ`.

The generator includes positive samples for the new form and rejects malformed/non-numeric-app-ID examples. `config/gitleaks.toml` was regenerated from the rule source.

### Checklist:

* [x] Does your PR pass tests? `go generate ./cmd/generate/config`, `go test ./...`, and `go run . version`
* [x] Have you written new tests for your changes? Added positive and negative rule-generator validation samples.
* [x] Have you lint your code locally prior to submission? `gofmt` applied and `git diff --check` passed. `go vet ./...` currently reports two pre-existing context-cancel paths in `detect/detect_test.go`; `go test -race ./...` cannot run here because the Windows environment has no CGO C compiler.

### DIFF
--- a/cmd/generate/config/rules/github.go
+++ b/cmd/generate/config/rules/github.go
@@ -76,7 +76,7 @@ func GitHubApp() *config.Rule {
 	r := config.Rule{
 		RuleID:      "github-app-token",
 		Description: "Identified a GitHub App Token, which may compromise GitHub application integrations and source code security.",
-		Regex:       regexp.MustCompile(`(?:ghu|ghs)_[0-9a-zA-Z]{36}`),
+		Regex:       regexp.MustCompile(`(?:ghu_[0-9a-zA-Z]{36}|ghs_(?:[0-9a-zA-Z]{36}|[0-9]+_eyJ[a-zA-Z0-9_-]+\.[a-zA-Z0-9_-]+\.[a-zA-Z0-9_-]+))`),
 		Entropy:     3,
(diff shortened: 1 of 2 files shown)


sanmaxdev (commented):
The pattern matches GitHub's documented `ghs_APPID_JWT` shape while preserving the legacy 36-character `ghs_` and `ghu_` forms. The numeric app ID and three base64url segments are covered by the validation samples, and the checked-in config remains stable after regeneration.

Verified:
- `go test ./cmd/generate/config/... ./config -count=1`
- `go generate ./... && git diff --exit-code`
- `go build ./...`
- `go test ./... -race -count=1`
- `git diff --check`

PR #2215, opened 2026-07-28
fix: calculate columns from line starts

### LINKED ISSUE #1424: Off-by-one error in startColumn and endColumn calculation
**Describe the bug**
`startColumn` and `endColumn` are calculated differently on first and all subsequent lines of the file.

**To Reproduce**
Run
```gitleaks detect -f sarif -r output.json -s some_dir --config example.toml --no-git```
`example.toml` and `some_dir/bugged.txt` are attached

**Expected behavior**
Column numbers are the same in both findings.

**Observed behavior**
Column numbers in the first finding are different from numbers in second finding

 - OS: macOs
 - Gitleaks Version: master (`26f34692fac6e9daec13c770421b4ed990d1c321`)

**Additional context**
It seems that the problem is located in [detect/location.go](https://github.com/gitleaks/gitleaks/blob/26f34692fac6e9daec13c770421b4ed990d1c321/detect/location.go#L44). On the first line `prevNewLine` (=0) points to the first character in the fragment, but on subsequent lines it points to character before line start, so `\n` from previous line is included in column number calculation.

example.toml: 
```
[[rules]]
    id = "f"
    description = "f"
    regex = '''f'''
    tags = ["f"]
```
bugged.txt: 
```
..f..
..f..
```
[output.json](https://github.com/user-attachments/files/15864672/output.json)

cc @zricethezav

### PR DESCRIPTION
**Pull request by Eljees**
### Description:

Fixes #1424.

Newline matches contain both the newline byte offset and the first byte of the next line. The location calculation previously used the newline byte itself as the column offset, so matches after the first line were shifted one column to the right. This change tracks the actual line start separately while preserving the existing line-slice indexes.

The regression coverage now derives newline indexes from realistic LF and CRLF input, verifies identical matches on the first and later lines, and updates the affected expected coordinates for plain and decoded findings.

Tested with Go 1.24.11:

```text
go test ./detect -run '^Test(GetLocation|Detect|FromGitStaged|FromFiles|DetectWithArchives|DetectRuleAllowlist)$' -count=1
```

### Checklist:


Eljees:
Ping on this one and #2219 - both open since 28 July with no review. #2215 is a column offset fix, #2219 narrows the hashicorp-tf-password rule; they are unrelated and can be taken separately.

Eljees:
Following up on my ping of 9 August, with a correction to it: CI has not run on either PR. The `Test` workflow on both is at `action_required` and needs a maintainer to approve it, so there are no results to judge them by yet.

#2215 calculates columns from line starts rather than from the match offset; #2219 requires the identifier to end with the keyword in `hashicorp-tf-password`. Unrelated to each other.

PR #2220, opened 2026-07-28
ci: attach provenance and SBOM attestations to the published images

kobihikri:
Hi, and thanks for gitleaks.

`.github/workflows/release.yml` publishes the image to both `zricethezav/gitleaks` on Docker Hub and `ghcr.io/gitleaks/gitleaks` on every release, but the pushed manifests carry no provenance or SBOM attestation.

I want to make the case for why this one is worth doing rather than just noting the gap. gitleaks is a scanner people run *inside their own CI*, over their own source, specifically to find secrets — so the container tends to be given a repository checkout and run in a job that has credentials around. That is about as much trust as a third-party image gets. Being able to verify that `zricethezav/gitleaks:latest` was built by this workflow from this repository is worth more here than it would be for most images.

The change is two lines on the build step:

```yaml
          push: true
          provenance: mode=max
          sbom: true
```

I used BuildKit's own attestations rather than a separate signing step because your single build pushes to two registries at once — BuildKit attaches the attestation to the image manifest, so Docker Hub and GHCR both get it with no extra plumbing and no second credential. It also means **no permissions change**: `contents: read` and `packages: write` stay exactly as they are, and nothing needs `id-token`.

Consumers can then verify with:

```
docker buildx imagetools inspect ghcr.io/gitleaks/gitleaks:<tag> --format '{{ json .Provenance }}'
```

Two caveats worth stating: `mode=max` records the full build including build args (`provenance: true` gives a smaller record if any have ever been sensitive), and attestations add an extra manifest to the index — Docker Hub and GHCR both support this, but it is worth knowing if a mirror sits in front.

No SLSA level claimed; the attestation is what BuildKit produces, and characterising the overall build is your call.

Disclosure: I used AI assistance to help spot this and prepare the change, and I read the workflow and both push targets myself.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -53,5 +53,7 @@ jobs:
           platforms: linux/amd64,linux/arm64
           context: .
           push: true
+          provenance: mode=max
+          sbom: true
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}


kobihikri:
**Correction — I got a fact wrong in this PR, and I would rather flag it myself than let it sit.**

I wrote that the pushed manifest "carries no provenance or SBOM attestation". That is half wrong, and the wrong half matters.

**Provenance is already there.** For *public* repositories, `docker/build-push-action` adds provenance attestations with `mode=max` by default — [Docker's documentation](https://docs.docker.com/build/ci/github-actions/attestations/) states it plainly: *"Public repos: provenance attestations with `mode=max` are automatically added"*. I checked published images and they do already carry attestation manifests. So the `provenance: mode=max` line in my diff makes existing behaviour **explicit**; it does not add anything new.

**The SBOM is genuinely new.** That part stands — the same page says *"SBOM attestations aren't automatically added to the image"*, and `sbom: true` is what enables them.

I also wrote in the caveats that `provenance: true` gives "a smaller record". That is wrong as well: `true` resolves to max on a public repo, and the smaller setting is `provenance: mode=min`.

**So the honest description of this PR is:** it adds an SBOM attestation, and pins the provenance mode explicitly instead of relying on the default. Both are still defensible — an explicit line means the behaviour will not change quietly if the default ever does — but it is a smaller change than my description implied, and you should judge it on that basis rather than on what I originally wrote.

Happy to retitle and rewrite the description accordingly, or to close this if the SBOM alone is not worth the diff to you. Either is fine — just say which and I will act on it.

Apologies for the inaccuracy. It was caught by a maintainer reviewing the same change on another project, and they were right to.

PR #2236, opened 2026-08-10
docs: clarify that git log -p scans only added lines, not pre-existing content

### LINKED ISSUE #2218: Pre-receive hook docs: --log-opts misses pre-existing secrets in changed files
## Problem

The official pre-receive hook guidance recommends:

gitleaks git --log-opts="--all commitA..commitB"

This only scans **added lines** in the git diff (`git log -p`). If a file already contains a secret and a developer edits a different line in that file, the push passes undetected.

This is a documentation gap — users deploying pre-receive hooks don't know about this limitation until they discover it in production.

## Steps to reproduce

1. Create a repo with a pre-existing secret in file `config.yml`
2. Push a commit that edits a different line in `config.yml` (not the secret)
3. Run `gitleaks git --log-opts="--all oldrev..newrev"` on the push range
4. No finding is reported — the zombie secret is missed

## Expected behavior

The documentation should clearly state that `--log-opts` only scans additions in git history and does **not** scan pre-existing content in changed files. Users relying on pre-receive hooks for enforcement need to be aware of this.

## Environment

- gitleaks v8.18.4
- GitHub Enterprise Server pre-receive hook context

RajeshRajendiran:
Fixes #2218

`git` mode's diff scanning (`sources/git.go`) pulls each hunk's content via `textFragment.Raw(gitdiff.OpAdd)`, which only includes lines with `Op == OpAdd` - context and unchanged lines are excluded. So a `git log -p` scan, range-restricted or not, only ever sees lines added within the scanned range, never a file's pre-existing content.

This is a real gap for anyone using the README's documented `--log-opts="--all commitA..commitB"` pattern for enforcement (e.g. a pre-receive hook scanning `oldrev..newrev`): if a secret was already committed before the range being scanned, and a later commit in that range only edits a different line in the same file, the secret is silently not found - the scan reports clean.

Added a note to the Git section of the README clarifying this behavior and recommending a periodic full-history scan alongside range-restricted ones.

## Test plan
- [x] Doc-only change, no code affected.


sanmaxdev (commented):
Confirmed the note matches the git scan behavior and the linked issue. `git diff --check upstream/master...HEAD` passes.

PR #2241, opened 2026-08-17
Fix panic on invalid regex in user config

### LINKED ISSUE #1846: bad gitleaks config causes regex panic
**Describe the bug**
bad allow list regex causes a panic

**To Reproduce**

use the following bad config

```toml
[extend]
    useDefault = true
     
[allowList]
    paths = [
        '''.*/test/.*''',
        '''*.test.js''',
        '''*.stories.js'''
    ]
```
[it will produce a panic when used
```bash
gitleaks detect -c .gitleaks.toml -v

    ○
    │╲
    │ ○
    ○ ░
    ░    gitleaks

panic: regexp: Compile(`*.test.js`): error parsing regexp: missing argument to repetition operator: `*`

goroutine 1 [running]:
regexp.MustCompile({0x1400009e150, 0x9})
        regexp/regexp.go:313 +0xb0
github.com/zricethezav/gitleaks/v8/regexp.MustCompile(...)
        github.com/zricethezav/gitleaks/v8/regexp/stdlib_regex.go:14
github.com/zricethezav/gitleaks/v8/config.(*ViperConfig).Translate(_)
        github.com/zricethezav/gitleaks/v8/config/config.go:189 +0x57c
github.com/zricethezav/gitleaks/v8/cmd.Config(_)
        github.com/zricethezav/gitleaks/v8/cmd/root.go:187 +0xc4
github.com/zricethezav/gitleaks/v8/cmd.runDetect(0x101001040, {0x100b18321?, 0x4?, 0x100b1824d?})
        github.com/zricethezav/gitleaks/v8/cmd/detect.go:58 +0x5c
github.com/spf13/cobra.(*Command).execute(0x101001040, {0x140000b0270, 0x3, 0x3})
        github.com/spf13/cobra@v1.9.1/command.go:1019 +0x810
github.com/spf13/cobra.(*Command).ExecuteC(0x101000d80)
        github.com/spf13/cobra@v1.9.1/command.go:1148 +0x350
github.com/spf13/cobra.(*Command).Execute(...)
        github.com/spf13/cobra@v1.9.1/command.go:1071
github.com/zricethezav/gitleaks/v8/cmd.Execute()
        github.com/zricethezav/gitleaks/v8/cmd/root.go:173 +0x24
main.main()
        github.com/zricethezav/gitleaks/v8/main.go:18 +0x9c

```
**Expected behavior**
i expect the scan to gracefully error out and not execute.

 - OS:
 - Gitleaks Version: 8.24.3 && 8.25.1

**Additional context**
we are scanning thousands of repositories and would like to continue offering people the ability to define their own config and a panic will crash the application ( we use the source code directly ).

cc @zricethezav

harriiinnii:
### Problem

When a `.gitleaks.toml` config file contains an invalid regex (e.g. `*.test.js` — a bare `*` with no operand), `config/config.go` calls `regexp.MustCompile`, which panics and crashes the entire process:

```
panic: regexp: Compile(`*.test.js`): error parsing regexp: missing argument to repetition operator: `*`
```

For services that load user-supplied configs at scale this makes the scan unrecoverable for the whole process instead of reporting a clear per-config error.

### Fix

- Add a `Compile(str string) (*Regexp, error)` wrapper to both regexp backends (`regexp/stdlib_regex.go` and `regexp/wasilibs_regex.go`)
- Replace every `regexp.MustCompile` call that processes **user-supplied patterns** in `config/config.go` (`Translate` for rule regex/path, `parseAllowlist` for allowlist regexes/paths) with `regexp.Compile`, returning a descriptive `error` instead of panicking

`MustCompile` is still used for the built-in `gitleaks.toml` rules, where a bad pattern is a developer bug and a panic is appropriate. The change only affects the code paths that process user-provided config values.

### Before / After

```
# Before
panic: regexp: Compile(`*.test.js`): error parsing regexp: ...

# After
Error: *.test.js: invalid allowlist path regex "*.test.js": error parsing regexp: missing argument to repetition operator: `*`
exit status 1
```

Fixes #1846


CAOShurong (approved):
Verified at exact head `e7948f4` against base `b58d3f1` (current `master`). Build, `go vet ./config/... ./regexp/...`, and the full suite (`go test ./...`) are green; `git diff --check` is clean.

**Independent reproduction (no run no claim):**
- Rule-regex path: a config with `regex = "([a-z"` panics on `master` (`panic: regexp: Compile(...): missing closing ]`) with exit 2. At the PR head it returns a clean error and exits 1: `FTL Failed to load config error="bad-regex-rule: invalid rule regex: ..."`.
- Allowlist-path path: `allowlist.paths = ["*.test.js"]` panics on `master`; at the head it returns `invalid allowlist path regex "*.test.js": missing argument to repetition operator`.
Both crashes are confirmed fixed. The fix correctly scopes `Compile` only to **user-supplied** patterns (`rule.regex`, `rule.path`, `allowlist.regexes`, `allowlist.paths`) and leaves `MustCompile` on the built-in `gitleaks.toml` rules, where a bad pattern is a developer bug. Adding `Compile` to both the stdlib and wasilibs backends keeps the WASM build consistent. The remaining `MustCompile` calls in `config/utils.go` compile internal constants, so they are correctly out of scope.

**Context:** issue #1846 is still the target and is not yet fixed on `master` (the bug reproduces there today). #2082 and #2228 are open duplicates of this same fix; this PR's diff is the smallest and cleanest of the three.

**Minor suggestion (not blocking):** the PR currently ships no regression test. A small table-driven test in `config/config_test.go` feeding an invalid `rule.regex` / `allowlist.paths` and asserting the returned error (no panic) would lock the behavior in and make review/merge cheaper.

**Verification:** built and ran the binary at both `master` and the PR head against the two repro configs above — confirmed panic→clean-error transition on both user-supplied paths.

_Disclosure: this review was prepared with AI assistance (independent build + reproduction); the verification steps above are reproducible from the PR diff._

CAOShurong (changes requested):
Re-reviewed exact new head `0c2402a9f` after the regression-test follow-up. The four added invalid-regex fixtures are the right coverage, but this commit currently breaks the focused config suite:

```text
go test ./config/... -count=1
FAIL
```

Two issues are mixed into `config/config_test.go`:

1. Existing expected regex literals were changed outside the new cases (generic/Discord/AWS fixtures), and the `rule_missing_id` expected text gained an extra space in `[ a-h0-9]`. Those changes make the existing `TestTranslate` / `TestTranslateExtend` comparisons fail and should be reverted to the `e7948f4` versions.
2. The two new global-allowlist expectations omit the existing `[[allowlists]] ` error prefix. Actual errors are `[[allowlists]] invalid allowlist regex ...` and `[[allowlists]] invalid allowlist path regex ...`.

The same command passes on the previously reviewed head `e7948f4`. I also verified locally that restoring the pre-existing fixture literals, removing the stray space, and adding the two allowlist prefixes makes `go test ./config/... -count=1` pass with all four new regression cases retained. `git diff --check` is otherwise clean.

Please keep the new fixtures, revert the unrelated expected-regex edits, and align the two allowlist messages with the actual prefixed errors.

Disclosure: I used AI assistance for diff inspection and local test execution; the failing command and exact-head comparison above were run and checked locally.

PR #2249, opened 2026-08-26
fix(sources): report files that could not be read instead of exiting 0

### LINKED ISSUE #2232: Unreadable files are silently skipped and the run still reports "no leaks found" with exit 0
Hi — I do reliability work on the "did this actually happen" class of bug, and I found one in
gitleaks while surveying for it across a few ecosystems. Filing it because the impact is quiet and
the fix is small. No pitch attached, and the issue stands whatever you decide about it.

**The short version:** if gitleaks cannot open a file, it skips it, and the run still prints
`no leaks found` and exits `0`. The "partial scan" machinery you already built never fires.

### The path

`sources/files.go:155-162`

```go
f, err := os.Open(scanTarget.Path)
if err != nil {
    if os.IsPermission(err) {
        logger.Warn().Msg("skipping file: permission denied")
    }
    wg.Done()
    return nil
}
```

That `return nil` is the only channel this callback has back to the caller — the signature at
`detect/detect.go:209` is `func(fragment sources.Fragment, err error) error`. So an open failure
cannot reach the `err` that `DetectSource` returns, which means it cannot reach
`findingSummaryAndExit`, which means `cmd/root.go:446` takes the `err == nil` branch and prints the
plain `no leaks found` rather than `no leaks found in partial scan`, and `:493`'s `os.Exit(1)` never
runs.

**The thing that makes this worth filing rather than shrugging at:** you already solved it. The
partial-vs-clean distinction at `cmd/root.go:446-461` is real and correct, in both the message and
the exit code. Somebody clearly thought about this. It is just unreachable from what I would guess
is the most common failure mode a file scanner has.

### The narrower half

`:157` only logs when `os.IsPermission(err)`. Every other `os.Open` failure is completely silent —
no warning, no counter. I grepped `detect.go` for anything tallying this and found only
`TotalBytes` (`:106`, `:278`, `:280`), so there is no count of skipped files anywhere to surface.

So there are three outcomes wearing two names:

| what happened | what the operator sees |
|---|---|
| scanned, clean | `no leaks found`, exit 0 |
| skipped, permission denied | `no leaks found`, exit 0 (+ one WARN line) |
| skipped, any other open error | `no leaks found`, exit 0 |

For a tool whose output gates CI, the third row is the one I would want back.

### What I have not verified, plainly

I have not run a fixture. Everything above is read from source at the cited lines, but I have not
confirmed that a real unreadable file reaches that branch in practice rather than failing earlier in
`scanTargets`. If it does not, this is wrong and I would like to know.

I am being explicit about that because I had a very similar-looking finding in a coverage library
the same day, wrote the headline, then built the fixture and it did not reproduce — the function did
what I said and the claim I built on top of it was about a pipeline I had not run. I published the
retraction. So: this is a source-level reading, offered as one.

### One reason I think the ask is reasonable rather than a preference

I checked whether this is just how scanners are, because if it were, this issue would be noise.
trufflehog ships `--fail-on-scan-errors` (`main.go:76`), and it is wired to real structure rather
than to logs: their filesystem source returns `reporter.ChunkErr(...)` on a stat failure
(`pkg/sources/filesystem/filesystem.go:437`), those land as per-job errors on a snapshot, and
`main.go:1233-1247` walks them. So a scan that could not read files can be made to exit non-zero,
by design.

I mention it only as evidence that the concept is considered worth having by people solving the
same problem — and you are closer to it than they are, because the partial-scan branch at
`root.go:446-461` already exists here. This is a wiring gap, not a missing feature.

### A possible shape for the fix

Count the skips and let them reach the summary — something like an atomic counter on the detector
incremented at `files.go:161`, then in `findingSummaryAndExit` treat `skipped > 0` the same way
`err != nil` is already treated: say "partial", and exit non-zero. That reuses the branch that
exists rather than adding a concept.

Also worth considering separately: `cmd/root.go:483` writes `findings` only, so even in the correct
partial branch the JSON report carries no partial or skipped field. Anything parsing that file
downstream loses the distinction the console preserved.

Happy to open a PR if the direction is welcome.

vaibhav8a:
Closes #2232.

## What was happening

`sources/files.go` returned `nil` when it could not open a file:

```go
f, err := os.Open(scanTarget.Path)
if err != nil {
    if os.IsPermission(err) {
        logger.Warn().Msg("skipping file: permission denied")
    }
    wg.Done()
    return nil
}
```

That `nil` was the callback's only channel back to the caller, so the open failure could not reach `DetectSource`'s error, which meant `findingSummaryAndExit` took its `err == nil` branch and printed the plain `no leaks found` with exit `0` over a scan that had skipped files.

As the issue notes, the partial-scan distinction in `cmd/root.go:446-461` is already correct in both message and exit code. It was simply unreachable from what is probably the most common failure mode.

## The part that turned out to matter

My first instinct was to return the error from the `s.Sema.Go` goroutine and let it surface. **That does not work, and it is worth recording why:** `s.Sema` is a `semgroup.Group` shared across sources, and `Wait()` is never called on it anywhere in the codebase. Errors returned from those goroutines are appended to the group's internal `errs` slice and never read.

That also means the existing `return err` at the end of the same goroutine — the one carrying failures out of `file.Fragments` — is swallowed today for the same reason. I have left that alone as out of scope for this issue, but it is the same gap and probably wants its own fix.

So the failures are collected in a mutex-guarded local slice and joined into `Fragments`' return value after `wg.Wait()`. That reaches the caller without changing the shared semaphore's semantics or the concurrency model.

## Scanning continues

An unreadable file does not stop the run. Everything reachable is still scanned, and the gap is reported at the end — `TestReadableFilesAreStillScannedAlongsideAnUnreadableOne` asserts the other three files are still yielded. Aborting at the first unopenable file would trade a silent partial scan for a much noisier one.

Non-permission open failures now log too, rather than only `EACCES` being mentioned.

## Tests

`sources/files_unreadable_test.go` — 4 tests, all passing:

- an unreadable file makes `Fragments` return an error naming the file — the case from the issue;
- **a directory of readable files returns no error.** A source that always errored would make every scan look partial, which loses the same signal in the other direction;
- the readable files are still scanned alongside an unreadable one;
- an empty directory is not an error — nothing to read is not the same fact as something that could not be read.

The permission tests skip on Windows and when running as root, where `chmod 000` does not deny access.

`go test ./...` passes across all packages, `go test -race ./sources/` is clean, and `go vet` is clean.


CAOShurong (commented):
**Verification review — built, vetted, and tested from head `509fce15ad2eb2536bc8d840a3ed3e374d73bb2d`.**

Reviewed with AI assistance (build/test reproduction + independent root-cause check). For transparency: I'm also the author of the related gitleaks #2252 (the git-scan silent-pass fix), so this is a complementary sibling fix, not a competing one.

**What I verified (exact head, fresh build):**
- `go build ./...` — clean
- `go vet ./sources/... ./cmd/...` — clean
- `go test ./sources/` — passes. The four new tests split 2 pass / 2 skip on my Windows runner because the permission-bit cases self-skip by design (`chmod 000` is meaningless on Windows, and the suite also skips under root). The two non-permission tests (`TestAllReadableFilesReportNoError`, `TestEmptyDirectoryIsNotAnError`) pass. The macOS/Linux CI path is where the permission tests actually exercise the new `else` branch.

**Root cause confirmed independently.** I grepped the tree: `s.Sema.Go(...)` / `d.Sema.Go(...)` is used in `sources/files.go`, `sources/git.go`, `detect/files.go`, `detect/git.go`, but **no `.Sema.Wait()` (or any Wait) is ever called anywhere in non-test code**. So the author's "the part that turned out to matter" is exactly right — errors returned from those goroutines are accumulated in the semgroup and never read. The previous `return nil` on open failure was the only channel back, so the failure could never reach `DetectSource`/`findingSummaryAndExit`.

**Design is the minimal correct fix.** Collecting unreadable-file errors in a mutex-guarded *local* slice and joining them into `Fragments`' return value reaches the caller without touching the shared semaphore's semantics. This is what finally makes the existing partial-scan branch in `cmd/root.go` (message + non-zero exit) reachable from the file-open path — previously unreachable from the most common failure mode. Preserving scan-continuation over readable siblings (the third test) is the right call: it trades a silent partial scan for a *reported* partial scan rather than a noisy early abort, and an empty directory stays a non-error.

**One follow-up worth flagging (out of scope here, but same root cause).** The `return err` at the end of the same goroutine (line 200 — the *fragment-scan* error from a readable file that fails mid-read) is still routed through `s.Sema.Go` and is therefore still swallowed by the identical missing-`Wait()` gap. You've correctly scoped it out and documented it. It's the same defect class as #2252's git path and would benefit from a sibling fix using the same local-error-collection pattern — happy to coordinate if you open it.

**Minor / non-blocking:** the final `skippedMu.Lock(); defer skippedMu.Unlock()` is technically redundant after `wg.Wait()` (all goroutines are done), but it's harmless and documents intent clearly.

LGTM on approach and correctness. The behavioral fix is real (a CI gitleaks run that hits an unreadable file will now report it instead of printing `no leaks found` and exiting 0), the tests cover the edge cases, and it composes cleanly with #2252.

PR #2253, opened 2026-08-27
Return source read errors from DetectSource

### LINKED ISSUE #1464: Gitleaks returns exit code 0 when a sequence of invalid flags is passed
**Describe the bug**
Gitleaks returns exit code 0 when the order of the flags should have caused it to fail (and return exit code 1).

**To Reproduce**
1. `touch config.toml`
2. `gitleaks detect -s --no-banner . -c config.toml`
3. `echo $?` (0 is returned, despite the output showing the error)

**Expected behavior**
Return exit code 1 (leaks or error encountered).

**Screenshots**
n/a

 - OS: macOS 14.5 (23F79)
 - Gitleaks Version: 8.18.4

**Additional context**
Not passing the config file flag (i.e., `gitleaks detect -s --no-banner .`) returns the expected exit code (1).
I ran into this by chance due to having misplaced the arguments, other flag combinations possibly have the same behavior.

cc @zricethezav

arpitjain099:
Fixes #1464.

It still reproduces on master, so I dug into why rather than assuming #1461 had covered it:

```
$ touch config.toml
$ gitleaks detect -s --no-banner . -c config.toml
ERR [git] fatal: cannot change to '--no-banner': No such file or directory
ERR error="stderr is not empty"
INF 0 commits scanned.
INF no leaks found
$ echo $?
0
```

The reporter's flag ordering is what starts it (`-s` swallows `--no-banner` as the source), and without `-c` you get exit 1 because `initConfig`'s `os.Stat(source)` runs. With `-c` that branch is skipped, so nothing validates the path and the failure lands further down instead.

That is where the actual problem is. `sources.Git` does report it: `listenForStdErr` sets `errEncountered` and the fragment loop does `yield(Fragment{}, err)`. `DetectSource`'s callback then logs it and returns nil:

```go
if err != nil {
    // Log the error and move on to the next fragment
    logger.Error().Err(err).Send()
    return nil
}
```

Skipping past a bad fragment is right, but the error stops there, so `DetectSource` returns nil, and `findingSummaryAndExit` takes its `err == nil` path: "no leaks found", exit 0. Nothing was scanned, and in CI that is indistinguishable from a clean repository.

This collects those errors, keeps skipping past them exactly as before, and joins them into the return value. `findingSummaryAndExit` already handles a non-nil error by reporting a partial scan and exiting 1; that branch was just unreachable for this class of failure. After the change:

```
ERR failed to scan Git repository error="stderr is not empty"
WRN scanned ~0 bytes (0)
WRN partial scan completed in 21.8ms
WRN no leaks found in partial scan
$ echo $?
1
```

`yield` is called from inside `s.Sema.Go`, so the collection is mutex-guarded.

Worth flagging since it is a behaviour change: any scan that previously logged a fragment error and exited 0 will now exit 1 and be labelled a partial scan. That is the point of the change, but it could surface in pipelines that have been quietly tolerating errors. `listenForStdErr` already allowlists the common benign git chatter (rename detection, auto packing, `git gc`), so this should not fire on ordinary scans.

## Testing

`TestDetectSourceReportsFragmentErrors` drives `DetectSource` with a source that yields an error, an error alongside a readable fragment, and neither. The middle case is the one that matters: it asserts the error comes back **and** that the readable fragment was still scanned, so a fix that bails out on the first error instead of skipping would not pass either. On master:

```
--- FAIL: TestDetectSourceReportsFragmentErrors/only_an_error
--- FAIL: TestDetectSourceReportsFragmentErrors/an_error_alongside_a_readable_fragment
```

I also checked a normal clean repo still exits 0, so this is not just turning every scan red.

`go build ./...` and `go test ./...` pass. `go vet ./detect/ ./sources/` reports two pre-existing `cancel` warnings in `detect_test.go` that are also present on an unmodified checkout.


noqt (approved):
Independent review from NOQT at `4f58f98a1424203af39cc00a26f9672e9b45b794`.

This fixes the false-green failure path while preserving useful partial results. On the base revision, an invalid Git range logged an error, reported `no leaks found`, and exited 0. At this revision, the same command reports a partial scan and exits 1. The regression test also confirms that a readable fragment is still scanned when another fragment fails.

Validation performed:

- `go test ./detect -run '^TestDetectSourceReportsFragmentErrors$' -count=1`
- `go test ./detect ./sources -count=1`
- `go test ./... -count=1`
- base/head CLI reproduction: exit 0 before, exit 1 after
- `git diff --check`

`go vet ./detect ./sources` reports the same two pre-existing context-cancel warnings on both the base and head revisions; this change adds no vet finding.

